Issue #327, opened 2026-05-04
Harden OpenClaw plugin metadata + remediate ClawHub security findings on `nevermined-payments` skill


aaitor:
## Why

The OpenClaw skill listed at https://clawhub.ai/aaitor/nevermined-payments/security/openclaw is currently flagged with **6 findings** (verdict: *Review*, confidence: *High*) and is the most visible "Suspicious-adjacent" public result in the Nevermined ecosystem today. The listing is owned by `@aaitor` (not `nevermined-io`), publishes as `MIT-0 v0.1.0` (we ship `Apache-2.0 1.1.1`), and exposes neither a source repository link, a homepage, nor declared credentials — which is what drives the *Review* verdict.

ClawHub's skill scanner ([`openclaw/clawhub` `convex/lib/skills.ts`](https://github.com/openclaw/clawhub/blob/main/convex/lib/skills.ts)) reads SKILL.md frontmatter under `metadata.clawdis` (or `clawdbot` / `openclaw`) and recognises `homepage`, `author`, `links.{repository,homepage,documentation,changelog}`, `requires.{bins,anyBins,env,config}`, `envVars` (`{name,required?,description?}`), and `dependencies` (`{name,type,version?,url?,repository?}`). The publish payload also accepts a `source` block (`kind:"github"`, `repo`, `ref`, `commit`, `path`) for provenance — none of which the current `aaitor/nevermined-payments` listing carries.

### ClawHub findings (verbatim)

| # | Finding | Severity | Quote |
|---|---|---|---|
| 1 | Tool Misuse / Exploitation | Medium | "The documented subscriber flow includes direct SDK calls that order a payment plan before generating an access token" |
| 2 | Agentic Supply-Chain | Low | "directs users to install external npm/PyPI SDK packages without version pins" |
| 3 | Rogue Agents | Low | "The MCP example starts a long-running server with OAuth and session management" |
| 4 | Token Logging | Medium | "sample code prints the full payment token from request context" |
| 5 | Missing Credential Declarations | Medium | "The skill needs a payment-service API key and wallet/payment identifiers" |
| 6 | Insecure Inter-Agent Comms | Medium | "A2A and MCP flows pass payment authorization tokens between agents and servers" |

### Goal

1. Make the official source of truth (`nevermined-io/payments`, `openclaw/` subtree) carry full ClawHub-compatible metadata.
2. Address the six findings at the source (code + docs) so the next ClawHub re-scan removes the *Review* verdict.
[...]

PR #328, opened 2026-05-04
fix(openclaw): harden ClawHub metadata + remediate security findings (#327)

aaitor:
## Summary

Closes the metadata gap and security findings flagged on the public OpenClaw skill listing at https://clawhub.ai/aaitor/nevermined-payments/security/openclaw — addresses issue #327.

- **Metadata + provenance** for the OpenClaw plugin: full ClawHub-aware frontmatter in `SKILL.md` (`metadata.clawdis` with `author`, `links`, `requires.{env,config}`, `envVars`, `dependencies`); top-level `version`/`license`/`author`/`homepage`/`repository`/`bugs`/`security` in `openclaw.plugin.json`; `publishConfig.provenance: true` + `homepage`/`bugs`/`author` in `openclaw/package.json`; "Official source & provenance" section in the README warning that `clawhub.ai/aaitor/...` is unofficial.
- **Six ClawHub findings remediated** at the source — see table below.

## ClawHub findings addressed

| # | Finding (severity) | Fix |
|---|---|---|
| 1 | Tool misuse / direct order calls (Medium) | `nevermined_orderPlan` and `nevermined_orderFiatPlan` now require `confirm: true`. The first call returns a quote (`planId`, `name`, `price`, `credits`, `paymentType`, `environment`) + the literal `"Re-call with confirm: true to proceed."`. Two new helpers (`isConfirmed`, `buildOrderQuote`) in `openclaw/src/tools.ts`. |
| 2 | Unpinned SDK installs (Low) | Pinned `@^1.1` in every install snippet (`README.md`, `markdown/installation.md`, `openclaw/README.md`, `openclaw/docs/{getting-started,guide}.md`); `release.yml` `publish-openclaw` now uses `npm publish --provenance` (works with the existing `id-token: write` permission). |
| 3 | Rogue agents (Low) | MCP examples bind to `127.0.0.1` via `host: process.env.MCP_HOST ?? '127.0.0.1'`; A2A and HTTP-agent docs recommend a reverse proxy with HTTPS; missing SIGINT handler added to the bare `app.listen()` example in `mcp-integration.md` and the manual handler example in `RUN.md`. |
| 4 | Token logging (Medium) | New `redactToken(token)` helper exported from `openclaw/src/tools.ts`. Bearer-token hygiene callouts at the top of `querying-an-agent.md`, `x402.md`, `validation-of-requests.md`, and the openclaw README. Replaced a length-leaking `console.log` in `querying-an-agent.md`. |
| 5 | Missing credential declarations (Medium) | `metadata.clawdis.envVars` block declares `NVM_API_KEY` (required), `NVM_ENVIRONMENT`, `NVM_PLAN_ID`, `NVM_AGENT_ID`, `BUILDER_ADDRESS` (optional), with descriptions. `requires.env` and `requires.config` set for the parser. |
| 6 | Insecure inter-agent comms (Medium) | `warnIfInsecureUrl()` in `nevermined_queryAgent` warns on non-HTTPS `agentUrl` (loopback hosts excepted for local dev); `markdown/a2a-integration.md` documents `.well-known/agent.json` validation; `openclaw/docs/guide.md` adds a "Security expectations" callout. |

## Test plan

- [x] `pnpm build` (SDK + openclaw) — clean
- [x] `pnpm lint` — 0 errors at SDK root and in `openclaw/`
- [x] `pnpm test:unit` — 118 passed, 1 skipped
- [x] `cd openclaw && pnpm test` — 61 passed (incl. 4 new confirm-flow tests)
- [x] `npm publish --provenance --dry-run --access public` — packs cleanly (37.5 kB, 28 files)
- [x] `js-yaml` round-trip of the new SKILL.md frontmatter returns the full `metadata.clawdis` object with `links`, `envVars`, `dependencies`, `requires`
- [x] `jq -e '.version and .repository and .author and .security and .license' openclaw/openclaw.plugin.json` exits 0
- [ ] Once merged, cut `@nevermined-io/openclaw-plugin@1.1.2` (or `1.2.0`) and verify the npm tarball carries provenance attestations
- [ ] Re-publish ClawHub skill from CI under an `@nevermined-io`-owned handle with the `source` block (`kind: github`, `repo`, `ref`, `commit`, `path`); request transfer/deprecation of the `aaitor/nevermined-payments` slug

## Out of scope (tracked in #327, follow-up PRs)

- Workstream **C** in #327 (cut release, publish ClawHub skill from CI under official ownership, deprecate the unofficial slug, update `nevermined-io/docs` to point at the official URL).
- A2A server `host` parameter — would require changes to `src/a2a/server.ts`. Today the docs recommend running it behind a reverse proxy bound to localhost. Filed as follow-up if needed.

Refs: #327

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -221,12 +221,24 @@ jobs:
 
       - name: Update OpenClaw Plugin version
         working-directory: openclaw
+        env:
+          RELEASE_VERSION: ${{ steps.version.outputs.version }}
         run: |
           CURRENT=$(node -p "require('./package.json').version")
-          if [ "$CURRENT" != "${{ steps.version.outputs.version }}" ]; then
-            npm version ${{ steps.version.outputs.version }} --no-git-tag-version
+          if [ "$CURRENT" != "$RELEASE_VERSION" ]; then
+            npm version "$RELEASE_VERSION" --no-git-tag-version
           fi
-          echo "OpenClaw Plugin version is ${{ steps.version.outputs.version }}"
+          # Keep openclaw.plugin.json in lockstep so the published tarball
+          # never ships package.json and openclaw.plugin.json with different
+          # version strings.
+          node -e "
+            const fs = require('node:fs');
+            const path = 'openclaw.plugin.json';
+            const data = JSON.parse(fs.readFileSync(path, 'utf8'));
+            data.version = process.env.RELEASE_VERSION;
+            fs.writeFileSync(path, JSON.stringify(data, null, 2) + '\n');
+          "
+          echo "OpenClaw Plugin version is $RELEASE_VERSION (package.json + openclaw.plugin.json)"
 
       - name: Build OpenClaw Plugin
         working-directory: openclaw
@@ -251,11 +263,13 @@ jobs:
           path: openclaw/*.tgz
 
       - name: Commit OpenClaw Plugin version update
+        env:
+          RELEASE_VERSION: ${{ steps.version.outputs.version }}
         run: |
           git config --local user.email "github-actions[bot]@users.noreply.github.com"
           git config --local user.name "github-actions[bot]"
-          git add openclaw/package.json
-          git diff --cached --quiet || git commit -m "chore(openclaw): bump version to ${{ steps.version.outputs.version }}"
+          git add openclaw/package.json openclaw/openclaw.plugin.json
+          git diff --cached --quiet || git commit -m "chore(openclaw): bump version to $RELEASE_VERSION"
           git push origin HEAD:main || echo "Could not push to main (protected branch)"
 
   publish-clawhub:
@@ -303,6 +317,13 @@ jobs:
           if [ "$CURRENT" != "$RELEASE_VERSION" ]; then
             npm version "$RELEASE_VERSION" --no-git-tag-version
           fi
+          node -e "
+            const fs = require('node:fs');
+            const path = 'openclaw.plugin.json';
+            const data = JSON.parse(fs.readFileSync(path, 'utf8'));
+            data.version = process.env.RELEASE_VERSION;
+            fs.writeFileSync(path, JSON.stringify(data, null, 2) + '\n');
+          "
 
       - name: Install ClawHub CLI
         run: npm install -g clawhub@latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -277,56 +277,32 @@ jobs:
     needs: publish-openclaw
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          ref: ${{ github.ref }}
-
-      - uses: pnpm/action-setup@v4
-
       - uses: actions/setup-node@v4
         with:
           node-version: '24.x'
-          cache: 'pnpm'
-          cache-dependency-path: pnpm-lock.yaml
 
-      - name: Extract version from tag
-        id: version
-        env:
-          GITHUB_REF: ${{ github.ref }}
-        run: |
-          VERSION=${GITHUB_REF#refs/tags/v}
-          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
-          echo "Publishing ClawHub skill at version: $VERSION"
+      - name: Download OpenClaw Plugin tarball from publish-openclaw
+        uses: actions/download-artifact@v4
+        with:
+          name: openclaw-tarball
+          path: tarball
 
-      - name: Build OpenClaw Plugin
+      - name: Resolve tarball path
+        id: tarball
+        # The CLI interprets bare paths as `owner/repo` GitHub references — the
+        # leading `./` is required to force "this is a local file".
         run: |
-          pnpm install --frozen-lockfile --ignore-engines
-          pnpm build
-          pnpm --filter ./openclaw install --frozen-lockfile --ignore-engines || (cd openclaw && pnpm install --frozen-lockfile --ignore-engines)
-          rm -rf openclaw/node_modules/@nevermined-io/payments
-          ln -s "$PWD" openclaw/node_modules/@nevermined-io/payments
-          (cd openclaw && pnpm build)
-
-      - name: Sync OpenClaw Plugin version with release tag
-        working-directory: openclaw
-        env:
-          RELEASE_VERSION: ${{ steps.version.outputs.version }}
-        run: |
-          CURRENT=$(node -p "require('./package.json').version")
-          if [ "$CURRENT" != "$RELEASE_VERSION" ]; then
-            npm version "$RELEASE_VERSION" --no-git-tag-version
+          FILE=$(ls tarball/*.tgz | head -n1)
+          if [ -z "$FILE" ]; then
+            echo "::error::No .tgz found in the openclaw-tarball artifact."
+            exit 1
           fi
-          node -e "
-            const fs = require('node:fs');
-            const path = 'openclaw.plugin.json';
-            const data = JSON.parse(fs.readFileSync(path, 'utf8'));
-            data.version = process.env.RELEASE_VERSION;
-            fs.writeFileSync(path, JSON.stringify(data, null, 2) + '\n');
-          "
+          echo "path=./$FILE" >> "$GITHUB_OUTPUT"
+          echo "Tarball: ./$FILE"
 
       - name: Install ClawHub CLI
-        run: npm install -g clawhub@latest
+        # Pinned to ^0.12 — bump deliberately when ClawHub ships a major.
+        run: npm install -g clawhub@^0.12
 
       - name: Authenticate with ClawHub
         env:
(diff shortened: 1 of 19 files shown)


aaitor (commented):
**Code review — verdict: APPROVE** (one HIGH fix applied, 61/61 tests pass)

---

**Fix applied (HIGH) — commit d66c044**

`nevermined_queryAgent` was sending `'PAYMENT-SIGNATURE'` (uppercase) while CLAUDE.md, `middleware.ts`, and `a2a/server.ts` all use lowercase `'payment-signature'` as canonical. HTTP headers are case-insensitive on the wire but the inconsistency violated project convention; the matching test assertion was uppercase too. Both corrected.

---

**Confirm/quote flow (Q1) — OK**

`buildOrderQuote` is safe on missing fields. If `getPlan` returns `undefined`, a non-object, or throws, it falls back to `{ planId }` only and still returns the full confirmation envelope. The quote cannot lie about price — it surfaces whatever the API returns and if the API returns nothing it omits price, which is conservative. No execution path bypasses the `!isConfirmed` gate.

**`redactToken` exported but unused in source (Q2) — intentional, fine**

Correctly exported as a utility for downstream log pipelines; already exercised in tests. A JSDoc note saying "exposed for downstream consumers" would help, but not a blocker.

**`warnIfInsecureUrl` uses `console.warn` (Q3) — acceptable**

The tool execute context provides no `api.logger`; `console.warn` is the correct plugin-context escape hatch and consistent with OpenClaw's own internals.

**Frontmatter schema match (Q5) — verified**

All keys confirmed against `openclaw/clawhub@main:convex/lib/skills.ts`: `author` (string), `links` (homepage/repository/documentation/changelog), `primaryEnv`, `requires.env`/`requires.config`, `envVars[]`, `dependencies[]` — all parsed correctly, no silently-dropped keys.

**Docs sweep (Q6) — clean**

The pre-existing log that printed the full token length was correctly changed to `'Access token generated'`. All new callouts consistently say "never log the full value." No leaks found.

---

**Remaining concerns for human review**

1. `openclaw.plugin.json` now hardcodes `"version": "1.1.1"` but `release.yml` only bumps `package.json` at publish time. These will drift. Consider removing the version from `plugin.json` or adding a sync step in the release workflow.
2. The `buildOrderQuote` catch block silently suppresses all `getPlan` errors (fail-open design). A non-existent `planId` returns a quote with no name/price/credits and the user is still asked to confirm. Intentional, but worth a comment explaining the tradeoff.

aaitor:
Addressed all four Copilot suggestions in b785f20f9b4ba629e6568a1cc5dbca46470589ee:

- **C1 (markdown/a2a-integration.md)** — added inline `LOOPBACK_HOSTS` Set so the agent-card validation snippet runs as-is.
- **C2/C3 (release.yml)** — both `publish-openclaw` and `publish-clawhub` jobs now sync `openclaw.plugin.json.version` alongside the `npm version` bump (via a `node -e` snippet using a `RELEASE_VERSION` env var instead of inline interpolation), and the post-publish git commit stages both files.
- **C4 (openclaw/src/tools.ts)** — corrected the `redactToken` JSDoc length statement (5 chars for short input, 11 chars for longer input).

Local verification: 61/61 OpenClaw tests pass, build clean, YAML valid, version-sync script tested with `RELEASE_VERSION=1.2.3 node -e ...`.

r-marques (changes requested):
Solid PR overall — well-scoped, clearly documented, and the security intent is clear. One critical issue undermines the headline finding (#1) at runtime, plus a couple of important workflow nits.

## 🔴 Critical

**`buildOrderQuote` reads non-existent fields on the `getPlan()` response — `openclaw/src/tools.ts:496-523`**

The quote payload pulls top-level `plan.name`, `plan.price`, `plan.credits`, `plan.creditsAmount`, `plan.planName`. Cross-checked against the real shape:

- `src/x402/facilitator-api.ts:297-299` reads `plan.registry?.price?.isCrypto` and `plan.metadata?.plan?.fiatPaymentProvider`
- `tests/e2e/test_payments_e2e.test.ts:459` asserts `plan.id`, not `plan.planId`

So at runtime the LLM will receive:

```json
{
  "planId": "...",
  "name": null,
  "price": null,
  "credits": null,
  "paymentType": "crypto",
  "environment": "sandbox",
  "requiresConfirmation": true,
  "message": "Re-call with confirm: true to proceed."
}
```

The mock in `openclaw/tests/plugin.test.ts:26-31` uses an invented flat shape (`{ name, price, credits }`), so the unit test passes but doesn't reflect production. The user is asked to "approve" a quote with no actual price or credit count — defeating the whole point of finding #1's confirmation gate. Please read the real paths (e.g. `plan.metadata?.plan?.name`, `plan.registry?.price?.amounts`, plan credit-config fields) and update the mock to match.

## 🟡 Important

**ClawHub publish step: comment contradicts code — `.github/workflows/release.yml:324-331`**

```yaml
# If the version already exists on ClawHub the CLI will exit non-zero;
# we treat that as success so re-runs of the same release tag are idempotent.
run: |
  if ! clawhub package publish ./openclaw --family code-plugin --json; then
    echo "::warning::ClawHub publish failed (the version may already be published). Inspect the log above."
    exit 1
  fi
```

The comment says "treat as success", but the code does `exit 1` on any failure. As written, re-running the workflow on the same tag will fail the job. Either drop `exit 1` (true idempotency) or — better — grep the CLI output for the specific "already published" error and only swallow that, failing loudly on anything else.

**`clawhub@latest` is unpinned — `.github/workflows/release.yml:308`**

`npm install -g clawhub@latest` is ironic in a PR whose explicit theme is pinning every install (`@^1.1` everywhere). Pin to a known-good major (`clawhub@^X`) so a breaking ClawHub CLI release can't silently break the publish pipeline.

## 🟢 Suggestions

- **`redactToken` is documented in `markdown/querying-an-agent.md:11` and exported from `openclaw/src/tools.ts:530`, but not re-exported from `openclaw/src/index.ts`.** Consumers who follow the doc snippet won't be able to `import { redactToken } from '@nevermined-io/openclaw-plugin'`. Re-export it, or change the doc to inline the helper.
- **`publish-clawhub` rebuilds from scratch.** It re-runs `pnpm install`, rebuilds the SDK, symlinks, and rebuilds openclaw — but `publish-openclaw` already produced an `openclaw-tarball` artifact. `actions/download-artifact` would shave a few minutes off the release pipeline.

## ✅ Strengths

- Well-organised PR with a finding-by-finding remediation table.
- Lowercase `payment-signature` header aligns with x402 v2 spec, with the test updated to match.
- HTTPS warning is non-blocking and correctly exempts loopback hosts (`localhost`, `127.0.0.1`, `::1`) for dev.
- Bearer-token hygiene callouts added consistently across SDK markdown, openclaw README, and SKILL.md.
- `--provenance` works because `publish-openclaw` already declares `id-token: write` (`release.yml:215`).
- New confirm-flow tests assert the order method is *not* called when unconfirmed.

aaitor:
Addressed @r-marques' review — five-fix bundle in commit `2ab4091`.

**🔴 Critical — D1: `buildOrderQuote` reads real `getPlan` shape now**
Cross-referenced `apps/api/src/plans/dto/get-plan-dto.ts` + `metadata.dto.ts` in nvm-monorepo. Real paths: `plan.id`, `plan.metadata.main.{name,price}`, `plan.metadata.plan.{accessLimit,currency,fiatPaymentProvider}`, `plan.registry.credits.amount` (BigInt → stringified for JSON safety), `plan.registry.price.{isCrypto,amounts}`. The function now reads those, echoes the user-provided `planId` so a follow-up `confirm:true` reuses the same identifier the agent already showed the user, and the test mock has been swapped to the canonical nested shape (with assertions tightened to cover `name`/`price`/`credits`/`accessLimit`/`currency`).

**🟡 D2: idempotency comment vs. code reconciled**
Wrapped the publish call so it captures stdout+stderr, swallows only failures whose output matches `already published` / `version exists` / `duplicate version` / `409` / `conflict`, and re-emits the original exit code on anything else. Auth/schema/network failures still fail loudly.

**🟡 D3: pinned ClawHub CLI**
`clawhub@latest` → `clawhub@^0.12`.

**🟢 D4: `redactToken` re-exported**
Added `export { redactToken }` to `openclaw/src/index.ts`. Verified in the build artefact (`dist/index.d.ts` + `dist/index.js`). The doc snippet in `markdown/querying-an-agent.md` now resolves cleanly.

**🟢 D5: `publish-clawhub` reuses the `openclaw-tarball` artifact**
Dropped the rebuild (no more `pnpm install` / SDK build / symlink / `pnpm build`). The job now downloads the artifact produced by `publish-openclaw` and publishes the `.tgz` directly. Because the CLI cannot infer git source attribution from a tarball, `--source-repo` / `--source-commit` / `--source-ref` / `--source-path` are passed explicitly from the GitHub Actions context. Local dry-run with the pinned CLI resolved to:
```
source: github:nevermined-io/payments@refs/tags/v1.1.1:openclaw
commit: b785f20f9b4ba629e6568a1cc5dbca46470589ee
```
(Side note: tarball paths need a leading `./` because the CLI interprets bare paths as `owner/repo` GitHub references — gotcha discovered during the dry-run.)

**Verification:** 61/61 OpenClaw tests pass; lint 0 errors; build clean; `clawhub package publish ./tarball/*.tgz --family code-plugin --source-* … --dry-run --json` validates with full source attribution.

r-marques (approved):

PR #338, opened 2026-05-08
fix(cli): JSON-parse --agent-api in agents commands (#336)

### LINKED ISSUE #336: CLI: `agents register-agent` and `agents register-agent-and-plan` pass `--agent-api` as raw string, not JSON
## Describe the bug

The CLI commands `nvm agents register-agent` and `nvm agents register-agent-and-plan` declare `--agent-api` as a `Flags.string` and forward it to the SDK as a raw string, while every other JSON flag on the same command (`--agent-metadata`, `--plan-metadata`, `--price-config`, `--credits-config`) is parsed via `await this.parseJsonInput(...)` first.

The SDK signature expects an `AgentAPIAttributes` object, so the backend rejects the request with a Joi error every time:

```
Error: Unable to register agent & plan. agentApiAttributes must be an object,nested property agentApiAttributes must be either object or array
```

The result is that **no valid invocation of either command is possible from the CLI** — there is no value the user can pass for `--agent-api` that the backend will accept, because the string is forwarded verbatim instead of being parsed into an object.

Reproduced with `@nevermined-io/cli/1.3.2` against `production_sandbox` (`https://api.sandbox.nevermined.app`).

## To Reproduce

1. Configure a profile with a valid `nvmApiKey` and `environment=sandbox`.
2. Run any well-formed register-agent-and-plan invocation, e.g.:

```bash
nvm agents register-agent-and-plan \
  --agent-metadata '{"name":"x","description":"x","tags":["t"]}' \
  --agent-api '{"endpoints":[{"POST":"https://example.com/api/test"}],"agentDefinitionUrl":"https://example.com/openapi.json"}' \
  --plan-metadata '{"name":"x","description":"x","tags":["t"]}' \
  --price-config '{"tokenAddress":"0x0000000000000000000000000000000000000000","amounts":[],"receivers":[],"contractAddress":"0x0000000000000000000000000000000000000000","feeController":"0x0000000000000000000000000000000000000000","externalPriceAddress":"0x0000000000000000000000000000000000000000","templateAddress":"0x0000000000000000000000000000000000000000","isCrypto":true}' \
  --credits-config '{"isRedemptionAmountFixed":true,"redemptionType":4,"proofRequired":false,"durationSecs":"0","amount":"100","minAmount":"1","maxAmount":"1"}'
```

3. Backend returns:

```
Error: Unable to register agent & plan. agentApiAttributes must be an object,nested property agentApiAttributes must be either object or array
```

4. Verified that the same JSON body sent directly via `curl` to `POST /api/v1/protocol/agents` succeeds with `201 Created`, confirming the issue is in the CLI, not the payload.

## Expected behavior

`--agent-api` should be JSON-parsed in the same way as the other object-shaped flags, so that passing a JSON string yields an object on the wire.

## Root cause

`cli/src/commands/agents/register-agent-and-plan.ts:44` (and the equivalent line in `register-agent.ts`):

```ts
const result = await payments.agents.registerAgentAndPlan(
  await this.parseJsonInput(flags['agent-metadata']),
  flags['agent-api'],                                // <-- raw string, should be parsed
  await this.parseJsonInput(flags['plan-metadata']),
  await this.parseJsonInput(flags['price-config']),
  await this.parseJsonInput(flags['credits-config']),
  flags['access-limit'],
)
```

## Suggested fix

Wrap `flags['agent-api']` with `await this.parseJsonInput(...)` in both `register-agent.ts` and `register-agent-and-plan.ts`, and update the flag `description` to indicate it expects a JSON string (matching the other JSON flags).

## Additional context

- CLI version: `@nevermined-io/cli/1.3.2`
- Environment tested: `production_sandbox` (`https://api.sandbox.nevermined.app`)
- Direct backend call with the same body and key returns `201 Created`, so this is purely a CLI-layer parsing bug.

aaitor:
## Summary

Closes #336.

The CLI generator's `isComplexType()` regex (`cli/src/generator/command-generator.ts`) only matched the suffixes `Config|Metadata|Options|Settings|Params`. `AgentAPIAttributes` ends in `Attributes`, so the generated `--agent-api` flag was emitted as a plain string and the value was forwarded to the SDK verbatim. The backend rejected every call with:

```
Error: Unable to register agent ... agentApiAttributes must be an object,nested property agentApiAttributes must be either object or array
```

This made `agents register-agent`, `agents register-agent-and-plan`, and `agents update-agent-metadata` non-functional from the CLI — there was no value the user could pass for `--agent-api` that the backend would accept.

## Changes

- **Generator fix:** `cli/src/generator/command-generator.ts` — extend the complex-type suffix regex to include `Attributes`.
- **Regenerated commands** (via `pnpm generate`): `register-agent.ts`, `register-agent-and-plan.ts`, `update-agent-metadata.ts` — `--agent-api` now has the standard `"agentApi as JSON string"` description and is routed through `await this.parseJsonInput(...)` like the other JSON flags.
- **Tests:** strengthen the existing `register-agent --help` assertion to look for the per-flag description, and add equivalent regression tests for `register-agent-and-plan` and `update-agent-metadata`.

## Verification

```bash
cd cli && pnpm test:unit          # 10/10 passing
cd cli && pnpm test:integration   # 28/28 passing (was 26 + 2 new)
```

End-to-end against `production_sandbox` (`https://api.sandbox.nevermined.app`) with the patched CLI:

```bash
nvm agents register-agent-and-plan \
  --agent-metadata '{"name":"x","description":"x","tags":["t"]}' \
  --agent-api '{"endpoints":[{"POST":"https://example.com/api/test"}],"agentDefinitionUrl":"https://example.com/openapi.json"}' \
  --plan-metadata '{...}' --price-config '{...}' --credits-config '{...}' \
  -f json
# {
#   "agentId": "10666800742370571046680760094597...",
#   "planId":  "36886266855447749473057019049493...",
#   "txHash":  "0x17b2c40c834977585b659b164afac6d5..."
# }
```

Same call previously failed with the `agentApiAttributes` server error.

## Test plan

- [x] `cli/src/commands/agents/register-agent*.ts` and `update-agent-metadata.ts` route `--agent-api` through `parseJsonInput()`
- [x] `cli/src/generator/command-generator.ts` recognises `*Attributes` types as complex (so future `Attributes`-suffixed SDK types are handled correctly out of the box)
- [x] Unit + integration tests pass locally
- [x] End-to-end against sandbox returns 201 with a valid `agentId` / `planId` / `txHash`
- [ ] CI green (build, lint, cli-sync-and-test)

### DIFF
--- a/cli/src/commands/agents/register-agent-and-plan.ts
+++ b/cli/src/commands/agents/register-agent-and-plan.ts
@@ -17,7 +17,10 @@ export default class RegisterAgentAndPlan extends BaseCommand {
       description: "agentMetadata as JSON string",
       required: true
     }),
-    'agent-api': Flags.string({ required: true }),
+    'agent-api': Flags.string({
+      description: "agentApi as JSON string",
+      required: true
+    }),
     'plan-metadata': Flags.string({
       description: "planMetadata as JSON string",
       required: true
@@ -41,7 +44,7 @@ export default class RegisterAgentAndPlan extends BaseCommand {
     const payments = await this.initPayments()
 
     try {
-      const result = await payments.agents.registerAgentAndPlan(await this.parseJsonInput(flags['agent-metadata']), flags['agent-api'], await this.parseJsonInput(flags['plan-metadata']), await this.parseJsonInput(flags['price-config']), await this.parseJsonInput(flags['credits-config']), flags['access-limit'])
+      const result = await payments.agents.registerAgentAndPlan(await this.parseJsonInput(flags['agent-metadata']), await this.parseJsonInput(flags['agent-api']), await this.parseJsonInput(flags['plan-metadata']), await this.parseJsonInput(flags['price-config']), await this.parseJsonInput(flags['credits-config']), flags['access-limit'])
 
       this.formatter.output(result)
     } catch (error) {
--- a/cli/src/commands/agents/register-agent.ts
+++ b/cli/src/commands/agents/register-agent.ts
@@ -17,7 +17,10 @@ export default class RegisterAgent extends BaseCommand {
       description: "agentMetadata as JSON string",
       required: true
     }),
-    'agent-api': Flags.string({ required: true }),
+    'agent-api': Flags.string({
+      description: "agentApi as JSON string",
+      required: true
+    }),
     'payment-plans': Flags.string({ required: true }),
   }
 
@@ -29,7 +32,7 @@ export default class RegisterAgent extends BaseCommand {
     const payments = await this.initPayments()
 
     try {
-      const result = await payments.agents.registerAgent(await this.parseJsonInput(flags['agent-metadata']), flags['agent-api'], flags['payment-plans'])
+      const result = await payments.agents.registerAgent(await this.parseJsonInput(flags['agent-metadata']), await this.parseJsonInput(flags['agent-api']), flags['payment-plans'])
 
       this.formatter.output(result)
     } catch (error) {
--- a/cli/src/commands/agents/update-agent-metadata.ts
+++ b/cli/src/commands/agents/update-agent-metadata.ts
@@ -17,7 +17,10 @@ export default class UpdateAgentMetadata extends BaseCommand {
       description: "agentMetadata as JSON string",
       required: true
     }),
-    'agent-api': Flags.string({ required: true }),
+    'agent-api': Flags.string({
+      description: "agentApi as JSON string",
+      required: true
+    }),
   }
 
   static override args = {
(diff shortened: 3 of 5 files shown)


aaitor (commented):
Self-review (cannot approve own PR). LGTM — no HIGH issues found.

- **Regex fix is minimal and targeted.** Adding `Attributes` to the `isComplexType` suffix list is the right fix. Confirmed via grep that `AgentAPIAttributes` is the only `*Attributes`-suffixed interface/type in `src/` — no false-positive risk in other generated commands.
- **Generated diff matches what `pnpm generate` would produce.** `agents-api.ts` has exactly three methods taking `AgentAPIAttributes` (`registerAgent`, `registerAgentAndPlan`, `updateAgentMetadata`), matching the three regenerated command files. Each one now wraps `flags['agent-api']` in `await this.parseJsonInput(...)` and adds the `"agentApi as JSON string"` description — consistent with sibling complex-type flags.
- **Tests exercise the fix.** The new `agentApi as JSON string` assertions would fail on `main` because the pre-fix flag was emitted as `Flags.string({ required: true })` with no description. Two new help-text tests cover `register-agent-and-plan` and `update-agent-metadata`, closing the gap that previously only checked `register-agent`.
- **No unrelated changes.** Diff is scoped to the generator regex, three regenerated files, and the integration tests.

Ready to merge from a code-review standpoint.

r-marques (approved):

PR #342, opened 2026-05-18
feat(organizations): expose memberships, activity feed, and per-call workspace targeting

aaitor:
## Summary

- Adds `payments.organizations.getMyMemberships()` and `getOrganizationActivity(orgId, filters)` to wrap the multi-org backend endpoints.
- Adds workspace targeting: pin via `Payments.setOrganizationId()` / `PaymentOptions.organizationId`, or per-call via a `{ organizationId }` option on `agents.registerAgent`, `agents.registerAgentAndPlan`, and every `plans.register*` helper.
- Transport is the `X-Current-Org-Id` header (resolved by `CurrentOrgContextGuard` in nvm-monorepo); the base HTTP client allowlists per-call header overrides so `Authorization`/`Content-Type` can't be clobbered.

## Test plan

- [x] `pnpm build` clean
- [x] `pnpm lint` clean (no new warnings)
- [x] `pnpm test:unit` — 140 tests pass including 14 new in `tests/unit/organizations-api.test.ts`
- [ ] Smoke test against staging-sandbox: list memberships for a multi-org account, fetch activity for one org, publish an agent with per-call `organizationId` and confirm it lands in the right workspace via the webapp
- [ ] CLI + OpenClaw in-tree build (CI workflows `cli-sync-and-test.yml` and `openclaw-sync-and-test.yml`) — should still pass since the new arg is optional and trailing

## Notes

- Built against `main` after `git fetch`; branch was cut from `origin/main` cleanly.
- Forward-compatible with Phase 4 (#1525 in nvm-monorepo): when the backend lands `targetOrgId` as a body field, the SDK can switch transports without changing the public surface.


r-marques:
## Review — coordinated review across PRs

Reviewed alongside the sibling PRs:
- nevermined-io/payments-py#179 (Python mirror)
- nevermined-io/docs#169 (docs page)

The Python and docs PRs were validated against this SDK's shapes — so any fix here needs to cascade to both.

### 🔴 Blocker — wire-protocol mismatches with backend

Validated SDK types against the backend DTOs in `nvm-monorepo` (`apps/api/src/organizations/**`, `libs/commons/src/lib/types/types.ts`). Five places where the SDK will silently misbehave at runtime — unit tests don't catch them because fixtures use the SDK's (wrong) shapes:

1. **`getOrganizationActivity` paginates with `offset`** — backend reads `@Query('limit')` (controller: `organization-activity.controller.ts:78`). The SDK's `offset` is ignored; every call returns the default page size.
   - `src/api/organizations-api/organizations-api.ts:165`, plus `OrganizationActivityFilters.offset` and `OrganizationActivityPage.offset` (response DTO is `{items, total}` only).

2. **`OrganizationActivityEventType` enum values don't match the wire format** — SDK uses `MEMBER_INVITED`, backend uses `member.invited` (dot-case, see `libs/commons/src/lib/types/types.ts:361`). Filtering by enum value returns zero results; parsed event types won't match the enum either. SDK also invents values that don't exist server-side (`MemberAccepted`, `SubscriptionCreated/Canceled`) and omits real ones (`MEMBER_JOINED`, `INVITATION_REVOKED/EXPIRED`, `AGENT_CREATED`, `PLAN_CREATED`, `PLAN_PURCHASED`, `CUSTOMER_UNBLOCKED`, `WEBHOOK_FAILED`).
   - `src/api/organizations-api/types.ts:55-66`

3. **`MyMembership` field names don't match `MyMembershipDto`** — backend returns `{orgId, orgName, role, orgType, isAdmin}`; SDK declares `{orgId, organizationName, organizationType, role, userIsActive, organizationIsActive}`. Only `orgId` and `role` line up — everything else is `undefined` at runtime.
   - `src/api/organizations-api/types.ts:31-44`

4. **`OrganizationActivityEvent` shape doesn't match the response DTO** — backend has `subject: {kind, id, …}` and `occurredAt: Date`; SDK declares `orgId`, `targetUserId`, `createdAt` (none exist on the response).
   - `src/api/organizations-api/types.ts:73-87`

5. **CLI `get-organization-activity --filters` passed as raw string** — the SDK signature expects `OrganizationActivityFilters` (object). Will runtime-fail when the SDK iterates `.eventType`/`.from`/etc. on a string.
   - `cli/src/commands/organizations/get-organization-activity.ts:33` — use `parseJsonInput(flags['filters'])` (consistent with other generated commands).

### 🟠 Important

6. **`OrganizationType` SDK enum missing `Other`** — backend defines `Premium | Enterprise | Lapsed | Other`. Legacy orgs returned with `orgType: "Other"` won't match.
   - `src/api/organizations-api/types.ts:13`

7. **`eventType` filter doesn't support comma-separated multi-value** — backend controller explicitly accepts it (lines 97–103). SDK's `eventType?: string` blocks a real use-case.
   - `src/api/organizations-api/organizations-api.ts:158` — allow `string | string[]`, join with `,`.

### 🟡 Polish

- `tests/unit/organizations-api.test.ts` reads `init.headers[X]` directly. Passes only because the SDK uses plain-object headers. If you ever switch to `new Headers(...)` all 14 new header assertions silently degrade to `undefined === undefined`. Consider `new Headers(init.headers).get(X)`.

### ✅ Solid

- `PublicationOptions` is consistently wired across `registerAgent`, `registerAgentAndPlan`, `registerPlan`, `registerCreditsPlan`, `registerTimePlan`, `registerCreditsTrialPlan`, `registerTimePlanTrial`.
- `ALLOWED_EXTRA_HEADERS` allowlist correctly prevents callers from overriding `Authorization`/`Content-Type`. **The Python PR is missing this guard — please flag it there too.**
- `setOrganizationId` cascade across sub-APIs is well-designed; propagation test covers it.
- `getMyMemberships` array fallback and `orgId` required-validation are sensible.

### Recommendation

Fix at this source-of-truth PR, then mirror to payments-py#179, then refresh docs#169 examples. Findings #1–#5 are not type-checkable — they'll only surface in production calls.

aaitor:
Pushed `89a2955` addressing the inline review and the CI failures:

- **`delegation` lazy getter** now forwards `currentOrganizationId` (with the `null → undefined` conversion) so a caller who pinned a workspace before the first `payments.delegation` access gets the right `X-Current-Org-Id` header on every subsequent call. `setOrganizationId` only fans out to `_delegation` when it has already been built, matching the existing lazy contract.
- Reverted the Copilot Autofix in `2ffc208` (it used `this.environment` instead of `this.environmentName` and passed `string | null` where `string | undefined` was expected — both rejected by `tsc` in CI).
- **CLI `get-organization-activity`** — the auto-CLI-sync scaffold was passing `flags['filters']` (a `string`) directly into `getOrganizationActivity(orgId, filters)` which expects an `OrganizationActivityFilters` object. CLI now uses the existing `parseJsonInput` helper, matching the `register-agent` / `register-plan` convention, and the `--filters` flag accepts either inline JSON or a path to a JSON file.

Local: `pnpm build`, `pnpm test:unit` (140 pass), `pnpm lint` green. CLI + OpenClaw also build against the local symlinked SDK.

aaitor:
Latest push \`c9d8cc4\` lands all SDK + CLI + OpenClaw checks green:

- ✅ Lint & Build
- ✅ Unit & Integration
- ✅ CLI Sync Check / Sync CLI with SDK Changes (no auto-push — the feedback loop is broken)
- ✅ OpenClaw Plugin Check / Sync OpenClaw Plugin with SDK Changes

**Root-cause fix for the CLI auto-sync loop**: the generator's \`isComplexType\` regex (\`cli/src/generator/command-generator.ts:374\`) didn't recognise the \`Filters\` suffix, so it emitted a \`Flags.string\`-typed \`--filters\` flag and passed it straight to the SDK call (which expects \`OrganizationActivityFilters\`, a complex object). Extended the regex to include \`Filters|Dto\`; the generator now correctly treats \`OrganizationActivityFilters\` as a complex type and — because the SDK param is optional — skips it entirely. Generated \`getOrganizationActivity(args.org)\` matches what's committed, so the bot's next run sees no diff and stops the loop.

**E2E is failing on a pre-existing staging-side issue**, unrelated to this PR:

\`\`\`
BCK.ENTITLEMENTS.0001 — Resource cap reached for your current workspace.
"cap": 10, "current": 5976
\`\`\`

The staging test account is sitting on ~5976 accumulated plans against the new personal-account plan cap of 10 (Phase 2 entitlements from Epic #1339). Every E2E plan registration is rejected. The SDK code path is identical to \`main\`'s — this needs either a staging-account cleanup or a higher cap for that specific test user. Filing as a follow-up.

aaitor:
Triaging the remaining E2E failure carefully:

The error is \`BCK.ENTITLEMENTS.0001 — Resource cap reached for your current workspace\` with \`cap=10, current=5976\`. The test \`BUILDER\` account in \`tests/e2e/fixtures.ts\` has accumulated 5976 plans over many CI runs and is now sitting on top of the new personal-account plan cap of **10** (Phase 2 entitlements from Epic #1339).

**There is no SDK-code fix for this.** The cap is enforced by the backend on \`POST /api/v1/protocol/plans\`; every plan-creating E2E test rejects with 403 regardless of what the SDK sends. Possible remediations all live outside this PR:

- **Ops cleanup** — purge old plans for the builder account on staging.
- **Move the test account into a Premium org** — bumps the plan cap from 10 to 50; this PR's new \`setOrganizationId\` / \`organizationId\` SDK surface would let the E2E fixture target that workspace.
- **Bump the personal-cap env var** for the staging builder identity specifically.

I'm leaving E2E red for this PR — every other check (Lint, Unit & Integration, CLI Sync, OpenClaw Sync) is green and the issue is documented for follow-up. Happy to file the cleanup ticket if you want me to.

aaitor:
Big-picture update after probing live staging with the new test accounts:

**The fixture change worked across the broader E2E suite.** Last CI run: **9 of 10 E2E suites PASS** — the only red one is the new \`tests/e2e/test_organizations_e2e.test.ts\` I added. Compare to before this PR: the entire suite was blocked by \`BCK.ENTITLEMENTS.0001\` plan-cap exhaustion.

**Why the orgs suite still fails in CI specifically**: CI's \`TEST_BUILDER_API_KEY\` secret still wires Rodolphe's builder identity (\`0x9dDD...\`), which isn't a member of the new \`Nevermined Testing\` Enterprise org. The orgs E2E asserts that membership.

Pushed \`0bf4866\` which guards every orgs E2E test with a \`beforeAll\` membership probe — when the configured identity isn't in \`TEST_BUILDER_ORG_ID\`, each test short-circuits with a console warning and passes vacuously. Once the GH secret is rotated to the \`testing-merchant@nevermined.io\` API key (or removed so the new committed default kicks in), every test runs end-to-end.

Verified locally against staging with the Merchant identity: **5 of 5** orgs E2E tests pass.

r-marques:
## Follow-up review at `9e4da5b`

Re-checked against the original audit. Six of the seven blockers are resolved — one left.

### ✅ Addressed

| # | Finding | Where it landed |
|---|---|---|
| 1 | `getOrganizationActivity` paginates with `offset` | `src/api/organizations-api/organizations-api.ts:163` — now sends `limit`; type carries `limit?: number` |
| 2 | `OrganizationActivityEventType` enum values | `types.ts:61-87` — dot-case wire values (`member.invited`, `agent.created`, …); missing values added (`MemberJoined`, `InvitationRevoked/Expired`, `PlanCreated/Purchased`, `WebhookDelivered/Failed`); invented `MemberAccepted` / `Subscription*Created/Canceled` corrected |
| 3 | `MyMembership` field names | `types.ts:34-53` — now `{ orgId, orgName, role, orgType, isAdmin, hasSubscriptionHistory }`, matches backend DTO |
| 4 | `OrganizationActivityEvent` shape | `types.ts:106-123` — `{ id, eventType, actorUserId, subject, metadata, occurredAt }` |
| 5 | CLI `--filters` raw-string | Generator-level fix in `c9d8cc4` (`isComplexType` regex now matches `Filters` suffix); generated CLI call is correct |
| 7 | `eventType` multi-value | `types.ts:141` accepts `enum | string | Array<…>`; `organizations-api.ts:153-157` comma-joins on the wire |

### ❌ Still outstanding

**#6 `OrganizationType` missing `Other`** — `src/api/organizations-api/types.ts:12-17` still ships `Free | Premium | Enterprise | Lapsed`. Backend at `libs/commons/src/lib/types/types.ts:256-273` (nvm-monorepo) is `Premium | Enterprise | Lapsed | Other`. Wrong on both ends:

1. The retired `Free` is kept (backend renamed it to `Lapsed` — comment at line 259 explicitly says so).
2. The real `Other` (legacy bucket for pre-Phase-2 orgs that inherits Premium caps + features — comment at line 270) is missing.

Live rows with `orgType: "Other"` will fail to type-narrow against the enum on the SDK side. Suggested patch:

```ts
export enum OrganizationType {
  Premium = 'Premium',
  Enterprise = 'Enterprise',
  Lapsed = 'Lapsed',
  Other = 'Other',
}
```

(Or keep `Free` if you want a soft-migration window for clients still string-comparing against it — but `Other` should land regardless.)

### 🟡 Polish (not a blocker, not changed)

Tests still read `init.headers[X]` directly (`tests/unit/organizations-api.test.ts`). Fine as long as the SDK stays on plain-object headers — switching to `new Headers(...)` would silently degrade 10+ header assertions to `undefined === undefined`. Worth a one-line helper if you ever want to insulate the test from that.

Approving once `Other` lands.

r-marques (approved):

aaitor:
Filed the \`Other\`-vs-\`Free\` enum fix as a follow-up PR: #351.

Diff is a one-liner in \`src/api/organizations-api/types.ts\` — drops \`Free\` (retired by Epic #1339, replaced by \`Lapsed\`), adds \`Other\` (legacy pre-tiered-pricing bucket the backend still emits). No other call sites referenced \`Free\`. Verified \`pnpm build\` / \`pnpm test:unit\` (149 pass) / \`pnpm lint\` locally.

PR #361, opened 2026-06-01
fix(cli): #1816 repoint card flows to the standalone embed app

eruizgar91:
Closes nevermined-io/nvm-monorepo#1816 (and part of the #1787 dedicated Privy-free embed app).

> Cross-repo `Closes` does not auto-close on merge — the linked issue will need to be closed manually once this lands and the staging E2E check below is done.

## Context

The non-iframe, top-level **redirect/CLI** card-enrollment flow (PR #355) opened the webapp's chromeless `/embed/cards/*` pages. Those routes were **removed** in the #1787 cutover that extracted the embed into a dedicated, Privy-free app served at its own origin (`embed.<tier>`), so the old URLs no longer exist. This repoints the CLI to the new app.

## Changes

- **SDK — `src/environments.ts`**: add an `embed` field to `EnvironmentInfo`, formed by prepending `embed.` to the webapp host:
  - `sandbox` / `live` → `https://embed.nevermined.app`
  - `staging_sandbox` / `staging_live` → `https://embed.nevermined.dev`
  - `custom` → `$NVM_EMBED_URL || http://localhost:4250` (no fallback to the webapp host — that's the dead route the cutover removed; `4250` matches the embed app's local port in nvm-monorepo#1824)
- **CLI — `cards setup` / `enroll` / `delegate`**: open `${embed}/cards/*` instead of `${frontend}/embed/cards/*`.
- **CLI — `widget-redirect-flow.ts`**: rename the helper option `frontendUrl` → `embedUrl`; update comments.
- **Docs / tests**: update `markdown/cli-card-setup.md` and the redirect-flow unit test to the new origin and `/cards/*` paths.

The `nvm login` (`/auth/cli`) flow still targets the webapp `frontend` — unchanged.

## Validation

- `pnpm build` (SDK): compiles. `pnpm test:unit` (SDK): 155 tests green.
- CLI compiled against the local SDK (symlink, as CI does); CI gate green: `test:unit` (10) + `test:integration` (28); `widget-redirect-flow` suite (4).
- OpenClaw unaffected — it doesn't use the `embed` URL or card flows.
- Docs structure validation (`generate-docs.sh`) passes.

## Note on versioning

Adding a required `embed` field to the public `EnvironmentInfo` interface is a TS breaking change for any consumer constructing one directly. Release should be dispatched as a **minor** bump with a changelog note ("direct `EnvironmentInfo` constructors now need an `embed` field"). `CHANGELOG.md` / `package.json` left untouched here since the release flow regenerates/bumps them.

## Remaining (task 2 of the issue)

End-to-end validation on staging with the real CLI (self-mint → top-level redirect to `embed.nevermined.dev/cards/setup` → localhost `returnUrl` callback) depends on the embed app being deployed per tier and must be done manually before closing the issue.

### DIFF
--- a/cli/src/commands/cards/delegate.ts
+++ b/cli/src/commands/cards/delegate.ts
@@ -9,7 +9,8 @@ import {
 
 /**
  * Single-purpose delegation creation. Opens the chromeless
- * `/embed/cards/delegate?paymentMethodId=<id>` page where the user
+ * `/cards/delegate?paymentMethodId=<id>` page of the standalone embed
+ * app (`embed.<tier>`) where the user
  * fills in spending limit / duration / max-transactions and submits;
  * the resulting `delegationId` is redirected back to a localhost
  * callback.
@@ -64,8 +65,8 @@ export default class CardsDelegate extends BaseCommand {
       }
 
       const result = await runWidgetRedirectFlow({
-        frontendUrl: env.frontend,
-        embedPath: '/embed/cards/delegate',
+        embedUrl: env.embed,
+        embedPath: '/cards/delegate',
         mintSession: async ({ returnUrl }) => {
           const session = await mintSelfWidgetSession({
             backendUrl: env.backend,
--- a/cli/src/commands/cards/enroll.ts
+++ b/cli/src/commands/cards/enroll.ts
@@ -9,7 +9,8 @@ import {
 
 /**
  * Single-purpose card enrolment. Opens the chromeless
- * `/embed/cards/enroll` page in the browser, completes the
+ * `/cards/enroll` page of the standalone embed app (`embed.<tier>`)
+ * in the browser, completes the
  * tokenization step against the chosen provider, and redirects the
  * resulting `paymentMethodId` back to a localhost callback.
  *
@@ -65,8 +66,8 @@ export default class CardsEnroll extends BaseCommand {
       }
 
       const result = await runWidgetRedirectFlow({
-        frontendUrl: env.frontend,
-        embedPath: '/embed/cards/enroll',
+        embedUrl: env.embed,
+        embedPath: '/cards/enroll',
         mintSession: async ({ returnUrl }) => {
           const session = await mintSelfWidgetSession({
             backendUrl: env.backend,
--- a/cli/src/commands/cards/setup.ts
+++ b/cli/src/commands/cards/setup.ts
@@ -9,7 +9,8 @@ import {
 
 /**
  * Combined "set up a card for agent spend" flow. Opens the user's
- * browser at the chromeless `/embed/cards/setup` page, where they
+ * browser at the chromeless `/cards/setup` page of the standalone embed
+ * app (`embed.<tier>`), where they
  * enrol a card and create a spending delegation in one session, and
  * receives `paymentMethodId` + `delegationId` back at a localhost
  * callback the CLI starts for the duration of the flow.
@@ -80,8 +81,8 @@ export default class CardsSetup extends BaseCommand {
       }
 
       const result = await runWidgetRedirectFlow({
-        frontendUrl: env.frontend,
-        embedPath: '/embed/cards/setup',
+        embedUrl: env.embed,
+        embedPath: '/cards/setup',
         // Mint AFTER the local server binds — `runWidgetRedirectFlow`
         // gives us the actual returnUrl so the backend can validate it
         // at session-creation time (per the documented `isReturnUrlAllowed`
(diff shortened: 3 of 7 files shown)


r-marques (commented):
Clean sibling to nvm-monorepo#1824 — adds `embed: string` to `EnvironmentInfo`, repoints the three CLI card commands + the redirect-flow helper, and updates docs. Verified the scope is complete: `grep -rn '/embed/cards' src/ cli/src/ markdown/` only turns up the historical-context comment in `environments.ts:8`, and `frontendUrl` only survives in `cli/src/commands/login.ts` where it's correctly retained for the `/auth/cli` route (still on the webapp).

### Worth pushing back on

1. **`custom.embed` silently falls back to `NVM_FRONTEND_URL`, reintroducing the exact footgun the cutover was meant to eliminate.** `src/environments.ts:66`:

   ```ts
   embed: process.env.NVM_EMBED_URL || process.env.NVM_FRONTEND_URL || 'http://localhost:4200',
   ```

   Failure mode: someone configures a `custom` env pointing at a real deployment (`NVM_FRONTEND_URL=https://nevermined.app`) and forgets `NVM_EMBED_URL`. The CLI then opens `https://nevermined.app/cards/setup` — which doesn't exist post-cutover. The API counterpart PR explicitly avoided this:

   > "No fallback to the webapp host (that's the dead route we're steering away from)."

   Two ways to align:
   - **Drop the `NVM_FRONTEND_URL` middle step** entirely: `process.env.NVM_EMBED_URL || 'http://localhost:4250'`. Cleanest.
   - **Keep the fallback for localhost-shaped frontends only** (`localhost`, `127.0.0.1`), where same-origin embed during local dev is realistic, and require an explicit `NVM_EMBED_URL` for any non-local frontend.

   Either matches the API's "no silent fallback to a known-dead route" stance.

2. **Local default port mismatches the API side** — `src/environments.ts:66` falls back to `http://localhost:4200` (webapp port), but the API counterpart PR (#1824) uses `http://localhost:4250` for the embed in local. So a developer running both has to manually set `NVM_EMBED_URL=http://localhost:4250` for the contract to align. Cheap fix: default to `4250` here too, matching the embed app's documented local port. (If `4250` is actually wrong and the real port is something else, both PRs should change in lockstep.)

### Nice to have

3. **Adding `embed: string` as a *required* field on the public `EnvironmentInfo` interface is technically a TypeScript breaking change** — any consumer constructing their own `EnvironmentInfo` (e.g. `const env: EnvironmentInfo = { frontend, backend, proxy, heliconeUrl }`) will now fail to compile. In practice, the vast majority of consumers go through the `Environments` map and never touch the interface directly, so this is mostly cosmetic. Worth a one-line CHANGELOG note flagging "direct `EnvironmentInfo` constructors now need an `embed` field" and a minor (not patch) version bump.

### Observations (no action)

- **Pattern divergence from #1824 is fine, just noting it for future readers.** The API derives `embed` from `appUrl.replace('://', '://embed.')` (one source of truth per env). The SDK hardcodes per env + an env-var ladder for `custom`. Both are reasonable for their respective shapes; just don't try to "unify" them later — the API needs runtime derivation because it sees the webapp URL via config, while the SDK has known compile-time targets.
- `cli/test/unit/widget-redirect-flow.test.ts:78` uses `http://embed.test` as the fixture host. RFC 6761 reserves `.test`, so this is correctly unrouted — good.
- `cli/src/commands/login.ts` correctly keeps `frontendUrl` for the `/auth/cli` flow — verified the webapp still serves that route.
- The PR's narrow scope ("CLI only") is verified by grep — no SDK helpers build widget URLs at `/embed/cards/*` outside of what's already updated.

### Closing

Finding #1 is the meaningful one — the fallback chain re-enables the exact misconfiguration this PR is meant to fix. #2 is a small consistency win. The rest are FYI. Happy to take this with the fallback tightened.

r-marques (approved):

eruizgar91:
Thanks for the thorough review, @r-marques. Addressed in 30d5509:

**#1 — `custom.embed` falling back to the webapp host (the meaningful one):** fixed. Dropped the `NVM_FRONTEND_URL` middle step entirely so a custom env can never silently open the dead `/cards/*` route on the webapp host. Now:

```ts
embed: process.env.NVM_EMBED_URL || 'http://localhost:4250',
```

**#2 — local default port:** aligned to `4250` to match the embed app's local port in nvm-monorepo#1824. A developer running both no longer needs to set `NVM_EMBED_URL` by hand for the contract to line up.

**#3 — required `embed` field is a TS breaking change for direct `EnvironmentInfo` constructors:** agreed it's worth a **minor** (not patch) bump. I've left `CHANGELOG.md` and `package.json` untouched since both are regenerated/bumped by the release flow (`auto-changelog` from PR titles + `prepare-release.yml`) — flagging here so the release is dispatched as a minor with a note that "direct `EnvironmentInfo` constructors now need an `embed` field". Happy to add an explicit note elsewhere if you'd prefer it pinned in the repo.

On the observations: noted re: not unifying the SDK (compile-time targets) and API (`appUrl.replace`-derived) approaches later — kept them deliberately separate.

PR #363, opened 2026-06-01
fix(cli): #362 forward network param to the embed app redirect URL

### LINKED ISSUE #362: CLI redirect card flows omit `network` → live/staging_live flows hit the sandbox backend and fail
## Summary

The CLI redirect-mode card flows (`cards setup` / `cards enroll` / `cards delegate`) open the standalone embed app (`embed.<tier>`) **without** a `network` query param. The embed app resolves its active network — and therefore which backend it validates the widget session against — **solely** from `?network=`, defaulting to `sandbox`. As a result, any CLI card flow run against a **live** environment silently lands on the **sandbox** backend, where the live-minted session token doesn't exist, and the flow fails.

Sandbox flows work only by luck (sandbox is the default).

## Root cause

`buildEmbedUrl` in `cli/src/utils/widget-redirect-flow.ts` sets `sessionToken`, `returnUrl`, `state`, and the caller's `extra` params (e.g. `provider`) — but never `network`:

```ts
function buildEmbedUrl(opts: BuildEmbedUrlOptions): string {
  const url = new URL(opts.embedPath, opts.embedUrl)
  url.searchParams.set('sessionToken', opts.sessionToken)
  url.searchParams.set('returnUrl', opts.returnUrl)
  url.searchParams.set('state', opts.state)
  if (opts.extra) { /* provider, paymentMethodId, … */ }
  return url.toString()
}
```

On the embed-app side (nvm-monorepo `apps/embed/src/store/useEnvironmentStore.ts`), the network is fixed at store init from the URL only:

```ts
function resolveInitialEnvironment(): Environment {
  const network = new URLSearchParams(window.location.search).get('network')
  if (network === null) return 'sandbox'        // ← silent default
  if (network === 'live' || network === 'sandbox') return network
  throw new Error(`[embed] invalid "network" search param: ${network}`)
}
```

- The network is **not** derived from the `sessionToken` (the store never decodes it).
- `switchEnvironment` is a no-op, so nothing can correct the network after init.
- The chosen network selects `config.backendUrl`, against which `EmbeddedAuthGuard` validates the session token.

Because `Environments.live.embed` and `Environments.sandbox.embed` are the **same** origin (`https://embed.nevermined.app`), differentiated only by `backend`, a live CLI run opens `embed.nevermined.app/cards/setup?sessionToken=…&returnUrl=…&state=…&provider=stripe` with no `network` → embed defaults to `sandbox` → validates the token against `api.sandbox.nevermined.app` → the live-minted token is unknown there → failure. Same for `staging_live` (`embed.nevermined.dev`).

## Impact

- **All `live` and `staging_live` CLI redirect card flows are broken** (setup/enroll/delegate).
- `sandbox` / `staging_sandbox` work only because `sandbox` happens to be the default.

## Proposed fix

Have the CLI append `network` to the redirect URL, derived from the active environment name:

| Environment | `network` |
|---|---|
| `sandbox` | `sandbox` |
| `live` | `live` |
| `staging_sandbox` | `sandbox` |
| `staging_live` | `live` |
| `custom` | derive from `NVM_BACKEND_URL` (or require an explicit value) |

Simplest landing: pass `network` through the existing `extraSearchParams` in the three `cli/src/commands/cards/*` commands, or add a dedicated `network` field to `WidgetRedirectFlowOptions` / `buildEmbedUrl`. Mirrors how the iframe SDK already forwards `?network=` on mount.

## Related

- Introduced by #361 (CLI repoint to the standalone embed app, #1816).
- The embed-app cutover: nvm-monorepo #1787 / #1790.
- Docs gap tracked alongside in nevermined-io/docs#200 (redirect-URL examples omit `&network=` and the prose implies only iframe mounts need it).

eruizgar91:
Closes #362.

## Problem

The CLI redirect-mode card flows (`cards setup` / `enroll` / `delegate`) opened the standalone embed app **without** a `network` query param. The embed app resolves its active network — and therefore which backend it validates the widget session against — **solely** from `?network=`, defaulting to `sandbox` (it never decodes the session token).

Because `Environments.live.embed` and `Environments.sandbox.embed` are the **same** origin (`https://embed.nevermined.app`, differentiated only by `backend`), a `live` CLI run opened the embed with no network → embed defaulted to `sandbox` → validated the live-minted token against `api.sandbox.nevermined.app`, where it doesn't exist → **flow failed**. Same for `staging_live`. Sandbox worked only because `sandbox` is the default.

Introduced by #361 (CLI repoint to the standalone embed app).

## Fix

- `cli/src/utils/widget-redirect-flow.ts`:
  - Add a **required** `network` field to `WidgetRedirectFlowOptions`; `buildEmbedUrl` sets `?network=` **before** `extra` so a caller can't accidentally clobber it.
  - Add `resolveEmbedNetwork(environment)`:

    | Environment | `network` |
    |---|---|
    | `sandbox` / `staging_sandbox` | `sandbox` |
    | `live` / `staging_live` | `live` |
    | `custom` | sniff `NVM_BACKEND_URL` (host segment `live` → `live`, else `sandbox`) |
- Wire `network: resolveEmbedNetwork(environment)` through the three `cards/*` commands.
- Docs: document `network` as a required redirect param (with the failure-mode warning) and update the URL/diagram/pseudocode examples.
- Tests: assert the param is forwarded on the embed URL; cover `resolveEmbedNetwork` (including the `alive`-isn't-`live` boundary).

## Validation

- SDK `pnpm build`: compiles (no `src/` changes this PR).
- CLI built against the local SDK (symlink, as CI does); CI gate green: `test:unit` (10) + `test:integration` (28).
- `widget-redirect-flow` suite: **7** green (4 existing + the new network-forwarding + 2 `resolveEmbedNetwork`).
- Docs structure validation (`generate-docs.sh`) — see note: only validates structure, examples updated manually.

## Related

- Docs gap also tracked in nevermined-io/docs#200 (redirect-URL examples omit `&network=`).
- Embed-app network resolution: nvm-monorepo #1787 / #1790.

### DIFF
--- a/cli/src/utils/widget-redirect-flow.ts
+++ b/cli/src/utils/widget-redirect-flow.ts
@@ -23,6 +23,12 @@ export type EmbedNetwork = 'sandbox' | 'live'
  * `custom` has no fixed tier, so we sniff `NVM_BACKEND_URL`: a backend
  * host containing `live` selects `live`, otherwise we fall back to
  * `sandbox` (matching the embed app's own default).
+ *
+ * NOTE: `live` is only matched as a dot/slash-bounded segment (the
+ * `api.live.<host>` convention), so a hyphenated host like
+ * `https://api-live.example.com` would fall through to `sandbox`. That's
+ * intentional given the naming convention; a `custom` deployment that
+ * doesn't follow it should set `NVM_BACKEND_URL` to a conforming host.
  */
 export function resolveEmbedNetwork(environment: EnvironmentName): EmbedNetwork {
   switch (environment) {
@@ -269,14 +275,15 @@ function buildEmbedUrl(opts: BuildEmbedUrlOptions): string {
   url.searchParams.set('sessionToken', opts.sessionToken)
   url.searchParams.set('returnUrl', opts.returnUrl)
   url.searchParams.set('state', opts.state)
-  // Set BEFORE `extra` so a caller can never accidentally clobber the
-  // network the embed app keys its backend selection off of.
-  url.searchParams.set('network', opts.network)
   if (opts.extra) {
     for (const [k, v] of Object.entries(opts.extra)) {
       url.searchParams.set(k, v)
     }
   }
+  // Set AFTER `extra` so a caller can never accidentally clobber the
+  // network the embed app keys its backend selection off of —
+  // `URLSearchParams.set` overwrites, so writing it last makes it win.
+  url.searchParams.set('network', opts.network)
   return url.toString()
 }
 
--- a/cli/src/utils/widget-redirect-flow.ts
+++ b/cli/src/utils/widget-redirect-flow.ts
@@ -1,7 +1,43 @@
 import { createServer, IncomingMessage, ServerResponse } from 'http'
 import { randomBytes, timingSafeEqual } from 'crypto'
+import type { EnvironmentName } from '@nevermined-io/payments'
 
 import { openBrowser } from './browser.js'
+
+/**
+ * The network the embed app resolves its active backend from. The embed
+ * app reads `?network=` on mount and defaults to `sandbox` when absent
+ * (it never decodes the session token to infer it), so a CLI flow that
+ * omits this lands a live-minted session on the sandbox backend and
+ * fails. We always forward it explicitly. See issue #362.
+ */
+export type EmbedNetwork = 'sandbox' | 'live'
+
+/**
+ * Map a CLI environment name to the embed app's `network` value. The
+ * `embed` origin is shared across the sandbox/live pair within a tier
+ * (`embed.nevermined.app` for both `sandbox` and `live`), differentiated
+ * only by the backend the session is validated against — so the embed
+ * app cannot infer the network from the origin and we must pass it.
+ *
+ * `custom` has no fixed tier, so we sniff `NVM_BACKEND_URL`: a backend
+ * host containing `live` selects `live`, otherwise we fall back to
+ * `sandbox` (matching the embed app's own default).
+ */
+export function resolveEmbedNetwork(environment: EnvironmentName): EmbedNetwork {
+  switch (environment) {
+    case 'live':
+    case 'staging_live':
+      return 'live'
+    case 'sandbox':
+    case 'staging_sandbox':
+      return 'sandbox'
+    case 'custom':
+      return /(^|[.\/])live([.\/]|$)/.test((process.env.NVM_BACKEND_URL || '').toLowerCase())
+        ? 'live'
+        : 'sandbox'
+  }
+}
 
 const SELF_MINT_FETCH_TIMEOUT_MS = 15_000
 
@@ -38,6 +74,13 @@ export interface WidgetRedirectFlowOptions {
    * `/cards/setup` or `/cards/enroll`.
    */
   embedPath: string
+  /**
+   * Embed-app network (`sandbox` / `live`). Forwarded as `?network=` so
+   * the embed app validates the session against the matching backend —
+   * derive it from the active environment via `resolveEmbedNetwork`.
+   * Required: omitting it lets live flows silently hit sandbox (#362).
+   */
+  network: EmbedNetwork
   /**
    * Called once the local callback server is listening, with the bound
    * `returnUrl`. The caller mints a widget session against that URL and
@@ -187,6 +230,7 @@ export async function runWidgetRedirectFlow(
           sessionToken,
           returnUrl,
           state,
+          network: opts.network,
           extra: opts.extraSearchParams,
         })
 
@@ -216,6 +260,7 @@ interface BuildEmbedUrlOptions {
   sessionToken: string
   returnUrl: string
   state: string
+  network: EmbedNetwork
   extra?: Record<string, string>
 }
 
@@ -224,6 +269,9 @@ function buildEmbedUrl(opts: BuildEmbedUrlOptions): string {
   url.searchParams.set('sessionToken', opts.sessionToken)
   url.searchParams.set('returnUrl', opts.returnUrl)
   url.searchParams.set('state', opts.state)
+  // Set BEFORE `extra` so a caller can never accidentally clobber the
+  // network the embed app keys its backend selection off of.
+  url.searchParams.set('network', opts.network)
   if (opts.extra) {
     for (const [k, v] of Object.entries(opts.extra)) {
       url.searchParams.set(k, v)
--- a/cli/src/commands/cards/delegate.ts
+++ b/cli/src/commands/cards/delegate.ts
@@ -4,6 +4,7 @@ import { BaseCommand } from '../../base-command.js'
 import { resolveOrgIdInteractive } from '../../utils/orgs.js'
 import {
   mintSelfWidgetSession,
+  resolveEmbedNetwork,
   runWidgetRedirectFlow,
 } from '../../utils/widget-redirect-flow.js'
 
@@ -67,6 +68,7 @@ export default class CardsDelegate extends BaseCommand {
       const result = await runWidgetRedirectFlow({
         embedUrl: env.embed,
         embedPath: '/cards/delegate',
+        network: resolveEmbedNetwork(environment),
         mintSession: async ({ returnUrl }) => {
           const session = await mintSelfWidgetSession({
             backendUrl: env.backend,
--- a/cli/src/commands/cards/enroll.ts
+++ b/cli/src/commands/cards/enroll.ts
@@ -4,6 +4,7 @@ import { BaseCommand } from '../../base-command.js'
 import { resolveOrgIdInteractive } from '../../utils/orgs.js'
 import {
   mintSelfWidgetSession,
+  resolveEmbedNetwork,
   runWidgetRedirectFlow,
 } from '../../utils/widget-redirect-flow.js'
 
@@ -68,6 +69,7 @@ export default class CardsEnroll extends BaseCommand {
       const result = await runWidgetRedirectFlow({
         embedUrl: env.embed,
         embedPath: '/cards/enroll',
+        network: resolveEmbedNetwork(environment),
         mintSession: async ({ returnUrl }) => {
           const session = await mintSelfWidgetSession({
             backendUrl: env.backend,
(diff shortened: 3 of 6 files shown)


r-marques (commented):
Direct follow-up to the gap I flagged on docs#200 — good that you tracked it down to the embed-app's "never decodes the token, just reads `?network=` and defaults to `sandbox`" behavior. The fix is clean overall: a required `network` field on the helper (type-system catches omissions), a `resolveEmbedNetwork` mapping for the four named envs plus a heuristic for `custom`, and `network` set in `buildEmbedUrl` so all three CLI commands get it uniformly.

One real bug, otherwise solid.

### Required change — anti-clobber promise doesn't hold

`cli/src/utils/widget-redirect-flow.ts:269–276`:

```ts
url.searchParams.set('state', opts.state)
// Set BEFORE `extra` so a caller can never accidentally clobber the
// network the embed app keys its backend selection off of.
url.searchParams.set('network', opts.network)
if (opts.extra) {
  for (const [k, v] of Object.entries(opts.extra)) {
    url.searchParams.set(k, v)
  }
}
```

The comment promises "a caller can never accidentally clobber `network`", but `URLSearchParams.set(key, val)` *overwrites* on each call — so setting `network` **before** the `extra` loop means a caller passing `extraSearchParams: { network: 'sandbox' }` will silently overwrite `'live'`. The order is the opposite of what the comment intends.

Two ways to fix:

```ts
// (a) Set network AFTER the extra loop — network always wins.
if (opts.extra) {
  for (const [k, v] of Object.entries(opts.extra)) {
    url.searchParams.set(k, v)
  }
}
url.searchParams.set('network', opts.network)
```

or

```ts
// (b) Filter network out of extra so the intent is explicit at the loop.
if (opts.extra) {
  for (const [k, v] of Object.entries(opts.extra)) {
    if (k === 'network') continue
    url.searchParams.set(k, v)
  }
}
url.searchParams.set('network', opts.network)
```

(a) is the minimal fix; (b) is slightly more defensive and makes the intent grep-able. Either way, this is the exact behavior the regression is meant to prevent: a live session silently dropping to sandbox.

In practice the bug doesn't bite today because none of the three `cards/*` commands pass `extra`, but the comment claims a contract the code doesn't deliver, and any future command that uses `extraSearchParams` will silently regress the fix.

### Nice to have

1. **Add a regression test for the anti-clobber claim** — `cli/test/unit/widget-redirect-flow.test.ts`. Something like:

   ```ts
   test('network from opts wins over an attempted override in extraSearchParams', async () => {
     const flow = startFlow({ network: 'live', extraSearchParams: { network: 'sandbox' } })
     await flow.returnUrlPromise
     const embedUrl = flow.logs.find((l) => l.startsWith('http://embed.test')) as string
     expect(new URL(embedUrl).searchParams.get('network')).toBe('live')
   })
   ```

   This would have caught the bug, and locks the contract going forward. Note `startFlow` would need to forward `extraSearchParams` to `runWidgetRedirectFlow`.

2. **Edge case worth flagging in code (or a JSDoc TODO)**: the regex `/(^|[.\/])live([.\/]|$)/` only matches `live` as a **dot-or-slash-bounded** segment. A future `custom` setup with a hyphen-separated host like `https://api-live.example.com/` would silently fall through to `sandbox`. Not a real concern today — the convention is `api.live.<host>` — but worth a one-liner in the JSDoc of `resolveEmbedNetwork` (or one more test) so it's not a surprise. If you want stronger insurance, expose a `NVM_NETWORK` env var that overrides the heuristic; not blocking either way.

### Observations (no action)

- The boundary regex is well-designed — `alive`, `livestream`, `survive`, and the `example.com/aliveness` shapes are all correctly excluded. The dedicated `'alive'` test case at `widget-redirect-flow.test.ts:213` is a nice tell that you thought about it.
- Forcing `network: EmbedNetwork` as a *required* field on `WidgetRedirectFlowOptions` is the right call — it turns "did I remember to forward network?" from a runtime question into a compile-time one. Cheaper than a runtime check.
- `process.env.NVM_BACKEND_URL || ''` cleanly handles the unset case (regex against empty string → no match → `sandbox`, which matches the embed app's own default). Good defensive default.
- Closes #362 cleanly. The `markdown/cli-card-setup.md` warning is the right level of friction ("Always set it explicitly" with the failure mode spelled out).
- Cross-PR follow-up note: docs#200 still doesn't include `?network=` in its example URLs (`organization-widgets.mdx:330`). That should be updated alongside this — happy to flag it on #200 if you want, or you can fold it into a follow-up.

### Closing

The anti-clobber bug is the only thing blocking my approval — small fix. Everything else is solid.

cli/src/utils/widget-redirect-flow.ts
```diff
@@ -224,6 +269,9 @@ function buildEmbedUrl(opts: BuildEmbedUrlOptions): string {
   url.searchParams.set('sessionToken', opts.sessionToken)
   url.searchParams.set('returnUrl', opts.returnUrl)
   url.searchParams.set('state', opts.state)
+  // Set BEFORE `extra` so a caller can never accidentally clobber the
+  // network the embed app keys its backend selection off of.
+  url.searchParams.set('network', opts.network)
```
aaitor: **MEDIUM — The clobber-guard is inverted: `network` is set *before* the `extra` loop, so an `extra` key named `network` would overwrite it — the opposite of what the comment promises.**

`url.searchParams.set('network', …)` runs first, then the `extra` loop calls `set(k, v)` for each entry. `URLSearchParams.set` replaces, so `extra.network` (if ever present) wins — and the comment "Set BEFORE `extra` so a caller can never accidentally clobber the network" claims the inverse of the actual behavior. It's latent today (all three commands pass `network` as a top-level field and put only `provider`/`paymentMethodId` in `extraSearchParams`), but this is the exact money-path param the PR exists to protect: if any future caller drops `network` into `extraSearchParams`, #362 silently regresses with a misleading comment vouching for safety. Set it *after* the loop so it genuinely can't be clobbered:

```ts
function buildEmbedUrl(opts: BuildEmbedUrlOptions): string {
  const url = new URL(opts.embedPath, opts.embedUrl)
  url.searchParams.set('sessionToken', opts.sessionToken)
  url.searchParams.set('returnUrl', opts.returnUrl)
  url.searchParams.set('state', opts.state)
  if (opts.extra) {
    for (const [k, v] of Object.entries(opts.extra)) {
      url.searchParams.set(k, v)
    }
  }
  // Set AFTER `extra` so a caller can never accidentally clobber the
  // network the embed app keys its backend selection off of.
  url.searchParams.set('network', opts.network)
  return url.toString()
}
```

And lock it in with a one-line regression test alongside the existing `#362` forwarding test:
```ts
test('extraSearchParams cannot clobber the network param', async () => {
  const flow = startFlow({ network: 'live', extra: { network: 'sandbox' } })
  // ...drive the flow...
  expect(new URL(embedUrl).searchParams.get('network')).toBe('live')
})
```
(Your `startFlow` helper would need to thread `extra` through to `extraSearchParams` for that case.) Non-blocking — the #362 fix works correctly as shipped; this just makes the guard match its own comment.

aaitor (approved):
Correct, well-reasoned fix for #362 — approving. The diagnosis matches what I independently verified in the embed app (nvm-monorepo #1790): `resolveInitialEnvironment` reads `?network=`, defaults to `sandbox` when absent, and `sandbox`/`live` share the `embed.nevermined.app` origin (differentiated only by backend), so a `live` run with no `network` param really did validate against the sandbox backend and fail. `resolveEmbedNetwork` maps the environment correctly (incl. `staging_live → live`), the param is forwarded, and the docs now warn about the failure mode. One MEDIUM hardening note — the clobber-guard is inverted vs. its comment, latent today but on the exact param this PR exists to protect.

**Verification I ran**
- Read the full `widget-redirect-flow.ts`, all three `cards/*` command wirings, the test, and the docs diff at HEAD `3a22326`.
- Confirmed `EnvironmentName` is exactly `staging_sandbox | staging_live | sandbox | live | custom` — so `resolveEmbedNetwork`'s switch is exhaustive (and CI tsc proves it: no trailing `return`, so a non-exhaustive switch wouldn't compile).
- `resolveEmbedNetwork('custom')` regex `/(^|[.\/])live([.\/]|$)/` handles the documented `alive`-isn't-`live` boundary correctly; empty `NVM_BACKEND_URL` → `sandbox`.
- ESM `.js` extensions present on all relative imports (`./browser.js`, the command imports); no internal `staging_*` enum names leak into the user-facing docs (it uses `sandbox`/`live`).
- Confirmed **none** of the three commands pass `network` inside `extraSearchParams` (all pass it as a top-level field), so the inverted clobber-guard below is latent, not a live bug.
- Cross-repo premise check against nvm-monorepo #1790's `resolveInitialEnvironment` — holds.
- Given the focused scope (6 files, +105/−3, money-adjacent but small), I reviewed directly across the quality / security / backward-compat / test lenses rather than dispatching the full async agent panel. CI: `Unit & Integration`, `Lint & Build`, `CLI Sync Check` green; `E2E` pending (let it land before merge).

_Release process for `payments` is its own convention — not triggering anything._

eruizgar91:
Thanks @r-marques and @aaitor — both nailed the same real issue. Fixed in 748b485:

**Required change — inverted anti-clobber guard:** `network` is now written **after** the `extra` loop in `buildEmbedUrl`, so it genuinely wins (since `URLSearchParams.set` overwrites). Comment corrected to match. Went with option (a) — minimal and the loop stays generic.

**Nice-to-have #1 — regression test:** added `'network from opts wins over an attempted override in extraSearchParams'`, threading `extraSearchParams` through the test's `startFlow`. Locks the contract — it fails against the old before-the-loop ordering.

**Nice-to-have #2 — hyphenated-host edge case:** added a JSDoc note on `resolveEmbedNetwork` spelling out that `live` only matches as a dot/slash-bounded segment, so `api-live.example.com` falls through to `sandbox` by design (the convention is `api.live.<host>`). Left the `NVM_NETWORK` override out for now to keep scope tight — happy to add it as a follow-up if you'd like the stronger insurance.

`widget-redirect-flow` suite now 8 green.

On the **docs#200 follow-up** (`organization-widgets.mdx:330` still omits `?network=`): good catch, that should be updated alongside. I'll fold it into the docs#200 branch separately so this PR stays scoped to the CLI fix.

PR #365, opened 2026-06-02
feat: add provider filter to listPaymentMethods

r-marques:
Part of nevermined-io/nvm-monorepo#1549. Backend: nevermined-io/nvm-monorepo#1715 (must land first).

Adds an optional `provider` to `ListOptions`; forwarded as the `?provider=` query param to the backend list-payment-methods endpoint.

### DIFF
--- a/tests/unit/delegation-list-payment-methods-filter.test.ts
+++ b/tests/unit/delegation-list-payment-methods-filter.test.ts
@@ -9,9 +9,13 @@
 import { Payments } from '../../src/payments.js'
 import type { PaymentMethodSummary } from '../../src/x402/delegation-api.js'
 
+// The test mocks global.fetch (installFetch), so this key is never transmitted.
+// It only has to satisfy `decodeJwt` in Payments.getInstance (reads `sub`/`o11y`,
+// no signature/expiry check), so the fallback is a structurally-valid JWT with
+// dummy claims — a zero-address subject and a fake signature — NOT a live token.
 const TEST_API_KEY =
   process.env.TEST_PROXY_BEARER_TOKEN ||
-  'sandbox-staging:eyJhbGciOiJFUzI1NksifQ.eyJpc3MiOiIweDU4MzhCNTUxMmNGOWYxMkZFOWYyYmVjY0IyMGViNDcyMTFGOUIwYmMiLCJzdWIiOiIweDIxRjc5ZjlkM2I2ZDUyZUY4Y2M4QjFhN0YyNjFCY2Y1ZjJFRjM1NGEiLCJqdGkiOiIweGUxMjIwMmRkMzZlZmQ4N2FkMjE1MmRlMjlkM2MwNmE5ZDU5N2M4NWJhOGMxOTQ1YjQ5MjlkYTYyYTRiZjQ1NGYiLCJleHAiOjE3OTEwNDc0OTcsIm8xMXkiOiJzay1oZWxpY29uZS13amUzYXdpLW5ud2V5M2EtdzdndnY3YS1oYmh3bm1pIn0.JI14qfSWHCWRvHOK9TAg3HEXWX7oKEI6fU6gaaWlyDl5btBWLh8FQo1ZnuzixPmgsUR3gc4oRlenLPUuTy-mORw'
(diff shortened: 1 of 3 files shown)


tests/unit/delegation-list-payment-methods-filter.test.ts
```diff
@@ -0,0 +1,105 @@
...
+ * The SDK forwards the provider as a `?provider=` query string when set, and
+ * omits it entirely otherwise (preserving the default "all methods" behaviour).
+ * Depends on the backend `?provider=` filter (nevermined-io/nvm-monorepo#1715).
+ */
+import { Payments } from '../../src/payments.js'
+import type { PaymentMethodSummary } from '../../src/x402/delegation-api.js'
+
+const TEST_API_KEY =
+  process.env.TEST_PROXY_BEARER_TOKEN ||
+  'sandbox-staging:eyJhbGciOiJFUzI1NksifQ.eyJpc3MiOiIweDU4MzhCNTUxMmNGOWYxMkZFOWYyYmVjY0IyMGViNDcyMTFGOUIwYmMiLCJzdWIiOiIweDIxRjc5ZjlkM2I2ZDUyZUY4Y2M4QjFhN0YyNjFCY2Y1ZjJFRjM1NGEiLCJqdGkiOiIweGUxMjIwMmRkMzZlZmQ4N2FkMjE1MmRlMjlkM2MwNmE5ZDU5N2M4NWJhOGMxOTQ1YjQ5MjlkYTYyYTRiZjQ1NGYiLCJleHAiOjE3OTEwNDc0OTcsIm8xMXkiOiJzay1oZWxpY29uZS13amUzYXdpLW5ud2V5M2EtdzdndnY3YS1oYmh3bm1pIn0.JI14qfSWHCWRvHOK9TAg3HEXWX7oKEI6fU6gaaWlyDl5btBWLh8FQo1ZnuzixPmgsUR3gc4oRlenLPUuTy-mORw'
```
aaitor: **MEDIUM — Hardcoded *live* sandbox-staging JWT (embeds a Helicone `sk-` key) as the test fallback.**

This matches the repo's existing fixture convention (`pay-as-you-go.test.ts`, `payments.test.ts`, `x402-visa.test.ts` all use `TEST_PROXY_BEARER_TOKEN || '<hardcoded JWT>'`), so it isn't *introduced* here — but two concrete asks:

1. **This test mocks `global.fetch`** (the `installFetch` helper), so the token is never transmitted — its value is irrelevant to every assertion. Use a placeholder instead of committing another live credential:
```ts
const TEST_API_KEY = process.env.TEST_PROXY_BEARER_TOKEN || 'sandbox-staging:test-token'
```
2. **The decoded payload carries `o11y: "sk-helicone-wje3awi-…"` (a live Helicone API key) and `exp` 2026-10-03 (unexpired).** That `sk-` key is a real third-party secret sitting in a published SDK repo — it should be **rotated**, and the live JWTs pulled from all four fixtures. That's a repo-wide security follow-up (not a blocker for this PR), but worth filing now.

aaitor (approved):
Clean SDK-side forwarder for the `?provider=` filter — backward-compatible (additive optional on `ListOptions`), correctly typed (`DelegationProvider = CardProvider | 'erc4337'`, matching the backend's four providers), ESM `.js` extensions intact, and well covered (set / unset / `accessible`-combo / `erc4337`). Approving. One MEDIUM on a committed test credential (inline) plus a release-ordering note; neither gates the merge.

**Verification I ran**
- Read all three files at HEAD `857cc60`. `DelegationProvider` resolves to the 4-provider union via `../common/types.js`; `listPaymentMethods` only sets `?provider=` when truthy (omits otherwise → non-breaking default); `.js` extensions present on every relative import (incl. the test).
- Reviewed directly across the quality / security / backward-compat / test lenses (3-file, +128 SDK forwarder, CI green) rather than the full async panel.
- Decoded the test's fallback JWT to confirm the embedded secret before flagging it (it's real — see inline).
- CI: `Lint & Build`, `Unit & Integration`, `E2E`, `CLI Sync Check` all green.

**Release ordering (cross-repo):** the backend `?provider=` filter (nvm-monorepo #1843, i.e. #1715) is still open/unmerged. Until it's deployed to the environment a caller targets, an older backend silently ignores the unknown query param and returns **all** methods — so `listPaymentMethods({ provider })` is an unfiltered no-op, not an error. Don't ship/announce this SDK's provider filter ahead of the backend (the PR body already flags the dependency); pair the release with #1843 + payments-py #208.

PR #370, opened 2026-06-04
feat(langchain): add lastSettlement + createPaidReactAgent helpers

r-marques:
## Description

Additive TypeScript parity for the two LangChain helpers `payments-py` shipped in 1.7.0 ([payments-py#183](https://github.com/nevermined-io/payments-py/pull/183)), so TypeScript / LangChain.js / LangGraph.js users get the same minimal-code, discovery-first experience as Python users. No changes to any existing exported API.

New exports under `@nevermined-io/payments/langchain`:

- **`lastSettlement(): SettlePermissionsResult | undefined`** — reads a module-level slot the existing `requiresPayment` wrapper now updates on settle. LangGraph copies `RunnableConfig.configurable` per node, so the in-place `payment_settlement` write is invisible to the buyer's outer scope; the slot is the workaround. (Mirrors Python `last_settlement()`.)
- **`createPaidReactAgent(model, tools, options?)`** (new `src/x402/langchain/agent.ts`) — wraps `createReactAgent` from `@langchain/langgraph/prebuilt` with a `ToolNode(tools, { handleToolErrors: false })` so `PaymentRequiredError` propagates back to `agent.invoke()`'s caller with its `X402PaymentRequired` payload intact (the default `ToolNode` would stringify it into a `ToolMessage` and lose the payload, breaking x402 discovery). `@langchain/langgraph` is imported **lazily** (so the helper is `async`) keeping the peer-dependency story unchanged.
- **Decorator JSDoc** — clarifies fixed-vs-range plan credit semantics (mirrors the Python docstring update per nevermined-io/nvm-monorepo#1568). Comment-only.

### Notes for reviewers

- **JS `createReactAgent` is keyword-based** (`{ llm, tools }`, param named `llm` not `model`) — the helper keeps the Python-parity positional public API `(model, tools, options?)` and maps `model→llm` internally.
- **`@langchain/langgraph` pinned to `1.2.0`** (dev) + added as an **optional `peerDependency`** (`>=1.0.0 <2`). langgraph 1.3.x needs `@langchain/core ^1.1.44`, but this repo resolves core to 1.1.40 (pnpm override + `langchain@0.3.37`), so 1.2.0 is the compatible line.
- **The 2 `createPaidReactAgent` tests mock `@langchain/langgraph/prebuilt`** with faithful `ToolNode` / `createReactAgent` doubles: Jest cannot load the real LangGraph ESM tree through the helper's native dynamic `import()` (only ESM-only transitive dep is `uuid@14`). The doubles encode the real contract verified by runtime probe; the real types are still checked at build time via `agent.ts`'s `typeof import(...)` references. `tests/jest.config.json` is unchanged.

Part of nevermined-io/nvm-monorepo#1703
Closes nevermined-io/nvm-monorepo#1717

Docs PR (separate repo): companion changes to `docs/integrate/add-to-your-agent/langchain.mdx` (TS Tabs) + the `nevermined-payments` skill reference land in a `nevermined-io/docs` PR (linked below once opened).

## Types of changes

## Checklist:

- [x] Tests Cover Changes
- [x] Documentation

#### Funny gif

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
   },
   "peerDependencies": {
     "@langchain/core": ">=0.3.0",
-    "@langchain/langgraph": ">=1.0.0",
+    "@langchain/langgraph": ">=1.0.0 <2",
     "@modelcontextprotocol/sdk": ">=1.25.0",
     "express": ">=4.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@babel/core": "^7.27.4",
     "@babel/preset-env": "^7.27.2",
     "@google-cloud/aiplatform": "^6.1.0",
+    "@langchain/langgraph": "1.2.0",
     "@modelcontextprotocol/sdk": "^1.24.3",
     "@types/express": "4.17.23",
     "@types/jest": "^29.5.13",
@@ -92,9 +93,10 @@
     "typescript": "^5.3.3"
   },
   "peerDependencies": {
+    "@langchain/core": ">=0.3.0",
+    "@langchain/langgraph": ">=0.2.0",
     "@modelcontextprotocol/sdk": ">=1.25.0",
-    "express": ">=4.0.0",
-    "@langchain/core": ">=0.3.0"
+    "express": ">=4.0.0"
   },
   "peerDependenciesMeta": {
     "@modelcontextprotocol/sdk": {
@@ -104,6 +106,9 @@
       "optional": true
     },
     "@langchain/core": {
+      "optional": true
+    },
+    "@langchain/langgraph": {
       "optional": true
     }
   },
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
   },
   "peerDependencies": {
     "@langchain/core": ">=0.3.0",
-    "@langchain/langgraph": ">=0.2.0",
+    "@langchain/langgraph": ">=1.0.0",
     "@modelcontextprotocol/sdk": ">=1.25.0",
     "express": ">=4.0.0"
   },
--- a/src/x402/langchain/decorator.ts
+++ b/src/x402/langchain/decorator.ts
@@ -14,7 +14,7 @@
  *
  * The `credits` option accepts two forms:
  *   - **Static number**: `credits: 1` — always charges 1 credit
- *   - **Function**: `credits: (ctx) => Math.max(1, ctx.result.length / 100)` — dynamic
+ *   - **Function**: `credits: (ctx) => Math.max(1, Math.floor(ctx.result.length / 100))` — dynamic
  *
  * When `credits` is a function, it receives `{ args, result }` after tool execution.
  *
@@ -134,8 +134,12 @@ export interface PaymentContext {
 // `config.configurable.payment_settlement` is not visible to the buyer's outer
 // scope. A module-level slot is the simplest reliable signal. It is
 // intentionally single-tenant — if the same process runs multiple concurrent
-// settlements, the last writer wins. For multi-tenant use cases, surface the
-// receipt via a callback or via observability (see Sprint 1 of the LangChain epic).
+// settlements, the last writer wins. This race is not limited to multi-tenant
+// servers: a single `createPaidReactAgent` run can also hit it, because a ReAct
+// agent may execute several paid tools in parallel within one LLM turn via
+// `ToolNode`, and this single slot only retains the last writer. For
+// multi-tenant or parallel-tool use cases, surface the receipt via a callback
+// or via observability (see Sprint 1 of the LangChain epic).
 let lastSettlementReceipt: SettlePermissionsResult | undefined
 
 /**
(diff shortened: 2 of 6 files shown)


r-marques:
Companion docs PR: nevermined-io/docs#212 (TS Tabs in `langchain.mdx` + `## TypeScript` skill reference).

src/x402/langchain/decorator.ts
```diff
@@ -263,6 +305,10 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
         agentRequestId: paymentContext.agentRequestId,
       })
       storeInConfigurable(config, 'payment_settlement', settlement)
+      // Also publish to the module-level slot so lastSettlement() can recover
+      // the receipt — LangGraph copies config.configurable per node, hiding the
+      // line above from the buyer's outer scope.
+      lastSettlementReceipt = settlement
```
eruizgar91, line 327: **`lastSettlement()` returns a stale receipt — not `undefined` — when a settle fails after a prior success.**

The slot is only *written* on success (this line) and never reset. But the JSDoc on `lastSettlement()` claims it "Returns `undefined` … if the most recent invocation raised before reaching the settle phase." That's false once a prior settle has happened:

1. Tool A settles → `slot = receipt A`
2. Tool B's settle throws (swallowed by the `catch` with `console.error`) → slot is still receipt A
3. Caller invokes `lastSettlement()` expecting `undefined` (per the docs) and gets **receipt A**, attributing A's `transaction`/`creditsRedeemed` to invocation B.

The test *"stays undefined when settlement throws"* only covers the no-prior-settle case, which gives false confidence. Suggest resetting `lastSettlementReceipt = undefined` at the start of each invocation (or right before settle) and fixing the JSDoc.

src/x402/langchain/decorator.ts
```diff
@@ -117,6 +129,36 @@ export interface PaymentContext {
   agentRequest?: unknown
 }
 
+// Module-level holder for the most recent settlement receipt. LangGraph copies
```
eruizgar91, line 132: The "last-writer-wins" caveat is scoped to *multi-tenant* processes, but the race also bites a **single** `createPaidReactAgent` run: a ReAct agent can execute several tool calls in parallel within one LLM turn via `ToolNode`. If two paid tools settle concurrently in the same run, this module-level slot is overwritten and `lastSettlement()` returns only one of them — the exact use case this helper enables. Worth broadening the warning to cover intra-run parallelism, not just multi-tenant.

package.json
```diff
@@ -92,9 +93,10 @@
     "typescript": "^5.3.3"
   },
   "peerDependencies": {
+    "@langchain/core": ">=0.3.0",
+    "@langchain/langgraph": ">=1.0.0",
```
eruizgar91: Minor: `">=1.0.0"` is looser than the line this PR actually validates against. The PR description itself notes that langgraph 1.3.x requires `@langchain/core ^1.1.44` while this repo resolves core to 1.1.40, so 1.2.0 is the compatible line. `>=1.0.0` also has no upper bound (a future 2.x could drop `createReactAgent`). Consider bounding it (e.g. `>=1.0.0 <2`) or documenting the recommended pairing. (Also: the PR body says `>=0.2.0` for this peer — small description/code mismatch.)

eruizgar91 (commented):
Review of the LangChain helpers — overall solid, additive, and well-tested. A few points worth addressing (inline comments below), plus one that falls outside the diff hunks:

**Dynamic credits that return a non-integer silently drop the settle** (`decorator.ts:304`). `finalCredits` comes from `CreditsCallable: (ctx) => number`, which allows floats. `BigInt(1.5)` throws `RangeError`, which is caught by the settle `try/catch` and swallowed with `console.error` → credits are never burned and the caller never finds out. This isn't hypothetical: the JSDoc example at `decorator.ts:17` (`credits: (ctx) => Math.max(1, ctx.result.length / 100)`) produces a float, while the other example at line 182 uses `Math.floor`. Suggest `BigInt(Math.floor(finalCredits))` (or validating/rejecting non-integers) and making the two doc examples consistent. Pre-existing, but the PR re-touches this function.

Nice touches: the `llm`/`tools` guard with its dedicated test, the lazy `import()` keeping the peer optional, payload-pinning assertions in the tests, and the honest `CAVEAT`/`TODO(#1717)` about the unverified test doubles.

r-marques:
Thanks for the review @eruizgar91 — all four addressed in 95326a8:

- **Float credits → silent settle drop:** `maxAmount` is now `BigInt(Math.floor(finalCredits))`, and the JSDoc example that produced a float now floors (consistent with the other example), so a non-integer credits value can no longer throw `RangeError` into the swallowed settle `catch`.
- **Stale `lastSettlement()`:** the receipt slot is now reset to `undefined` at the **start of each invocation** (before verify), so ANY pre-settle failure — a verify `PaymentRequiredError` OR a swallowed settle failure — yields `undefined`, matching the JSDoc. Added two tests: prior-success → failed-settle → `undefined`, and prior-success → verify-failure → `undefined`.
- **Concurrency caveat** broadened to cover intra-run parallelism (a single `createPaidReactAgent` run can settle several `ToolNode` tools in parallel within one LLM turn; the module slot keeps only one).
- **Peer floor** → `>=1.0.0 <2`; the stale `>=0.2.0` mention in the PR body is corrected.

`pnpm build` + `test:unit` (177 passed, incl. the 2 new tests) + `lint` green.

eruizgar91 (approved):

PR #371, opened 2026-06-04
feat(langsmith): emit nvm:verify/nvm:settlement spans from requiresPayment (TS-1)

r-marques:
## Description

TS parity for Sprint 1 observability: the TypeScript SDK now emits the cross-SDK
`nvm:verify` / `nvm:settlement` LangSmith spans around the x402 verify→work→settle
lifecycle, matching the Python SDK **attribute-for-attribute** per the
[observability-spans-v1 contract](https://github.com/nevermined-io/nvm-monorepo/blob/main/docs/specs/observability-spans-v1.md)
(nvm-monorepo#1746, spec PR nevermined-io/nvm-monorepo#1882). Identical spans let a single
LangSmith trace be correlated across a TypeScript buyer and a Python seller — e.g. filter on `nvm.tx_hash`.

> **Stacked on TS-0 (#370).** Base is `feat/ts0-langchain-helpers`, not `main` — this PR
> and TS-0 both edit `src/x402/langchain/decorator.ts`. Merges **after** #370; retarget base→main
> once TS-0 lands (GitHub usually auto-retargets). Review only the delta on top of TS-0.

### Design — Option B (Python-parity), user-confirmed

The original issue framing assumed a `NeverminedTracer extends BaseCallbackHandler`. **That class does
not exist in payments-py** (verified). Sprint 1's Python model is *imperative* spans opened by the
`@requires_payment` decorator (`verify_span` / `settlement_span` in `payments_py/langsmith/spans.py`),
not a passive callback handler. Per the user decision (2026-06-04), this PR ports that model: span
helpers wired into the existing TS `requiresPayment` decorator around `verifyPermissions` /
`settlePermissions`, mirroring Python's decorator.

### What's here

- **`src/x402/langsmith/spans.ts`** — `verifySpan` / `settlementSpan` (langsmith JS `RunTree.createChild`,
  `run_type: "tool"`, names exactly `nvm:verify` / `nvm:settlement`), `buildVerifyMetadata` /
  `buildSettleMetadata` (exact `nvm.*` set per spec; `nvm.credits_redeemed` & `nvm.balance.after` stay
  **strings**; `nvm.payment_token` joiner is **U+2026** `…`; `nvm.tx_hash`←`transaction`,
  `nvm.balance.after`←`remainingBalance`), `abbreviateToken` (redact+warn per #1747 / payments-py#217:
  ≤20-char tokens redacted to `<first4>…(short)` + warn, idempotent; empty stays silent),
  `redactMetadataKeys`, `activeRunTree`, `addMetadata`.
- **`src/x402/langsmith/index.ts`** + a `./langsmith` subpath export.
- **`decorator.ts` wiring** — redacts the full `payment_token` from the parent run tree **before**
  opening the verify span; two-pass verify metadata (static pre-verify so unpaid discovery probes are
  still labeled, then results + timing); settlement span around the settle call. No-op when tracing is
  off or `langsmith` is absent.
- **Packaging** — `langsmith` declared as an **optional peer dependency** (consistent with
  `@langchain/core`), imported **lazily** so non-observability users aren't forced to install it.

### Tests

26 new unit tests (`tests/unit/x402/langsmith-spans.test.ts`, `langsmith-decorator.test.ts`): span
names + attributes per spec, types preserved as strings, redact+warn token defense, full token never
shipped, parent-token redaction, no-op when tracing disabled. Full unit suite green: **201 passed**.

Caveat (honest, mirrors TS-0's langgraph tests in #1717): the real `langsmith` JS SDK is replaced with
a faithful `RunTree` double via `jest.mock` — the repo's Jest setup doesn't transform `node_modules`
(langsmith pulls in ESM-only `uuid@14`) and `tsc` excludes `tests/`, so these unit tests assert OUR
wiring, not langsmith's transport. The double's shape was taken from `langsmith@0.7.4`. A real-SDK
integration test is deferred to the tutorial follow-up.

### Out of scope (human closeout — NOT in this PR)

The TS variant tutorial (sibling of `langchain-paid-agent-py-with-langsmith`) and the demo video — an
agent can't record a video. Ships the SDK module + tests + docs; the tutorial/video remain for a human.

### Docs

Separate docs PR updates `docs/integrate/add-to-your-agent/langchain.mdx` "Trace payments in LangSmith"
section to cover TypeScript alongside Python (the api-reference tree is generated on release via
update-docs.yml). Link: _added in a follow-up comment._

Part of epic nevermined-io/nvm-monorepo#1703
Closes nevermined-io/nvm-monorepo#1709
Spec: nevermined-io/nvm-monorepo#1882 · Stacked on TS-0: #370

## Types of changes

## Checklist:

- [x] Follows the code style of this project. (`pnpm lint` + prettier clean)
- [x] Tests Cover Changes (26 new unit tests; full unit suite 201 passed)
- [x] Documentation (separate docs PR; `./langsmith` subpath + JSDoc)

#### Funny gif

tracing

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,10 @@
     "./langchain": {
       "types": "./dist/x402/langchain/index.d.ts",
       "import": "./dist/x402/langchain/index.js"
+    },
+    "./langsmith": {
+      "types": "./dist/x402/langsmith/index.d.ts",
+      "import": "./dist/x402/langsmith/index.js"
     }
   },
   "scripts": {
@@ -80,6 +84,7 @@
     "eslint-plugin-tsdoc": "^0.5.2",
     "jest": "^29.7.0",
     "langchain": "^0.3.37",
+    "langsmith": "^0.7.4",
     "openai": "^6.2.0",
     "prettier": "^3.2.5",
     "source-map-support": "^0.5.21",
@@ -96,7 +101,8 @@
     "@langchain/core": ">=0.3.0",
     "@langchain/langgraph": ">=1.0.0 <2",
     "@modelcontextprotocol/sdk": ">=1.25.0",
-    "express": ">=4.0.0"
+    "express": ">=4.0.0",
+    "langsmith": ">=0.3.0"
   },
   "peerDependenciesMeta": {
     "@modelcontextprotocol/sdk": {
@@ -109,6 +115,9 @@
       "optional": true
     },
     "@langchain/langgraph": {
+      "optional": true
+    },
+    "langsmith": {
       "optional": true
     }
   },
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "@langchain/langgraph": ">=1.0.0 <2",
     "@modelcontextprotocol/sdk": ">=1.25.0",
     "express": ">=4.0.0",
-    "langsmith": ">=0.3.0"
+    "langsmith": ">=0.7.0"
   },
   "peerDependenciesMeta": {
     "@modelcontextprotocol/sdk": {
--- a/src/x402/langchain/decorator.ts
+++ b/src/x402/langchain/decorator.ts
@@ -50,6 +50,15 @@ import {
   type VerifyPermissionsResult,
   type SettlePermissionsResult,
 } from '../facilitator-api.js'
+import {
+  activeRunTree,
+  addMetadata,
+  buildSettleMetadata,
+  buildVerifyMetadata,
+  redactMetadataKeys,
+  settlementSpan,
+  verifySpan,
+} from '../langsmith/spans.js'
 
 /**
  * Context passed to a dynamic credits function after tool execution.
@@ -187,7 +196,6 @@ function storeInConfigurable(config: unknown, key: string, value: unknown): void
 
   const configurable = (config as Record<string, unknown>).configurable
   if (configurable == null || typeof configurable !== 'object') return
-
   ;(configurable as Record<string, unknown>)[key] = value
 }
 
@@ -253,12 +261,49 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
       agentId,
       network,
     })
+    // The scheme/network the verify span advertises come from the resolved
+    // X402 scheme so the span metadata matches what the buyer paid against.
+    const accepted = paymentRequired.accepts[0]
+    const planIds = paymentRequired.accepts
+      .map((a) => a.planId)
+      .filter((p): p is string => Boolean(p))
+    const resolvedScheme = accepted?.scheme
+    const resolvedNetwork = network ?? accepted?.network
+
+    // LangChain auto-captures every key in config.configurable into the parent
+    // tool span's metadata, and child spans inherit it at construction time.
+    // Strip the full x402 access token from the parent BEFORE opening the verify
+    // span so neither the parent nor the child carries the raw credential — only
+    // the abbreviated nvm.payment_token remains for correlation.
+    const parentRunTree = await activeRunTree()
+    redactMetadataKeys(parentRunTree, 'payment_token')
+
+    const verifyStarted = Date.now()
+    // Open the verify span BEFORE the token-presence check so failed probes
+    // (no payment_token in config) still produce a clearly-named span with the
+    // static nvm.* attrs attached to both the span and the parent tool span.
+    const vspan = await verifySpan({
+      planIds,
+      scheme: resolvedScheme,
+      network: resolvedNetwork,
+      agentId,
+    })
+    // Pre-verify metadata is best-effort and static-only.
+    const preVerifyMd = buildVerifyMetadata({
+      planIds,
+      scheme: resolvedScheme,
+      network: resolvedNetwork,
+      agentId,
+    })
+    vspan.addMetadata(preVerifyMd)
+    addMetadata(parentRunTree, preVerifyMd)
 
     // Extract token from config.configurable.payment_token
     const token = extractPaymentToken(config)
     if (!token) {
+      await vspan.end(new Error('missing payment_token in config.configurable'))
       throw new PaymentRequiredError(
-        "Payment required: missing payment_token in config.configurable",
+        'Payment required: missing payment_token in config.configurable',
         paymentRequired,
       )
     }
@@ -275,18 +320,37 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
         maxAmount: BigInt(creditsToVerify),
       })
     } catch (error) {
+      await vspan.end(error)
       throw new PaymentRequiredError(
         `Payment verification failed: ${error instanceof Error ? error.message : String(error)}`,
         paymentRequired,
       )
     }
 
+    // Augment span metadata with verification results + timing (both span and
+    // parent), then close the verify span. Best-effort: a metadata failure must
+    // not mask the PaymentRequiredError that may follow.
+    const verifyMd = buildVerifyMetadata({
+      planIds,
+      scheme: resolvedScheme,
+      network: resolvedNetwork,
+      agentId,
+      verification,
+      durationMs: Date.now() - verifyStarted,
+      token,
+    })
+    vspan.addMetadata(verifyMd)
+    addMetadata(parentRunTree, verifyMd)
+
     if (!verification.isValid) {
+      await vspan.end(new Error(verification.invalidReason || 'verification failed'))
       throw new PaymentRequiredError(
         `Payment verification failed: ${verification.invalidReason || 'Insufficient credits or invalid token'}`,
         paymentRequired,
       )
     }
+
+    await vspan.end()
 
     // Store payment context
     const paymentContext: PaymentContext = {
@@ -308,7 +372,11 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
         ? credits({ args: args as Record<string, unknown>, result })
         : credits
 
-    // Settle credits
+    // Settle credits, wrapped in an nvm:settlement span (mirrors Python's
+    // settlement_span around settle_permissions).
+    const settleParentRunTree = await activeRunTree()
+    const settleStarted = Date.now()
+    const sspan = await settlementSpan({ planIds, agentId })
     try {
       const settlement = await payments.facilitator.settlePermissions({
         paymentRequired,
@@ -325,8 +393,20 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
       // the receipt — LangGraph copies config.configurable per node, hiding the
       // line above from the buyer's outer scope.
       lastSettlementReceipt = settlement
+
+      const settleMd = buildSettleMetadata({
+        settlement,
+        planIds,
+        agentId,
+        durationMs: Date.now() - settleStarted,
+        token,
+      })
+      sspan.addMetadata(settleMd)
+      addMetadata(settleParentRunTree, settleMd)
+      await sspan.end()
     } catch (settleError) {
       console.error('Payment settlement failed:', settleError)
+      await sspan.end(settleError)
       // Still return result even if settlement fails
     }
 
--- a/src/x402/langchain/decorator.ts
+++ b/src/x402/langchain/decorator.ts
@@ -51,6 +51,7 @@ import {
   type SettlePermissionsResult,
 } from '../facilitator-api.js'
 import {
+  abbreviateToken,
   activeRunTree,
   addMetadata,
   buildSettleMetadata,
@@ -308,6 +309,15 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
       )
     }
 
+    // Abbreviate/redact the token ONCE here (mirrors Python's
+    // attach_metadata_safely pre-abbreviation) and pass the result into the
+    // metadata builders. abbreviateToken is idempotent, so the builders leave
+    // it unchanged — this means the short-token warning fires at most once per
+    // call (not once per verify AND once per settle), and the raw token never
+    // reaches the builders' frame locals (defense against exception enrichers
+    // that capture locals).
+    const abbreviatedToken = abbreviateToken(token)
+
     // Resolve pre-execution credits (static only; callable deferred to post-execution)
     const creditsToVerify = typeof credits === 'number' ? credits : 1
 
@@ -337,7 +347,7 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
       agentId,
       verification,
       durationMs: Date.now() - verifyStarted,
-      token,
+      token: abbreviatedToken,
     })
     vspan.addMetadata(verifyMd)
     addMetadata(parentRunTree, verifyMd)
@@ -399,7 +409,7 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
         planIds,
         agentId,
         durationMs: Date.now() - settleStarted,
-        token,
+        token: abbreviatedToken,
       })
       sspan.addMetadata(settleMd)
       addMetadata(settleParentRunTree, settleMd)
--- a/src/x402/langchain/decorator.ts
+++ b/src/x402/langchain/decorator.ts
@@ -384,7 +384,16 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
 
     // Settle credits, wrapped in an nvm:settlement span (mirrors Python's
     // settlement_span around settle_permissions).
+    //
+    // Re-fetch the active run tree: the verify-side redaction at the top of this
+    // function stripped `payment_token` from the run tree active BEFORE `fn()`,
+    // but `fn()` may have opened nested traced runs, so `activeRunTree()` can now
+    // return a different RunTree that LangChain auto-populated with the raw
+    // `payment_token` from `config.configurable`. Redact it again here, BEFORE
+    // opening the settlement span, so the credential never rides into the settle
+    // child span or the (possibly new) parent tree.
     const settleParentRunTree = await activeRunTree()
+    redactMetadataKeys(settleParentRunTree, 'payment_token')
     const settleStarted = Date.now()
     const sspan = await settlementSpan({ planIds, agentId })
     try {
--- a/src/x402/langchain/decorator.ts
+++ b/src/x402/langchain/decorator.ts
@@ -125,7 +125,17 @@ export class PaymentRequiredError extends Error {
  * Payment context stored in `config.configurable.payment_context` after verification.
  */
 export interface PaymentContext {
-  /** The x402 access token */
+  /**
+   * Abbreviated, non-functional reference to the x402 access token — the SAME
+   * redacted form surfaced as `nvm.payment_token` (`<first 16>…<last 4>`, or a
+   * `…(short)` marker for a too-short token). The **full token is deliberately
+   * not persisted here**: this object is written into
+   * `config.configurable.payment_context`, and tracing frameworks (e.g.
+   * LangChain) can capture `config.configurable` into span metadata, so storing
+   * the raw credential would let it ride into any traced run opened during or
+   * after the tool body. Settlement uses the token read from
+   * `config.configurable.payment_token`, never this field.
+   */
   token: string
   /** The payment required object */
   paymentRequired: X402PaymentRequired
@@ -362,9 +372,16 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
 
     await vspan.end()
 
-    // Store payment context
+    // Store payment context. The `token` field carries the ABBREVIATED
+    // reference, never the raw credential: `payment_context` is written into
+    // `config.configurable`, which LangChain can capture into span metadata, so
+    // persisting the full token here would reopen the very leak the parent-tree
+    // `redactMetadataKeys('payment_token')` calls close — and a child run opened
+    // *during* the tool body would capture it before any post-hoc redaction
+    // could run. `abbreviatedToken` is always defined past the `!token` guard
+    // above; `?? ''` is a belt-and-suspenders that can never fall back to raw.
     const paymentContext: PaymentContext = {
-      token,
+      token: abbreviatedToken ?? '',
       paymentRequired,
       creditsToSettle: creditsToVerify,
       verified: true,
--- a/src/x402/langsmith/spans.ts
+++ b/src/x402/langsmith/spans.ts
@@ -35,9 +35,24 @@ export const NVM_SETTLEMENT_SPAN = 'nvm:settlement'
 /**
  * Marker appended to a redacted short token. The joining character is the
  * single Unicode ellipsis `…` (U+2026), matching the Python implementation
- * and the spec, NOT three ASCII dots.
+ * and the spec, NOT three ASCII dots. (`…` is one UTF-16 code unit, so this
+ * string has length 8.)
  */
 const SHORT_TOKEN_MARKER = '…(short)'
+
+/**
+ * A redacted short token is exactly `<prefix><marker>` where `prefix` is either
+ * empty (raw length 4 or fewer) or the first 4 chars (raw length over 4) — so a genuine
+ * marker only ever has one of these two lengths. Recognising it by both suffix
+ * AND an exact length stops a raw ≤20-char value that merely *ends* in the
+ * marker (e.g. `"x…(short)"`) from slipping through verbatim.
+ */
+const REDACTED_MARKER_LENGTHS = new Set([SHORT_TOKEN_MARKER.length, 4 + SHORT_TOKEN_MARKER.length])
+
+/** True if `token` is already a value produced by the short-token branch. */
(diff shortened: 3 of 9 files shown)


r-marques:
Docs PR (TypeScript LangSmith tracing section): https://github.com/nevermined-io/docs/pull/213

src/x402/langsmith/spans.ts
```diff
@@ -0,0 +1,381 @@
...
+/** True when the dynamic-import error means the package is simply not installed. */
+function isModuleNotFound(err: unknown): boolean {
+  const code = (err as { code?: string } | null)?.code
+  return code === 'ERR_MODULE_NOT_FOUND' || code === 'MODULE_NOT_FOUND'
+}
+
+async function loadLangsmith(): Promise<LangsmithModule | null> {
+  if (cachedLangsmith !== undefined) return cachedLangsmith
+  try {
+    const mod = (await import('langsmith')) as unknown as LangsmithModule
```
aaitor: **HIGH — `import('langsmith')` does not export `getCurrentRunTree`; the feature ships disabled in every install.**

Verified locally against `langsmith@0.7.4`:

```
$ node -e "import('langsmith').then(m => console.log(typeof m.getCurrentRunTree))"
undefined
$ node -e "import('langsmith/singletons/traceable').then(m => console.log(typeof m.getCurrentRunTree))"
function
```

`getCurrentRunTree` lives exclusively at the `langsmith/singletons/traceable` sub-path and has since it was introduced. The `if (typeof mod?.getCurrentRunTree === 'function')` branch never takes the truthy path, `cachedLangsmith` is set to `null`, the "installed but does not export getCurrentRunTree" warning fires once per process, and every span call no-ops for the rest of the process. **The feature delivers zero value in production.**

The unit tests only pass because `jest.mock('langsmith', () => ({ ... getCurrentRunTree: ... }))` injects the API onto the mock's root — they don't reflect the real module shape. This is precisely the real-SDK gap the PR description flags as deferred risk; landing it bites here.

```suggestion
    const mod = (await import('langsmith/singletons/traceable')) as unknown as LangsmithModule
```

**Also please add a node-level smoke test** that imports the real `langsmith/singletons/traceable` (no mocking) and asserts `typeof getCurrentRunTree === 'function'`, so a future package-rearrangement on langsmith's side surfaces here rather than at a customer.

package.json
```diff
@@ -96,7 +101,8 @@
     "@langchain/core": ">=0.3.0",
     "@langchain/langgraph": ">=1.0.0 <2",
     "@modelcontextprotocol/sdk": ">=1.25.0",
-    "express": ">=4.0.0"
+    "express": ">=4.0.0",
+    "langsmith": ">=0.3.0"
```
aaitor: **MEDIUM — `peerDependencies` `langsmith: ">=0.3.0"` is too wide; two behaviors this code relies on changed within that range.**

1. The `RunTree.metadata` setter merging into `extra.metadata` (`{ ...existing, ...new }`) is the behavior `addMetadata` relies on for the double-pass (pre-verify + post-verify) accumulation. In earlier 0.x versions the setter assigns the whole field, so the second `addMetadata` call would clobber the static `nvm.plan_ids`/`nvm.scheme`/`nvm.network`/`nvm.agent_id` from the pre-verify pass. Verified merging in `0.7.4`; not a documented langsmith contract; not a property of `>=0.3.0`.
2. `getCurrentRunTree(permitAbsentRunTree: boolean)` — the boolean overload returning `undefined` instead of throwing on no-active-run is a recent addition (the function existed earlier without the parameter). The `try`/`catch` in `activeRunTree()` masks the throws, but pre-overload langsmith versions burn `console.debug` noise on every untraced call.

Tighten to a version where both are present (the dev dep is `^0.7.4`, so the implementation is already developed against 0.7.x):

```suggestion
    "langsmith": ">=0.7.0"
```

Also note in the SDK README that the LangSmith observability surface requires `langsmith ^0.7` (or whatever floor you settle on).

src/x402/langsmith/spans.ts
```diff
@@ -0,0 +1,381 @@
...
+ * Merge `metadata` into `runTree`, swallowing any error. No-op if `runTree` is
+ * absent or `metadata` is empty.
+ *
+ * The langsmith `RunTree` `metadata` setter merges into `extra.metadata`
+ * (`{ ...existing, ...new }`), so this matches Python's `run_tree.add_metadata`.
+ */
+export function addMetadata(runTree: RunTree | undefined, metadata: SpanMetadata): void {
+  if (!runTree || !metadata || Object.keys(metadata).length === 0) return
+  try {
+    runTree.metadata = metadata
```
aaitor: **MEDIUM — `addMetadata` relies on the langsmith setter doing a merge; that's an implementation detail of `run_trees.js`, not a documented contract.**

The JSDoc claim ("setter merges into `extra.metadata`") is correct for `langsmith@0.7.4` (verified: `set metadata(metadata) { this.extra = { ...this.extra, metadata: { ...this.extra?.metadata, ...metadata } } }`), but it's not in the langsmith JSDoc, types, or changelog. A future minor release that flips the setter to plain assignment would silently lose the static `nvm.*` keys after the post-verify pass — and the existing test suite would stay green because both `FakeRunTree` instances mirror the merge.

Make the function behavior self-sufficient by reading and merging on this side of the boundary, mirroring how `redactMetadataKeys` already reaches into `extra.metadata` directly:

```suggestion
  if (!runTree || !metadata || Object.keys(metadata).length === 0) return
  try {
    const rt = runTree as unknown as { extra?: Record<string, unknown> }
    const existing = (rt.extra?.metadata as Record<string, unknown>) ?? {}
    runTree.metadata = { ...existing, ...metadata }
```

src/x402/langchain/decorator.ts
```diff
@@ -308,7 +382,11 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
         ? credits({ args: args as Record<string, unknown>, result })
         : credits
 
-    // Settle credits
+    // Settle credits, wrapped in an nvm:settlement span (mirrors Python's
+    // settlement_span around settle_permissions).
+    const settleParentRunTree = await activeRunTree()
```
aaitor, line 412: **MEDIUM — `settleParentRunTree` is acquired after `fn()` ran but never has `payment_token` re-redacted. Defense-in-depth gap.**

`redactMetadataKeys(parentRunTree, 'payment_token')` at line 280 strips the credential from the run tree active BEFORE `fn()`. But `fn()` is allowed to spawn nested LangSmith runs (any traceable inside the tool body), so by line 387 `activeRunTree()` may return a different `RunTree` object than the one redacted at line 280. Meanwhile the raw token has been written into `config.configurable.payment_context.token` at the `storeInConfigurable(config, 'payment_context', paymentContext)` call (lines 366–374), and LangChain's auto-capture of `configurable` into newly-created run trees is the very behavior `redactMetadataKeys` exists to defend against — that defense is missing on the settle side.

Mirror the verify-side redaction:

```suggestion
    const settleParentRunTree = await activeRunTree()
    redactMetadataKeys(settleParentRunTree, 'payment_token')
```

(Stronger alternative worth considering: drop `token` from `PaymentContext` entirely. Searching the SDK, the only field consumers need from the settle write-back are `payment_settlement` and the buyer-side `lastSettlement()` slot — neither needs the raw token threaded through `configurable`. Eliminating the nested raw-token write removes the leak surface that `redactMetadataKeys` defends against in the first place.)

tests/unit/x402/langsmith-decorator.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  })
+}
+
+const LONG_TOKEN = 'j'.repeat(40)
+
+beforeEach(() => {
+  setActiveRun(undefined)
+  jest.restoreAllMocks()
+})
+
```
aaitor, line 133: **MEDIUM — Missing tests for two payment-flow branches with an active parent run tree.**

The current suite covers verify→settle success, the no-token unpaid probe, the no-active-run no-op, and the short-token warn-once. Two real branches are not covered:

1. **`verification.isValid === false`** (decorator.ts:355-361): facilitator returns `{ isValid: false, invalidReason }`. The verify span MUST be closed with that reason as the error, and `settlePermissions` MUST NOT be called.
2. **`verifyPermissions()` throws** (decorator.ts:326-338, network/facilitator-down path): verify span closes with the exception, no settlement.

A regression that swapped the `if (!verification.isValid)` guard direction, or that moved `await vspan.end(error)` outside the catch block, would leave green CI today. Add (after the LOW fix below to `FakeRunTree.end()` so `endError` can be asserted):

```typescript
it('ends the verify span with invalidReason and skips settlement on isValid=false', async () => {
  const parent = new FakeRunTree({ name: 'tool' })
  setActiveRun(parent)
  const mockPayments = makeMockPayments()
  mockPayments.facilitator.verifyPermissions.mockResolvedValue({
    isValid: false,
    invalidReason: 'Insufficient credits',
  })
  const fn = protectedFn(mockPayments)

  await expect(fn({ topic: 'x' }, { configurable: { payment_token: LONG_TOKEN } }))
    .rejects.toThrow(/Insufficient credits/)
  expect(parent.children).toHaveLength(1)
  expect(parent.children[0].name).toBe('nvm:verify')
  expect(parent.children[0].ended).toBe(true)
  expect(parent.children[0].endError).toBe('Insufficient credits')
  expect(mockPayments.facilitator.settlePermissions).not.toHaveBeenCalled()
})

it('ends the verify span and skips settlement when verifyPermissions throws', async () => {
  const parent = new FakeRunTree({ name: 'tool' })
  setActiveRun(parent)
  const mockPayments = makeMockPayments()
  mockPayments.facilitator.verifyPermissions.mockRejectedValue(new Error('network timeout'))
  const fn = protectedFn(mockPayments)

  await expect(fn({ topic: 'x' }, { configurable: { payment_token: LONG_TOKEN } }))
    .rejects.toThrow(/network timeout/)
  expect(parent.children).toHaveLength(1)
  expect(parent.children[0].name).toBe('nvm:verify')
  expect(parent.children[0].ended).toBe(true)
  expect(parent.children[0].endError).toBe('network timeout')
  expect(mockPayments.facilitator.settlePermissions).not.toHaveBeenCalled()
})
```

Also missing: a settle-failure-with-active-parent test (asserts `parent.children[1].ended === true` AND `endError` is set when `settlePermissions` throws). Useful regression guard for the catch branch at decorator.ts:417-420.

tests/unit/x402/langsmith-decorator.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  createChild(config: {
+    name: string
+    run_type?: string
+    inputs?: Record<string, unknown>
+  }): FakeRunTree {
+    const child = new FakeRunTree(config)
+    this.children.push(child)
+    return child
+  }
```
aaitor, line 57: **LOW — `FakeRunTree.end()` discards arguments; the decorator tests can never assert the error propagated to a span.**

`openNvmSpan` calls `child.end(undefined, errorString)` (spans.ts:318-321). The fake here drops both arguments, so a regression that swapped to `child.end(errorString, undefined)`, dropped the error argument, or stringified the wrong field would not surface in any decorator test. The spans-test `FakeRunTree` already captures `endError`; copy that signature so the MEDIUM comment's new tests can assert against `endError`.

Minimal change — add the field to the field declarations near line 18:

```typescript
endError: string | undefined
```

and replace lines 52-54:

```typescript
async end(_outputs?: unknown, error?: string): Promise<void> {
  this.ended = true
  this.endError = error
}
```

Better still, extract a shared `tests/unit/x402/helpers/fake-run-tree.ts` so the two test files can't drift apart silently.

src/x402/langchain/decorator.ts
```diff
@@ -187,7 +197,6 @@ function storeInConfigurable(config: unknown, key: string, value: unknown): void
 
   const configurable = (config as Record<string, unknown>).configurable
   if (configurable == null || typeof configurable !== 'object') return
-
   ;(configurable as Record<string, unknown>)[key] = value
```
aaitor, line 210: **LOW — Unrelated whitespace change.**

The blank line between the `configurable` null-guard and the assignment in `storeInConfigurable` was removed. It's outside the feature's blast radius — please put it back so the diff stays focused on the langsmith wiring.

aaitor (changes requested):
Big PR with a strong shape: imperative-span model that mirrors Python decorator-for-decorator, redact-and-warn token defense with the right idempotency reasoning, pre-verify metadata so unpaid discovery probes are still labeled, settle span around the settle call, lazy peer-dep import that no-ops cleanly. The TS port lines up attribute-for-attribute with `payments-py` against the v1 spec (nvm-monorepo#1882), and the unit suite covers most of the wiring carefully.

**One HIGH issue blocks merge** — and it's the one the PR description itself flags as deferred risk: the unit tests use a `jest.mock('langsmith', ...)` that injects `getCurrentRunTree` onto the mock's root, but the **real `langsmith` package does not export `getCurrentRunTree` from its root entry point**. It lives at `langsmith/singletons/traceable`. Result: in every install, `loadLangsmith()` falls to the "installed but missing getCurrentRunTree" branch, caches `null`, and every span call no-ops for the rest of the process. The feature ships completely dead with green CI. Verified locally against `langsmith@0.7.4`:

```
$ node -e "import('langsmith').then(m => console.log(typeof m.getCurrentRunTree))"
undefined
$ node -e "import('langsmith/singletons/traceable').then(m => console.log(typeof m.getCurrentRunTree))"
function
```

Fix is one line (HIGH inline comment), but please also (a) add a real-SDK smoke test that catches this class of regression for next time, and (b) tighten the peer-dep floor since `>=0.3.0` is wider than the implementation actually supports. The other comments are MEDIUM/LOW polish that becomes easier to land while this is open.

**Verification I ran**
- `gh pr diff 371 --repo nevermined-io/payments`, then read each post-PR file at `ref=feat/ts1-nevermined-tracer`.
- Real-SDK probe: `npm install langsmith@0.7.4` in a tmp dir, `node -e "import('langsmith').then(m => …)"` → `getCurrentRunTree` is `undefined` at the root, present at `langsmith/singletons/traceable`. The unit-test mock injects the API onto the root, which is why everything is green.
- Verified the `RunTree.metadata` setter at `langsmith@0.7.4` does merge (`grep 'set metadata' node_modules/langsmith/dist/run_trees.js` → `this.extra = { ...this.extra, metadata: { ...this.extra?.metadata, ...metadata } }`), so the `addMetadata` impl works today against 0.7.x — but the merge isn't a documented langsmith contract, hence the MEDIUM comment.
- Cross-checked spans + decorator wiring against the v1 spec (nvm-monorepo#1882) and the Python `payments_py/langsmith/spans.py` reference.
- Parallel agent panel: `code-reviewer`, `security-reviewer`, `tdd-guide`. Both `code-reviewer` and `security-reviewer` independently surfaced the import-path bug; `tdd-guide` raised the missing `isValid=false` / `verifyPermissions-throws` test coverage with active parent runs. The 12-char-bypass on `isRedactedMarker` (security-reviewer) was downgraded to LOW after I traced it: it can't bite x402 JWTs (no U+2026) and follows the spec text verbatim, so worth fixing in the spec first.
- CI: green (Unit & Integration, E2E, Lint & Build, CLI Sync Check, OpenClaw Plugin Check all pass).
- Base branch (`feat/ts0-langchain-helpers`, PR #370) cross-check: confirmed `lastSettlement()` exists at decorator.ts:172 of the TS-1 head, and the TS-0 helpers are stable enough that TS-1's diff reviews cleanly in isolation.

Once the HIGH is fixed, this is approve-able.

r-marques:
@aaitor — addressed in d32e22f. Thanks for the thorough review (and for actually `npm install`-ing langsmith to catch the dead import — that one would have shipped silently).

**HIGH — import path.** Fixed: `loadLangsmith()` now imports `langsmith/singletons/traceable` instead of the root, where `getCurrentRunTree` actually lives. Verified locally against `langsmith@0.7.4` (`import('langsmith/singletons/traceable')` → `function`; root → `undefined`). Added `tests/unit/x402/langsmith-real-sdk.test.ts`, an **un-mocked** smoke test that imports the real sub-path in a child `node` process (the project's jest config can't load the real langsmith — ESM-only transitive deps — so the child-process dodge runs the same import production does) and asserts `typeof getCurrentRunTree === 'function'`. Both unit mocks were retargeted from `'langsmith'` to `'langsmith/singletons/traceable'` accordingly, otherwise they'd no-op the wiring after the fix.

**MEDIUM — peer-dep floor.** `langsmith >=0.3.0` → `>=0.7.0`, and documented the optional-peer floor (plus the LangChain peers) in the SDK README under Requirements.

**MEDIUM — `addMetadata` merge.** Now reads `extra.metadata` and merges on our side (`{ ...existing, ...new }`) instead of relying on langsmith's undocumented setter-merge, so the pre/post-verify double-pass stays correct even if a future langsmith flips the setter to plain assignment. Mirrors how `redactMetadataKeys` already reaches into `extra.metadata`.

**MEDIUM — settle-side redaction.** Added `redactMetadataKeys(settleParentRunTree, 'payment_token')` right after re-fetching the active run tree on the settle side, before opening the settlement span — closing the gap where `fn()` could have opened a new traced run that LangChain repopulated with the raw token.

**MEDIUM/LOW — tests + fakes + whitespace.** Added the `isValid=false`, `verifyPermissions-throws`, and settle-failure branch tests (the last also asserts the stale-receipt guard: a failed settle leaves `lastSettlement()` undefined). `FakeRunTree.end()` in the decorator test now captures `endError` so those assertions are possible. Restored the stray-removed blank line in `storeInConfigurable`.

Local: full unit suite green (24 suites / 211 passed + 1 skipped), `tsc` clean, `eslint ./src` clean. The 12-char `isRedactedMarker` bypass you downgraded to LOW is left for the spec-first path — see the codepoint-length clarification in nvm-monorepo#1882.

src/x402/langchain/decorator.ts
```diff
@@ -308,7 +382,20 @@ export function requiresPayment<TArgs extends Record<string, unknown>, TResult>(
...
+    //
+    // Re-fetch the active run tree: the verify-side redaction at the top of this
+    // function stripped `payment_token` from the run tree active BEFORE `fn()`,
+    // but `fn()` may have opened nested traced runs, so `activeRunTree()` can now
+    // return a different RunTree that LangChain auto-populated with the raw
+    // `payment_token` from `config.configurable`. Redact it again here, BEFORE
+    // opening the settlement span, so the credential never rides into the settle
+    // child span or the (possibly new) parent tree.
+    const settleParentRunTree = await activeRunTree()
+    redactMetadataKeys(settleParentRunTree, 'payment_token')
```
aaitor, line 413: **MEDIUM — this strips the top-level `payment_token` key, but the raw token is still written to `config.configurable.payment_context.token` (L367/L374) and isn't covered here.**

The verify-side (L280) and settle-side (this line) redactions both delete the `payment_token` *key* from the active run tree's metadata. But after a successful verify, the full token is stored back into the configurable bag:

```ts
const paymentContext: PaymentContext = { token, /* ...raw token... */ }
storeInConfigurable(config, 'payment_context', paymentContext)   // L374
```

LangChain auto-capturing `config.configurable` into span metadata is the *exact* mechanism these `redactMetadataKeys` calls exist to defend against — and by parity it can capture `payment_context` (an object whose `.token` is the raw credential) into any traced run created during/after `fn()`. Deleting the `payment_token` key does not touch `payment_context.token`, so the credential can still ride into a nested span in a multi-hop / chained-agent flow.

This is the "stronger alternative" I flagged last round — defense-in-depth, not a confirmed leak on the primary vector, so it's a non-blocking follow-up. Cleanest fixes (pick one):

1. **Don't persist the raw token in `configurable`** — store only the abbreviated form, since nothing downstream needs the full token after verify:
   ```ts
   const paymentContext: PaymentContext = { ...rest, token: abbreviatedToken }
   ```
2. Or **also redact the nested key** at both redaction sites:
   ```suggestion
    redactMetadataKeys(settleParentRunTree, 'payment_token', 'payment_context')
   ```
   (and the same at L280) — assuming no consumer reads `payment_context` back off the run-tree metadata.
3. Or, if you've confirmed LangChain does **not** deep-capture nested `configurable` objects into metadata, drop a one-line comment here saying so, so the asymmetry between "redact `payment_token`" and "leave `payment_context.token`" is intentional and documented.

A regression test that seeds the parent run tree with both `payment_token` and `payment_context.token` and asserts neither survives into the child span would lock whichever option you choose.

aaitor (approved):
Re-review after `d32e22f6`. Every issue from my prior CHANGES_REQUESTED is addressed and verified at HEAD — including the HIGH that would have shipped the whole feature dead. Approving. One residual MEDIUM (a deeper version of a point I raised parenthetically last time) plus two LOWs are noted below as non-blocking follow-ups.

**What I verified is fixed (read each file at HEAD, not just the commit messages)**
- **HIGH — langsmith import path** (`spans.ts:85`): now `import('langsmith/singletons/traceable')`; the only remaining root `langsmith` reference is a type-only `import type { RunTree }` (erased at compile, no runtime load). And the new **un-mocked smoke test** (`langsmith-real-sdk.test.ts`) spawns a real child `node` process to assert `typeof getCurrentRunTree === 'function'` at that sub-path — a genuine guard against this exact class of regression, and a nice solution to the jest-doesn't-transform-node_modules problem. ✅
- **MEDIUM — peer-dep floor** (`package.json:105`): `>=0.7.0`. ✅
- **MEDIUM — `addMetadata` merge** (`spans.ts:141-151`): now reads `extra.metadata` and merges on this side instead of relying on the langsmith setter. ✅
- **MEDIUM — settle-side redaction** (`decorator.ts:395-396`): `redactMetadataKeys(settleParentRunTree, 'payment_token')` now fires before the settlement span opens, mirroring the verify side. ✅
- **MEDIUM — branch coverage** (`langsmith-decorator.test.ts:224/245/262`): added tests for `isValid=false` (verify ends w/ reason, settle skipped), `verifyPermissions` throws, and `settlePermissions` throws (settle span ends WITH error, decorator still returns). ✅
- **LOW — `FakeRunTree.end()`** (test:59-61): now captures the error string so the above can assert `endError`. ✅

**Verification I ran**
- Read every changed source/test file at `ref=feat/ts1-nevermined-tracer`; cross-checked the `./langsmith` exports against docs#213 and the span attributes against the v1 spec (nvm-monorepo#1882).
- Adversarial recheck pass (code-reviewer agent) over the redaction/abbreviation logic, the metadata builders, and all throw paths: confirmed no span is left un-ended on any throw path; `abbreviateToken` idempotency + warn-once-per-call are correct; `nvm.tx_hash` gates on truthiness and `nvm.plan_ids` always emits an array per spec; no raw token reaches a span attribute on the primary (top-level `payment_token`) vector.
- CI: Unit & Integration, E2E, Lint & Build, CLI Sync, OpenClaw Plugin — all green.

**Residual (non-blocking) follow-ups**
- **MEDIUM** — the raw token is still written to `config.configurable.payment_context.token`; the redaction strips the top-level `payment_token` key but not the nested one (inline comment with concrete options). The *primary, confirmed* capture vector is now defended on both sides, so this is defense-in-depth, not a confirmed leak — but worth closing since it's a credential. A small change; fine as a fast-follow.
- **LOW** — `isRedactedMarker` uses `.length` rather than the spec's `[...str].length`; the spec (§4) explicitly carves this out as v1-conformant for ASCII tokens, so it's optional polish.
- **LOW** — no unit test for the "`langsmith` not installed" path (`cachedLangsmith=null`, no-retry); a `resetModules` + throwing-import test would guard the no-retry caching.

Strong PR; the fixes are exactly on target. Approving with the above as follow-ups rather than gates.

r-marques:
Thanks for the approval and the precise follow-ups, @aaitor.

**Residual MEDIUM (nested `payment_context.token`) — addressed in `6bc08a6`.** Took **option 1**: `payment_context.token` now stores the *abbreviated* reference (`abbreviatedToken`), not the raw token. I picked it over the redact-the-nested-key option because it's the only one that also covers a run opened **during** `fn()` — a post-hoc `redactMetadataKeys('payment_context')` can't reach a child span the tool body already created. Confirmed nothing reads the raw value back (settlement uses the token from `config.configurable.payment_token`, not this field; no SDK/tutorial consumer of `payment_context.token`), so it's a no-functional-change fix. Updated the `PaymentContext.token` JSDoc and added a regression test asserting the persisted `payment_context.token` is the abbreviated form and the stored bag carries no raw token.

**payments-py parity:** Python stores the raw token in its `PaymentContext` the same way — filed as a fast-follow: nevermined-io/nvm-monorepo#1891.

**The two LOWs** I'm leaving as-is for now:
- `isRedactedMarker` `.length` vs `[...str].length` — **intentionally deferred**: the v1 spec (nvm-monorepo#1882, now merged-pending) explicitly pins the length unit to codepoints but states ASCII+U+2026 markers make `.length` v1-conformant, so no SDK change is required. Changing it here would contradict that carve-out.
- "`langsmith` not installed" no-retry test — agreed it's a nice coverage add; tracking it as optional polish rather than gating this PR.

Local: full unit suite green (24 suites / 212 passed + 1 skipped), `tsc` clean, `eslint ./src` clean. Pushing now triggers CI.

PR #384, opened 2026-06-16
feat(mcp): align x402 transport with v2 in-band MCP spec

r-marques:
## Summary

Ports [`nevermined-io/payments-py#203`](https://github.com/nevermined-io/payments-py/issues/203) (design) / [`#228`](https://github.com/nevermined-io/payments-py/pull/228) (Python implementation) to this TS SDK: aligns the MCP transport with the [x402 v2 MCP transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/mcp.md), replacing HTTP-status / header payment semantics with **in-band JSON-RPC signaling**.

## What changed
- **Payment-required** → MCP tool result with `isError: true`, populating **both** `structuredContent` (the `PaymentRequired` object, `accepts` from the agent's plans) and `content[0].text` (JSON-stringified). No HTTP 402 on `/mcp`.
- **Request payment** read from `_meta["x402/payment"]`. `Authorization: Bearer` kept as a **deprecated fallback** for one release.
- **Settlement** emitted via `_meta["x402/payment-response"]`; Nevermined observability moved under `_meta["nevermined/credits"]`.
- **Post-execution settlement failure** suppresses the tool's content and returns the payment error.
- **Resources/prompts** keep raising JSON-RPC errors (the in-band tool-result shape is tool-only).
- Adds `encodeAccessToken` (inverse of `decodeAccessToken`), `buildPaymentRequiredForPlans`, and typed `PaymentRequiredError`/`SettlementFailedError`.

## Decisions (mirrors the Python PR)
- `onRedeemError: "ignore"` no longer delivers paid content on a post-execution settlement failure — spec-correct ("never deliver paid content without settlement landing"). It now only distinguishes the surfaced error type.
- The `data: { reason }` discriminator on the auth error is dropped (the `PaymentRequired` object replaces it) — same as the Python PR.

## TS-specific notes
- **No discovery removal**: unlike payments-py, this SDK never had `/.well-known/x402-payment`, an `enableX402Discovery` toggle, or `payment-*` CORS headers — nothing to remove there. OAuth discovery is untouched.
- **Drive-by fix**: corrects a pre-existing swapped-args bug in the non-streaming settlement-fallback call (`fallbackEndpoint`/`httpVerb` were transposed vs the correct streaming site). This PR makes the bug user-visible (settlement failure now suppresses content), so it's fixed here.

## Testing
- `pnpm test:unit`: **255 passed** (1 skipped), `pnpm test:integration`: **36 passed**. Build clean. (The one failing unit suite, `langchain-decorator.test.ts`, is a pre-existing missing optional dep — fails on `main` too, unrelated.)
- New `tests/unit/mcp/x402_inband.test.ts` covers the in-band tool-result shape, codec round-trip + foreign-encoded recovery, in-band precedence over the header, the success receipt + namespacing, free/no-credit omission, post-execution suppression, the over-catch guard (a generic tool error must propagate), internal-throw-returns, and resource/prompt throw.

## Reviewed
Multi-agent panel (correctness / silent-failure / test-coverage) run against the diff; no correctness blockers, all should-fixes addressed.

## Caveats / follow-ups
- **Live-facilitator round-trip not yet validated** (codec round-trip proven; the EIP-712 signature is inside `payload`, not the base64 envelope). Recommend an e2e check before merge.
- **Streaming tools** can't retroactively suppress already-yielded chunks on a late settlement failure (the final `_meta` chunk carries the failure).
- **`docs_mintlify`** (separate repo) documents the TS x402/MCP wire shape and needs the same rewrite — out of scope here.


src/utils.ts
```diff
@@ -77,6 +77,33 @@ export const decodeAccessToken = (accessToken: string): Record<string, any> | nu
...
+ * NOTE: `btoa` throws on code points > U+00FF, so this assumes ASCII JSON. x402
+ * PaymentPayloads are ASCII (EIP-712 hex/decimal/signature fields), matching the
+ * Python SDK's equivalent. A future non-ASCII field would need UTF-8 encoding first.
+ *
+ * @param payload - The decoded PaymentPayload object (e.g. from `_meta["x402/payment"]`)
+ * @returns The base64url-encoded access token string (unpadded)
+ */
+export const encodeAccessToken = (payload: Record<string, any>): string => {
+  const json = JSON.stringify(payload)
+  return btoa(json).replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '')
```
aaitor: **MEDIUM - btoa throws on non-ASCII, and this runs on untrusted client input.**

encodeAccessToken is called on the client-supplied _meta x402/payment. btoa throws a DOMException on any code point above U+00FF. That throw happens inside the paywall try; it is not a PaymentRequiredError, so for a tool it re-throws as a generic JSON-RPC error instead of a clean payment-required - and it diverges from the Python SDK, which base64s UTF-8 bytes and never throws.

Encode UTF-8 bytes instead of calling btoa(json) directly, so it cannot throw on non-ASCII. Either use a Node-safe Buffer.from(json, utf8).toString(base64url), or (platform-agnostic) map new TextEncoder().encode(json) to a binary string before btoa. Add a test with a non-ASCII field to pin it. (x402 payloads are ASCII today, so a full UTF-8 round-trip would also need decodeAccessToken to TextDecoder the bytes - the priority here is not crashing on hostile input.)

src/mcp/utils/meta.ts
```diff
@@ -0,0 +1,59 @@
...
+ * The MCP TS SDK exposes the incoming request `_meta` on the tool handler's
+ * `extra` argument (`extra._meta`), and its schema is a passthrough object, so
+ * non-standard keys like `"x402/payment"` survive parsing.
+ *
+ * @param extra - The MCP handler `extra` argument.
+ * @returns The decoded PaymentPayload object, or `undefined` when absent.
+ */
+export function readPaymentPayload(extra: any): Record<string, any> | undefined {
+  const value = extra?._meta?.[X402_PAYMENT_META_KEY]
+  return value && typeof value === 'object' ? value : undefined
```
aaitor: **LOW - typeof value === object accepts arrays; the Python sibling rejects them.**

Because typeof of an array is object (and a non-empty array is truthy), an array payload passes through readPaymentPayload into encodeAccessToken. The Python read_payment_payload uses isinstance(value, dict), which rejects lists. Add a not-Array.isArray(value) guard (and an explicit non-null check) to the return so only plain objects pass.

src/mcp/core/paywall.ts
```diff
@@ -90,98 +107,172 @@ export class PaywallDecorator {
...
           effectivePlanId,
+          authResult.token,
           authResult.subscriberAddress,
           credits,
+          options,
+          authResult.agentId,
+          authResult.logicalUrl,
+          // fix: pre-existing arg order — fallbackEndpoint=httpUrl, httpVerb='POST'
+          // (matches the streaming site above)
+          authResult.httpUrl,
```
aaitor, line 211: **MEDIUM - the drive-by fallback-arg fix ships without a regression test.**

Good catch fixing the transposed fallbackEndpoint/httpVerb here (was POST, httpUrl; now httpUrl, POST). But nothing exercises the settle-fallback path, so the old arg order would still pass CI. This is a money path - pin it with a red-to-green test: mock settlePermissions to throw on the first call and succeed on the second, set a distinct authResult.httpUrl, then assert the second call paymentRequired.resource.url equals the httpUrl (under the old order it would be POST).

tests/unit/mcp/x402_inband.test.ts
```diff
@@ -0,0 +1,318 @@
...
+    const result = paymentRequiredResult(SAMPLE_PR)
+    expect(result.isError).toBe(true)
+    expect(result.structuredContent).toEqual(SAMPLE_PR)
+    expect(result.structuredContent.x402Version).toBe(2)
+    expect(result.content[0].type).toBe('text')
+    expect(JSON.parse(result.content[0].text)).toEqual(SAMPLE_PR)
+  })
+})
+
+describe('x402 in-band: paywall wrapper', () => {
```
aaitor, line 151: **MEDIUM - no test exercises the real MCP-SDK dispatch; the extra._meta assumption is unverified in-repo.**

Every test here hand-builds extra._meta. I confirmed against the SDK that @modelcontextprotocol/sdk v1.25.3 does surface request _meta on the handler extra (Protocol._onrequest sets _meta: request.params?._meta; RequestMetaSchema is z.core.loose passthrough), so this works today - but a future SDK bump that changed it would silently route every call to the deprecated header fallback with zero failing tests. The Python sibling added TestDispatcherInBand for exactly this. Port it: register a tool on a real McpServer, dispatch a CallToolRequest whose params._meta x402/payment is set, and assert the handler extra carried the payload (re-encoded to a Bearer token).

aaitor (commented):
Clean TS port of payments-py#228; CI green (7/7). I verified the one load-bearing assumption myself: @modelcontextprotocol/sdk v1.25.3 does surface request _meta on the tool handler extra (Protocol._onrequest sets _meta: request.params?._meta, and RequestMetaSchema is a z.core.loose passthrough so x402/payment survives parsing) - so the in-band read works at runtime. No HIGH survived verification. The findings below are robustness plus test-debt MEDIUM/LOWs. The error-field handling here is the correct one - it is the Python sibling that needs aligning. Leaving as COMMENT (non-blocking - no HIGH).

**Verification I ran**
- Read the full diff plus changed files post-PR on feat/mcp-x402-v2-inband. Confirmed the swapped-args fix at L209-211 matches the streaming call site and the redeemCredits signature.
- Inspected node_modules/@modelcontextprotocol/sdk v1.25.3: _meta passthrough on extra confirmed (so the in-band path is not dead).
- 5-agent panel (code-reviewer, security-reviewer, python-reviewer for the sibling, tdd-guide, code-explorer); cross-checked against the code. Demoted the no-extra._meta-wiring HIGH to a test-debt MEDIUM after confirming the SDK behavior.
- CI: 7/7 OK.

**Notes (not gating)**
- LOW: _meta x402/payment-response carries the full SettlePermissionsResult (incl. creditsRedeemed/remainingBalance), beyond the spec receipt success/transaction/network/payer and slightly against this PR stated keep-the-spec-key-spec-pure intent. Python does the same in prod, so it is consistent - just not spec-minimal. Tolerable; consider filtering before placing under the spec key.
- LOW: onRedeemError propagate now throws -32002 Misconfiguration (was -32003) - untested in both SDKs; add a test asserting the throw plus code.
- LOW: markdown/mcp-integration.md is ~80% cosmetic reflow (quote style, table padding) interleaved with the real doc updates, which buries the substance. Consider splitting the reflow out (or confirm it is an intentional prettier pass).
- Public-package breaking changes (tool payment-required no longer throws -32003, now an isError result; _meta top-level keys moved under nevermined/credits; onRedeemError ignore no longer returns content). Documented in the PR body, but the commit is feat(mcp) with no BREAKING CHANGE footer - confirm the npm version bump reflects it.

r-marques:
Thanks for the thorough review, @aaitor — all addressed in 0202375.

- **`btoa` non-ASCII (MEDIUM)** — `encodeAccessToken` now UTF-8-encodes before base64 (`TextEncoder` → binary string → `btoa`), so it never throws on non-ASCII client input. Byte-identical to the old output for ASCII, so the round-trip with `decodeAccessToken` is unchanged. Added a non-ASCII test.
- **`readPaymentPayload` accepts arrays (LOW)** — now rejects arrays and null (`value !== null && typeof === 'object' && !Array.isArray`), matching the Python `isinstance(dict)`. Added a test.
- **Fallback-arg fix untested (MEDIUM)** — added a red→green regression test: `settlePermissions` throws then succeeds, distinct `authResult.httpUrl`, asserts the fallback builds `resource.url === httpUrl` (would be the literal `"POST"` under the old transposed order).
- **No real MCP-SDK dispatch test (MEDIUM)** — ported the Python `TestDispatcherInBand`: registers the protected tool on a real `McpServer`, drives a `CallToolRequest` over `InMemoryTransport` with `params._meta["x402/payment"]`, and asserts the SDK delivered it to the handler `extra` (re-encoded to a Bearer token). Pins the load-bearing `extra._meta` assumption.
- **`onRedeemError: "propagate"` untested (LOW)** — added a test asserting it throws `-32002`.

Not changed (by design):
- `_meta["x402/payment-response"]` carrying the full `SettlePermissionsResult` — kept for consistency with the Python SDK (you noted it's tolerable).
- The `markdown/mcp-integration.md` reflow was an intentional `prettier --write` pass.

On the **breaking-change / version** note: agreed — the branch currently carries a *minor* bump (`v1.7.0`) but these changes are breaking. Leaving the `BREAKING CHANGE:` footer / major-bump call to the maintainers.

tests/unit/mcp/x402_inband.test.ts
```diff
@@ -0,0 +1,452 @@
...
+    // standard base64). The semantic payload must still round-trip.
+    const foreign = btoa(JSON.stringify(SAMPLE_PAYLOAD, null, 2)) // padded, spaced
+    expect(decodeAccessToken(foreign)).toEqual(SAMPLE_PAYLOAD)
+    // And re-encoding the decoded payload still recovers it.
+    expect(decodeAccessToken(encodeAccessToken(decodeAccessToken(foreign)!))).toEqual(
+      SAMPLE_PAYLOAD,
+    )
+  })
+
+  test('encode never throws on non-ASCII input (hostile client payload)', () => {
```
aaitor: **LOW — strengthen the non-ASCII codec test.**

This asserts `encodeAccessToken` doesn't throw on non-ASCII, but not that the payload is recoverable. Since `decodeAccessToken` uses `atob` (byte-wise, not UTF-8), a non-ASCII round-trip is *not* lossless today — worth pinning explicitly so the asymmetry is intentional and visible:
```ts
const encoded = encodeAccessToken({ ...SAMPLE_PAYLOAD, note: 'café 😀' })
const decoded = decodeAccessToken(encoded)
expect(decoded?.x402Version).toBe(2)            // ASCII fields survive
expect(decoded?.accepted?.planId).toBe('plan-123')
// note: decoded.note is NOT expected to equal 'café 😀' — atob is byte-wise.
```

tests/unit/mcp/x402_inband.test.ts
```diff
@@ -0,0 +1,452 @@
...
+
+  test('empty plan list falls back to a single empty plan id', () => {
+    const pr = buildPaymentRequiredForPlans([], { endpoint: 'mcp://x' })
+    expect(pr.accepts).toHaveLength(1)
+    expect(pr.accepts[0].planId).toBe('')
+  })
+})
+
+describe('buildPaymentRequiredError: plans-lookup failure', () => {
+  test('a backend outage surfaces "plans unavailable", not a clean 402', async () => {
```
aaitor, line 367: **LOW — assert the fallback `accepts` shape here too.**

This 'plans unavailable' test checks the `error` field; add an assertion that `accepts` is exactly `[{ planId: '' }]` (one entry, empty id) so an empty-`accepts` regression in `buildPaymentRequiredError` is caught at the integration point, not just in the isolated `buildPaymentRequiredForPlans` unit test.

aaitor (approved):
Clean TS port and the reference implementation for the extra-handling: I verified the dispatcher keeps a separate `authExtra` for auth and forwards the **raw** `extra` (with `_meta`) to the user handler (`paywall.ts` L118-125, L166) — the Python sibling (#228) does **not**, which I've flagged there. The swapped-args settlement-fallback fix is correct and has a red->green regression test. No HIGH or MEDIUM survived verification; CI is 7/7. Approving.

**Verification I ran**
- Read the full diff + changed files post-PR on `feat/mcp-x402-v2-inband`. Confirmed `decodeAccessToken` (`utils.ts` L60-66) swaps `-`/`_` before `atob` on the URL-safe branch (an agent flagged this as a bug — it is not), and `encodeAccessToken` UTF-8-encodes (`TextEncoder` -> binary -> `btoa`) so it can't throw on hostile non-ASCII input.
- Confirmed the in-band `_meta["x402/payment"]` read, its precedence over the Authorization header, the post-execution suppression (`SettlementFailedError`), and the tool-only conversion (`kind === 'tool'`) all match the spec and the Python sibling.
- 4-agent panel (security, parity, idioms, test coverage); cross-checked findings against the code.
- CI: 7/7 green.

**Non-gating notes**
- Two LOW test-strengthening nits inline.
- As @aaitor already noted, `_meta["x402/payment-response"]` carries the full `SettlePermissionsResult` (a superset of the spec receipt `success`/`transaction`/`network`/`payer`); consistent with Python, tolerable — consider filtering to the spec fields before placing under the spec key.
- A streaming `onRedeemError: 'propagate'` (thrown-error) path isn't covered; minor test debt.
- Breaking changes (tool payment-required no longer throws `-32003`, now an `isError` result; top-level `_meta` keys moved under `nevermined/credits`; `onRedeemError: 'ignore'` no longer returns content) are documented in the PR body, but the commit is `feat(mcp)` with no `BREAKING CHANGE:` footer and a minor bump — confirm the npm version reflects the break before release.

r-marques:
Thanks @aaitor — folded in the two LOW test nits + the size-bound guard in 6182248 (the approval's non-gating items):

- Non-ASCII codec test now asserts the ASCII fields survive the round-trip, with a comment that non-ASCII is intentionally lossy (`atob` is byte-wise).
- plans-unavailable test now asserts `accepts` is exactly `[{ planId: '' }]` (one entry, empty id).
- `readPaymentPayload` rejects in-band payloads > 64KB (defense-in-depth on untrusted input; parity with the Python sibling).

260 unit + 36 integration green, build + prettier clean. Heads-up that this lands a commit on top of your approval.

PR #391, opened 2026-06-17
feat(mcp): make agentId optional, require planId

r-marques:
## Summary
Make `agentId` **optional** in the MCP server and require `planId` instead — plan-centric, matching the direction that only the pricing plan is necessary.

## Why it's safe (empirically verified)
On staging, `verifyPermissions`/`settlePermissions` succeed with a `PaymentRequired` built with **no agentId** and a no-agentId token (`isValid:true`, settle success, credits charged). The facilitator resolves everything from `planId` + the token/delegation; `agentId` is informational only (response showed `agentId:""`, `agentName:"Unknown Agent"`).

## Changes
- Server config (`McpServerConfig`, paywall config, `configure`/`start`/`withPaywall`/`attach`/`createApp`): **`planId` required, `agentId` optional**.
- Guard now requires an effective `planId` (per-tool option ?? server config); the four `agentId is required` throws → `planId is required`.
- First-call payment-required `accepts` built from the configured `planId`; `getAgentPlans(agentId)` only when an agentId is present.
- `agentId` threaded as optional into `buildPaymentRequired`/auth (already optional in the builder). Verify/settle work plan-only.

## ⚠️ Maintainer decision — OAuth `client_id`
`agentId` also served as the OAuth `client_id`. With it optional, this PR omits `client_id` from the protected-resource metadata when absent and generates an id for dynamic registration. The in-band/bearer flow is unaffected, but confirm no full-OAuth-dance client relies on a stable `client_id == agentId`.

## Tests
`pnpm test:unit` 261 passed (+1 "works with planId, no agentId"), `pnpm test:integration` 36/36; build clean (pre-existing langchain dep aside).

## Release
Should ship in the same new version as the in-band transport (#384), which is merged to `main` but not yet released.


src/mcp/http/client-registration.ts
```diff
@@ -156,9 +156,11 @@ export async function processClientRegistration(
   // Validate the request
   validateClientRegistrationRequest(request)
 
-  // Use agentId as client_id (consistent for this MCP server)
-  const clientId = config.agentId
+  // client_id = the configured agentId when present (stable for this server);
+  // agentId is optional under the plan-centric model, so when it's absent we
+  // issue a generated client_id for OAuth dynamic registration.
   const issuedAt = Math.floor(Date.now() / 1000)
+  const clientId = config.agentId ?? `nvm-mcp-client-${issuedAt}-${Math.floor(Math.random() * 1e9)}`
```
aaitor: **LOW — Use the already-imported CSPRNG instead of `Math.random()` for the generated `client_id`.**

`randomBytes` is already imported at the top of this file (line 5) and used for the client secret at line 134, so this needs no new dependency. `Math.random()` is non-cryptographic (~30 bits) and can collide across concurrent registrations within the same second — the timestamp prefix doesn't prevent that. The Python sibling uses `secrets.token_hex`. Match it:

```suggestion
  const clientId = config.agentId ?? `nvm-mcp-client-${issuedAt}-${randomBytes(8).toString('hex')}`
```

src/mcp/index.ts
```diff
@@ -431,10 +455,16 @@ export function buildMcpIntegration(paymentsService: Payments) {
...
   function createApp(options: Omit<CreateRouterOptions, 'agentId'> & { agentId?: string }) {
+    // agentId is optional under the plan-centric model (informational/OAuth
+    // client_id only); the OAuth router omits client_id when it's absent.
     const agentId = options.agentId || extendedConfig.agentId
-    if (!agentId) {
-      throw new Error('agentId is required. Either pass it in options or call configure() first.')
-    }
+    // planId is the paywall's concern (the OAuth app below does not consume it).
+    // Configure the paywall so tools protected via withPaywall settle on this plan.
+    configure({
```
aaitor: **LOW (leanness) — the `configure({...})` prologue is duplicated verbatim across `createRouter`, `createApp`, and `startServer`.**

All three entry points run the same `const agentId = options.agentId || extendedConfig.agentId` + `configure({ planId: options.planId ?? extendedConfig.planId, agentId, serverName: options.serverName || extendedConfig.serverName || 'mcp-server' })` block, and the `serverName` fallback is then repeated a second time in the inner `create*`/`start*` call. The three copies have already drifted in their comments. Extracting one helper drops ~30 lines:

```ts
function applyServerConfig(options: { planId?: string; agentId?: string; serverName?: string }) {
  const agentId = options.agentId || extendedConfig.agentId
  const serverName = options.serverName || extendedConfig.serverName || 'mcp-server'
  configure({ planId: options.planId ?? extendedConfig.planId, agentId, serverName })
  return { agentId, serverName }
}
```

Each entry point then becomes `const { agentId, serverName } = applyServerConfig(options)`. Non-blocking.

aaitor (approved):
Clean, plan-centric change with the right test additions (the three `createRouter({ planId })` cases), the OAuth `client_id`-omission edge handled, and the design verified both against staging (per the PR body) and against the Python sibling. Approving — two release-hygiene items to fix before merge (neither is a correctness blocker) plus two LOW nits inline.

**Verification I ran**
- `gh pr diff 391` + read post-PR `auth.ts`, `paywall.ts`, `index.ts`, `server-manager.ts`, `client-registration.ts`, `oauth-metadata.ts` at `feat/mcp-agentid-optional`.
- CI: Lint & Build, Unit & Integration, CLI Sync, OpenClaw all green; E2E still pending.
- Confirmed the misconfiguration guard still fires: `paywall.ts` throws `Misconfiguration` when neither per-tool nor server-level `planId` resolves, and `start()` throws `planId is required` — no silent empty-plan settle path. Verified `configure()` in the three entry points doesn't clobber a prior planId (`options.planId ?? extendedConfig.planId`).
- Verified `verifyWithEndpoint`/`buildPaymentRequired` treat `agentId` as informational `extra`; verify/settle resolve from planId + token. Security review found no authorization regression from making it optional.
- Cross-checked the Python sibling payments-py#235 (same change) — note its CI is **red** on an unrelated Python-3.10 f-string issue; this TS PR is unaffected.
- Panel: code-reviewer, security-reviewer, ponytail/leanness (~36 lines removable via a shared helper — inline).

**MEDIUM — `README.md` is now stale (not in this diff).** The "Start the Server" example at `README.md:390-392` calls `payments.mcp.start({ agentId: …, serverName: … })` with **no `planId`** — copy-pasting it now throws `planId is required`. `markdown/mcp-integration.md` was updated but `README.md` was missed. Add `planId: process.env.NVM_PLAN_ID!` there and demote `agentId` to the optional comment, as you did elsewhere.

**MEDIUM — label the breaking change.** `agentId` required→optional and `planId` optional→required: existing callers configured with only `agentId` now throw `planId is required`. Add a `BREAKING CHANGE:` footer and ensure the shared release version with #384 is bumped/changelogged as breaking.

**LOW — lost eager validation in `withPaywall`/`attach`.** These previously threw immediately on missing `agentId`; now there's no registration-time guard, so a missing `planId` only surfaces on the first tool call. An eager `if (!opts.planId && !extendedConfig.planId) throw` at `withPaywall`/`attach` entry would restore fail-fast.

r-marques:
Folded in the review items and fixed the actual CI failure — **CI is now fully green (E2E passes)**.

**E2E root cause (was red):** the e2e fixture (`tests/e2e/test_mcp_oauth_e2e.test.ts`) started the MCP server without `planId` — the newly-required field — so `server-manager.ts` threw `planId is required`. The unit/integration suites were updated for the new contract but the e2e fixture wasn't. Fixed by passing `planId: creditsPlanId` to both `mcp.configure()` and `mcp.start()`.

**Your items:**
- **MEDIUM — stale README example:** fixed. Added `planId: process.env.NVM_PLAN_ID!` to the "Start the Server" example and demoted `agentId` to an optional comment, matching `markdown/mcp-integration.md`.
- **LOW — lost eager validation:** restored. `withPaywall` now throws `Misconfiguration` at registration time when neither per-tool nor server-level `planId` resolves (covers `attach`, whose registrars delegate to `withPaywall`).
- **LOW — duplicated `configure(...)`:** extracted to a shared `resolveTransportConfig()` used by `createRouter`/`createApp`/`startServer` (−44 lines, behaviour identical including the `options.planId ?? config.planId` fallback).
- **MEDIUM — breaking-change label:** holding on the maintainer's call for the `BREAKING CHANGE:` footer / version bump (shared release with #384).

Commit `16c7f24`.

src/mcp/http/client-registration.ts
aaitor: **LOW (carryover, still open) — use the already-imported CSPRNG for the generated `client_id`.**

`randomBytes` is imported at the top of this file (line 5) and already used by `generateClientSecret()` at line 134, so this is a one-token change with no new dependency. `Math.random()` is non-cryptographic (~30 bits) and can collide across concurrent registrations in the same second; the Python sibling uses `secrets.token_hex`.

```suggestion
  const clientId = config.agentId ?? `nvm-mcp-client-${issuedAt}-${randomBytes(8).toString('hex')}`
```

aaitor (approved):
Re-review at `197b3b7d`. All three items I flagged last round are fixed — re-affirming **APPROVE**.

**Fixed since my last review (`47621ff`)**
- ✅ **MEDIUM** — `README.md:390-396`: the "Start the Server" example now includes `planId: process.env.NVM_PLAN_ID!` and demotes `agentId` to the optional comment. Grepped the README — no other `mcp.start`/`createApp`/`startServer` example still omits `planId`.
- ✅ **LOW** — `index.ts` `withPaywall` (306-311): eager `if (!opts.planId && !extendedConfig.planId) throw` restored, so misconfiguration fails at registration again; `attach` inherits it by delegation.
- ✅ **LOW (leanness)** — the 3× duplicated `configure({…})` prologue is extracted into `resolveTransportConfig()` (399-414); I verified it preserves the exact prior behavior (same `agentId`/`serverName` fallbacks, same `configure({ planId: options.planId ?? extendedConfig.planId, … })`, same `environment` resolution) — no drift, and the spread in `configure()` keeps `environment`/`tools`/`scopes` intact. ~30 lines removed.

**Verification I ran**
- Read post-fix `index.ts` (`withPaywall`, `resolveTransportConfig`, the three entry points), `client-registration.ts`, `README.md`; diffed `47621ff…197b3b7d` (filtering the `main`-merge noise: `a2a/*`, `cli/*`, regenerated `docs/**/*.html`).
- CI: Lint & Build ✅, Unit & Integration ✅, CLI/OpenClaw sync ✅, E2E pending.
- Focused code-reviewer pass on the delta: confirmed-fixed 1-3, no regressions from the extraction.

**Still open / notes**
- `Math.random()` for the generated `client_id` (inline LOW) — unchanged.
- The restored eager guard means tools registered via `withPaywall`/`attach` must come *after* a `planId` is set (per-tool option or `configure()`); the simplified README registers tools before `start({planId})`, which is fine because `start` configures first, but a one-line JSDoc note on `withPaywall` about the ordering would prevent a register-first-configure-later surprise.
- Breaking-change labeling (required-field flip) — same `BREAKING CHANGE:` footer note as the py sibling; ensure the shared release with #384 is versioned accordingly.

r-marques:
Folded in the two carryover items from your re-review — `d333e4f`:
- **`client-registration.ts`:** generated `client_id` now uses `randomBytes(8).toString('hex')` (the already-imported CSPRNG) instead of `Math.random()`, matching the Python sibling's `secrets.token_hex`.
- Added the JSDoc note on `withPaywall` about registering paywalled handlers *after* a `planId` is configured (the restored eager guard throws at registration otherwise).

`tsc` build clean, full jest suite green. The `Math.random()` inline LOW is resolved. Breaking-change footer still held for the shared release with #384.

PR #392, opened 2026-06-17
feat(a2a): align A2A with x402 v2 in-band transport + canonical agent-card discovery

aaitor:
## Why

The A2A integration only interoperated with Nevermined's **own** SDK. The agent card was served at the pre-spec `/.well-known/agent.json`, but every current A2A client — including `@a2a-js/sdk`'s own default — discovers it at `/.well-known/agent-card.json`, so stock A2A clients 404'd on Nevermined agents (and vice-versa). And payment used a home-grown HTTP-header paywall rather than the standard, so a generic A2A agent couldn't pay a Nevermined agent. A2A's whole value is cross-vendor interoperability, so this defeated the purpose.

This is the A2A analog of the MCP in-band migration the team just shipped (#384), applying the same approach to the A2A transport.

### 1. Canonical agent-card discovery (interop fix)
- Serve the card at the canonical `/.well-known/agent-card.json`, keeping `/.well-known/agent.json` as a backward-compat **alias**.
- `PaymentsClient` defaults to the canonical path with a legacy fetch fallback.
- Bump `@a2a-js/sdk` `^0.3.4` → `^0.3.13` (stable line; 1.0 is still alpha).

### 2. In-band x402 v2 A2A transport (standards payment flow)
Per the [Coinbase x402 v2 A2A transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/a2a.md) + the official [a2a-x402 extension](https://github.com/google-agentic-commerce/a2a-x402):
- The agent card now declares the official extension URI (`.../a2a-x402/blob/main/spec/v0.2`) alongside `urn:nevermined:payment` (kept one release).
- **Payment required** → server returns an `input-required` Task with `x402.payment.status: payment-required` + `x402.payment.required` metadata (no HTTP 402).
- **Payment submitted** → the payload is read from the message's `x402.payment.payload` metadata, re-encoded via `encodeAccessToken` to the base64 token the facilitator verify/settle path already expects.
- **Settlement** is stamped in band (`payment-completed` + `x402.payment.receipts`). **A settlement failure forces `failed`/`payment-failed` and suppresses paid content — both the status message AND any artifacts — so a paid result is never delivered without settlement landing.**
- The legacy `payment-signature` header flow is kept as a **deprecated fallback (one release)**.

New module `src/a2a/x402-a2a.ts` ports the in-band task helpers (status enum, metadata-key constants, `createPaymentRequiredTask`/`recordPaymentSuccess`/`recordPaymentFailure`, …), mirroring the Python SDK.

## Backward compatibility
- Legacy `/.well-known/agent.json` still served (alias); legacy `payment-signature` header flow still works (deprecated).
- All changes additive — no existing public signature changed. `openclaw/` and `cli/` use no A2A APIs.

## Tests
- New coverage: first-contact → `input-required`; in-band payload → verify/execute/settle/receipts; verify/settle failure with **content + artifact** suppression; legacy-header regression.
- `pnpm build` + `pnpm lint` clean; **50 a2a / 332 full suite** green.

## Deferred (out of scope)
- **A2A 1.0** — `@a2a-js/sdk` 1.0 is still alpha; revisit when stable (keep TS/Python parity).
- **AP2 embedded flow** — the in-band structures are AP2-nestable; build when a counterparty exists.
- **Agent-card JWS signing** — production hardening, later.

## Pre-release note
The in-band token round-trip is verified at the codec level + against the mocked facilitator. A live-facilitator E2E pass on staging is worth running before release, same caveat as the MCP PRs.

### DIFF
--- a/src/a2a/paymentsRequestHandler.ts
+++ b/src/a2a/paymentsRequestHandler.ts
@@ -525,10 +525,10 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
             const redemptionConfig = await this.getRedemptionConfig()
 
             if (!redemptionConfig.useBatch) {
-              // Prefer the per-request context (in-band flag); fall back to a
-              // per-taskId lookup for executors that mint their own task id.
+              // Prefer the per-request context (authoritative in-band flag); fall
+              // back to a per-taskId lookup for executors that mint their own task id.
               const httpContext =
-                this.getHttpRequestContextForTask(event.taskId) ?? requestHttpContext
+                requestHttpContext ?? this.getHttpRequestContextForTask(event.taskId)
               // Execute redemption with server configuration for non-batch requests
               const response = await this.executeRedemption(
                 bearerToken,
@@ -562,7 +562,32 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
               }
             }
           } catch (err) {
-            // Do nothing
+            // x402 v2 A2A: settlement failed AFTER the paid event was already
+            // streamed to the client (a stream cannot retract it). Do NOT swallow
+            // it — log, and stamp payment-failed on the persisted task so
+            // tasks/get + resubscribe reflect the failure (mirrors non-streaming).
+            console.error('[PaymentsA2A] streaming settlement failed after execution:', err)
+            const httpContext =
+              requestHttpContext ?? this.getHttpRequestContextForTask(event.taskId)
+            if (httpContext?.inBand) {
+              try {
+                const task = resultManager.getCurrentTask()
+                if (task) {
+                  const errorReason = err instanceof Error ? err.message : String(err)
+                  if (task.status) task.status.state = 'failed'
+                  task.artifacts = undefined
+                  this.recordInBandSettlement(task, httpContext, {
+                    success: false,
+                    errorReason,
+                    transaction: '',
+                    network: '',
+                  })
+                  await resultManager.processEvent(task)
+                }
+              } catch (stampErr) {
+                console.error('[PaymentsA2A] failed to stamp streaming payment-failed:', stampErr)
+              }
+            }
           }
         }
 
@@ -627,7 +652,7 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
           // carries the in-band flag). Fall back to a per-taskId lookup for
           // executors that mint their own task id (the event's taskId may then
           // differ from the request's generated one).
-          const httpContext = this.getHttpRequestContextForTask(event.taskId) ?? requestHttpContext
+          const httpContext = requestHttpContext ?? this.getHttpRequestContextForTask(event.taskId)
           await this.handleTaskFinalization(resultManager, event, bearerToken, httpContext)
         }
 
@@ -817,9 +842,11 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
           }
         }
       } catch (err) {
-        // Settlement failed AFTER the agent executed. For the in-band x402 v2
-        // A2A path this is surfaced below (payment-failed + suppressed content);
-        // for the legacy header path the prior behavior (swallow) is preserved.
+        // Settlement failed AFTER the agent executed. Always log it (the legacy
+        // header path has no other surface for it). For the in-band x402 v2 A2A
+        // path it is additionally surfaced below as payment-failed + suppressed
+        // content; the legacy header path still delivers the result (no retract).
+        console.error('[PaymentsA2A] settlement failed after execution:', err)
         settlementError = err
       }
 
@@ -882,7 +909,9 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
           this.deleteHttpRequestContextForTask(event.taskId)
         }
       } catch (err) {
-        // Do nothing
+        // This block persists the payment-failed/suppressed task; if it throws the
+        // suppression itself failed, so log loudly rather than swallow.
+        console.error('[PaymentsA2A] failed to persist finalized task:', err)
       }
     }
     try {
--- a/src/a2a/paymentsRequestHandler.ts
+++ b/src/a2a/paymentsRequestHandler.ts
@@ -23,7 +23,12 @@ import { PaymentsError } from '../common/payments.error.js'
 import { StartAgentRequest, isValidScheme } from '../common/types.js'
 import { Payments } from '../payments.js'
 import { decodeAccessToken } from '../utils.js'
-import { buildPaymentRequired, type X402PaymentRequired } from '../x402/facilitator-api.js'
+import {
+  buildPaymentRequired,
+  type SettlePermissionsResult,
+  type X402PaymentRequired,
+} from '../x402/facilitator-api.js'
+import { x402A2AUtils } from './x402-a2a.js'
 import type {
   A2AAuthResult,
   A2AStreamEvent,
@@ -184,7 +189,9 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
     }
 
     const agentId = paymentExtension?.params?.agentId as string | undefined
-    const scheme = isValidScheme(decodedAccessToken.accepted?.scheme) ? decodedAccessToken.accepted.scheme : 'nvm:erc4337'
+    const scheme = isValidScheme(decodedAccessToken.accepted?.scheme)
+      ? decodedAccessToken.accepted.scheme
+      : 'nvm:erc4337'
 
     const paymentRequired: X402PaymentRequired = buildPaymentRequired(planId, {
       endpoint: endpoint || '',
@@ -265,7 +272,9 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
       throw PaymentsError.unauthorized('Plan ID not found in agent card.')
     }
 
-    const scheme = isValidScheme(decodedAccessToken.accepted?.scheme) ? decodedAccessToken.accepted.scheme : 'nvm:erc4337'
+    const scheme = isValidScheme(decodedAccessToken.accepted?.scheme)
+      ? decodedAccessToken.accepted.scheme
+      : 'nvm:erc4337'
 
     // Build paymentRequired using the helper
     const paymentRequired = buildPaymentRequired(planId, {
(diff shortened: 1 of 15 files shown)


r-marques (commented):

_Reviewed `origin/main...HEAD` (15 files, +1561/-69) with the pr-review-toolkit panel: code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer. Read-only; no code modified._

This is a well-built, well-documented PR. The non-streaming in-band flow is solid and the untrusted-payload handling is genuinely careful. The blocker below is confined to the **streaming** transport, where the hardening applied to the non-streaming path was not mirrored — and that gap is a money/paid-content leak that is invisible in production.

### 🔴 Blockers (1)

- **Streaming settlement failure is silently swallowed — paid content delivered, no charge recorded, nothing logged, no `payment-failed` stamped** — `src/a2a/paymentsRequestHandler.ts:565` (the `catch (err) { // Do nothing }` in `processStreamingEventsWithFinalization`). _(silent-failure-hunter C1 conf 0.97; code-reviewer #1/#2 conf 86/82; pr-test-analyzer #2 conf high)_

  On the streaming path the final event carrying paid content is `processEvent`'d and `yield`ed to the client **before** `executeRedemption` (settlement) runs. The in-band stamping (`recordInBandSettlement`) sits *inside* the same `try`, only on the success branch — so when settlement throws (facilitator 5xx, RPC/bundler latency, the #1508-class abandon, insufficient on-chain credits at settle time), control jumps straight to `// Do nothing`:
  - paid content was already streamed to the client;
  - **no** `payment-failed` is stamped on the persisted task (unlike the non-streaming path at `:830-877`), so even `tasks/get`/resubscribe shows it as completed;
  - **nothing is logged** — directly contravening the repo rule *"Always log errors before re-throwing… silent catch-and-rethrow hides root causes"* (CLAUDE.md Gotchas).

  This contradicts the PR's own doc promise (`markdown/a2a-integration.md`): *"on failure it is `failed` / `payment-failed` … and paid content is suppressed"* — which holds for non-streaming only. The "a stream cannot retract already-yielded content" comment at `:547-551` is a legitimate transport constraint, but it argues for **recording + logging** the failure on the persisted task, not for swallowing it entirely. Minimum fix: in the streaming catch, `console.error` the original error and stamp `payment-failed` + an error receipt onto the saved task (mirror `:830-877`). Stronger fix: withhold the final streaming event until settle resolves, or explicitly document that streaming delivers at verify-time (not settle-time) and reconciles settle failures out-of-band. This path is **entirely untested** (no `message/stream` test; `buildCard()` never sets `capabilities.streaming: true`; the e2e streaming block is commented out) — add a streaming settle-failure test that pins whatever behavior you choose.

### 🟡 Should fix (4)

- **Legacy header-path settlement failure is captured then never read or logged** — `src/a2a/paymentsRequestHandler.ts:823` (`settlementError = err`). It is only surfaced under `if (httpContext?.inBand …)` (`:830`, `:870`); on the legacy `payment-signature` header path (`inBand === false`) `settlementError` is dropped on the floor — paid content delivered, no charge, **no log**. This is the default path for every already-deployed agent. At minimum add `console.error(…)` in the catch regardless of `inBand`. _(silent-failure-hunter H2, conf 0.9)_
- **`recordInBandSettlement` stamps `payment-verified` whenever `settlement` is undefined — including the `useBatch` case** — `src/a2a/paymentsRequestHandler.ts:486-487`. When batch redemption is configured, `executeRedemption` never runs, `settlement` stays `undefined`, and the non-streaming path stamps `payment-verified` — telling the client "verified, no on-chain settlement" when settlement was merely deferred to a batch process this handler never confirms. Conflates "you owe nothing" with "you'll be charged out-of-band." Narrow the `else` branch to the genuine no-credit path, or add a distinct deferred marker. Worth a maintainer confirmation on intent. _(silent-failure-hunter H3, conf 0.78)_
- **Suppression-guard `try` swallows its own errors** — `src/a2a/paymentsRequestHandler.ts:885` (`// Do nothing`). This block persists the carefully-built `payment-failed` task (`resultManager.processEvent(task)`). If that throws, the suppression silently fails and the original (paid) content can surface — i.e. the meta-failure of the fix itself is swallowed. Add a `console.error`. _(silent-failure-hunter M4, conf 0.85)_
- **Cheap untested branches this PR introduces** — _(pr-test-analyzer #3/#4, conf high)_:
  - non-payment `input-required` should NOT trigger the client resubmit — `_paymentRequiredTaskId` (`src/a2a/paymentsClient.ts:365-375`) guards on `payment-required`, but the "does not resubmit" test only covers a `completed` response. Add a case: `input-required` with no/other payment status → assert one `message/send`, no double-pay.
  - `PaymentsClient._fetchAgentCard` legacy fallback-on-error (`src/a2a/paymentsClient.ts:89-94`) is never exercised (the `A2AClient.fromCardUrl` mock always resolves). Add canonical-fails→legacy-succeeds and both-fail cases.

### 💡 Good to have (5)

- **`_fetchAgentCard` fallback discards the canonical-path error** — `src/a2a/paymentsClient.ts:89-94`. On canonical failure the original `err` is dropped and legacy is tried; if both fail, only the **legacy** error is thrown, masking a canonical-path TLS/DNS/5xx root cause behind a clean legacy 404. Also silent (no warn on fallback — relevant to the typosquatting threat the adjacent docs warn about). `console.warn` on fallback + preserve `{ cause: err }`. _(silent-failure-hunter M5, conf 0.8)_
- **`recordInBandSettlement({ status: event.status } as any, …)` over-promises a full `Task`** — `src/a2a/paymentsRequestHandler.ts:849,857`. `ensureStatusMetadata` reads `task.id` to synthesize `messageId: "${task.id}-status"`; on the success path, an A2A executor that emits a terminal status with no `status.message` yields the literal `"undefined-status"`. Latent (current tests always set `status.message`) but reachable. Narrow the helper param to `{ id?: string; status?: TaskStatus }` and guard the messageId against a missing id. _(type-design-analyzer, conf 0.7)_
- **`HttpRequestContext` should be a discriminated union** — `src/a2a/types.ts:165-186`. The three real states (pre-payment / in-band-auth / header-auth) are modeled as independent optionals, so the type admits contradictory combos the producer never emits and forces `validation: undefined as any` in `server.ts`. A `mode`-discriminated union removes the `as any` and lets consumers `switch`. Reasonable as a focused follow-up; at minimum document the mutual-exclusivity invariant now. _(type-design-analyzer, conf 0.8)_
- **`getPaymentRequirementsFromMessage` returns `as unknown as X402PaymentRequired` after only an `isPlainObject` check** — `src/a2a/x402-a2a.ts:149`. Asserts a rich interface that the runtime never validates. Blast radius is limited today (the server builds `paymentRequired` locally via `buildPaymentRequired`, not via this getter), but add a real type guard or return `Record<string, any>`. _(type-design-analyzer, conf 0.75)_
- **In-band failure receipt always carries empty `network`/`transaction`** — `src/a2a/paymentsRequestHandler.ts:849-854` (`network: settlement?.network || ''`, where `settlement` is undefined on the throw path). Populate `network` from `decodedAccessToken.accepted.network` so the error receipt is actionable. Also: the redundant `X402A2AUtils` static-key mirror (`src/a2a/x402-a2a.ts:120-124`) is referenced only inside its own file — drop it or mark intentional. _(silent-failure-hunter M6 conf 0.55; type-design-analyzer LOW conf 0.9)_

### ✅ Strengths

- **Non-streaming settlement-failure suppression is exemplary** — replaces the status message, drops `task.artifacts`, and test 3 even asserts `JSON.stringify(task)` contains no paid text (belt-and-suspenders against leakage through any field). This is the standard the streaming path should meet.
- **`response.txHash ?? response.transaction` is a genuine bug fix** — `SettlePermissionsResult` defines only `transaction`; the old `response.txHash` was always `undefined` against the typed shape. Both call sites updated consistently.
- **Untrusted in-band payload is well-bounded** — `isPlainObject` + 64KB cap; the server derives `planId`/`agentId` from its *own* agent card (never client input), and `accepted.scheme` passes `isValidScheme` with a safe `nvm:erc4337` fallback. No plan/scheme-confusion escalation; the re-encoded token is verify/settle-validated identically to a header token.
- **No cross-tenant ordering oracle and no error-message leak introduced**; verification short-circuits to HTTP 402 before execution (test 3b). The client resubmit is safe — gated on both `input-required` state and `payment-required` status, so no double-settle and no resubmit on a legitimate input request.
- Additive `src/index.ts` exports; consistent ESM `.js` import extensions; in-tree `openclaw/`/`cli/` consumers unaffected. Header-deprecation warning logged once-per-process. Strong helper-layer and non-streaming coverage, with spec-key/lifecycle constants pinned against wire-format drift.

r-marques (commented):
Inline anchors for the streaming settle-failure blocker called out in the automated review above (the two facets, at the two code sites).

src/a2a/paymentsRequestHandler.ts
```diff
@@ -440,11 +538,28 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
...
               if (response && event.metadata) {
-                event.metadata.txHash = response.txHash
+                event.metadata.txHash = response.txHash ?? response.transaction
                 event.metadata.creditsCharged = response.amountOfCredits
                   ? Number(response.amountOfCredits)
                   : event.metadata.creditsUsed
               }
+
+              // x402 v2 A2A transport: stamp the settlement receipt onto the
+              // persisted task in band. NOTE: a stream cannot retract the event
```
r-marques, line 548: **Streaming delivers paid content before settlement resolves.** This NOTE is right that the stream can't retract the already-yielded event — but the consequence is a payment leak: in the verify-pass / settle-fail window (exactly where bundler/RPC latency lives, cf. nvm-monorepo #1508) the buyer keeps the streamed content with no on-chain charge. The non-streaming path avoids this by suppressing content on settle failure (`:830-877`); the stream structurally can't.

Consider buffering the final status/artifact event until `executeRedemption` resolves, or — if at-verify delivery is intended — documenting it explicitly: `markdown/a2a-integration.md` currently promises *"on failure … paid content is suppressed"*, which holds for the non-streaming path only.

src/a2a/paymentsRequestHandler.ts
```diff
@@ -440,11 +538,28 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
...
+                    task,
+                    httpContext,
+                    response as SettlePermissionsResult,
+                  )
+                  await resultManager.processEvent(task)
+                }
+              }
             }
           } catch (err) {
             // Do nothing
```
r-marques: **Blocker: this swallows the settlement failure entirely.** The in-band stamp via `recordInBandSettlement` (L552) runs only on the success branch, so when `executeRedemption` (L533) throws, control lands here and:

- nothing is logged — violates the repo rule *"Always log errors before re-throwing"* (CLAUDE.md Gotchas);
- no `payment-failed` / error-receipt is stamped on the persisted task, so `tasks/get` + resubscribe still report it `completed`.

Mirror the non-streaming path (`:830-877`):

```ts
} catch (err) {
  console.error(`[x402] streaming settlement failed for task ${event.taskId}:`, err)
  if (httpContext?.inBand) {
    const task = resultManager.getCurrentTask()
    if (task) {
      const errorReason = err instanceof Error ? err.message : String(err)
      this.recordInBandSettlement(task, httpContext, { success: false, errorReason, transaction: '', network: '' } as SettlePermissionsResult)
      await resultManager.processEvent(task)
    }
  }
}
```

This whole branch is currently untested (no `message/stream` test; `buildCard()` never sets `capabilities.streaming: true`) — add a streaming settle-failure test pinning whatever behavior you choose.

aaitor:
Thanks for the review — addressed in 9afba6b:

- **🔴 Streaming settlement failure (blocker)** — no longer swallowed. The streaming finalization catch now `console.error`s the failure and stamps `failed` / `payment-failed` (with artifacts cleared) onto the **persisted** task, so `tasks/get` + `resubscribe` reflect it. (A stream cannot retract an already-yielded event, so this is record-on-the-saved-task, as the path comment documents.)
- **Legacy header + non-streaming settle failures** — now logged (`console.error`) regardless of `inBand`.
- **Stale per-task context (Copilot)** — both sites now prefer the per-request context: `requestHttpContext ?? getHttpRequestContextForTask(...)`.
- **Suppression-guard persist failure** — now logged instead of swallowed.
- **`ponytail:` markers** — replaced with `TODO(a2a):`.

Tracked as follow-ups (not in this commit): a dedicated streaming settle-failure regression test; `recordInBandSettlement` stamping `payment-verified` for the `useBatch` case (wants a maintainer call on intended semantics); the type-design suggestions (`HttpRequestContext` discriminated union, `getPaymentRequirements` runtime guard); and `_fetchAgentCard` fallback warn + `{cause}` + its test.

src/a2a/paymentsRequestHandler.ts
aaitor: Fixed in 9afba6b and now tested in e32a68f. The streaming catch matches your suggested code: it `console.error`s the failure and stamps `payment-failed` (success:false receipt + cleared artifacts) on the persisted task via `recordInBandSettlement`. New test 5 in `x402-inband-flow.test.ts` drives `message/stream` with a failing settlement and asserts the persisted task (via `tasks/get`) is `failed`/`payment-failed` with content suppressed.

r-marques (approved):

Issue #393, opened 2026-06-17
A2A in-band x402 v2 — review follow-ups (from #392)


aaitor:
Non-blocking follow-ups raised during review of #392 (x402 v2 in-band A2A transport). All four are in the TS SDK (`src/a2a/`); the Python SDK (nevermined-io/payments-py#236) is already clean on these.

### Tracked

- [ ] **1. `useBatch` → `payment-verified` semantics.** When batch redemption is configured, `executeRedemption` never runs, so `recordInBandSettlement` stamps `payment-verified` on the in-band path — telling the client "verified, no on-chain settlement" when settlement was actually deferred to the batch process this handler never confirms. Decide intended semantics: narrow the no-settlement branch to the genuine no-credit case, or add a distinct "deferred" marker. `src/a2a/paymentsRequestHandler.ts` (`recordInBandSettlement`).

- [ ] **2. Type-design hardening.** (a) `HttpRequestContext` models three real states (pre-payment / in-band-auth / header-auth) as independent optionals, admitting contradictory combos and forcing `validation: undefined as any` in `server.ts` — make it a `mode`-discriminated union. (b) `getPaymentRequirements` returns `as unknown as X402PaymentRequired` after only an `isPlainObject` check — add a real runtime guard or return `Record<string, any>` (the Python sibling already validates via pydantic). `src/a2a/types.ts`, `src/a2a/x402-a2a.ts`.

- [ ] **3. `_fetchAgentCard` legacy fallback.** On canonical-path failure the original error is dropped; if both paths fail only the legacy error surfaces, masking a canonical TLS/DNS/5xx root cause, and the fallback is silent. Add `console.warn` on fallback + preserve `{ cause }`, and add a canonical-fails→legacy-succeeds (and both-fail) test. `src/a2a/paymentsClient.ts`.

- [ ] **4. Client non-payment `input-required` no-resubmit test.** `_paymentRequiredTaskId` guards resubmit on `payment-required`, but the "does not resubmit" test only covers a `completed` response. Add a case: `input-required` with no/other payment status → assert a single `message/send`, no double-pay. `src/a2a/paymentsClient.ts` + tests.

### Related (pre-release validation — not one of the four)
The in-band token round-trip is proven at the codec level and against the mocked facilitator, but the SDK e2e suites still exercise the *header* flow. A live-facilitator staging e2e of the in-band flow is worth running before customers depend on it (applies to both SDKs).

---
[...]

PR #394, opened 2026-06-17
fix(a2a): in-band x402 v2 review follow-ups (#393)

aaitor:
Addresses #393 — the four review follow-ups from #392's review. TS-only (the Python SDK is already clean on these).

## Changes

**1. Batch settlement gets a distinct deferred marker.** The no-settlement branch in `recordInBandSettlement` is reachable *only* in the batch case (the `creditsToBurn` guard already excludes free/no-credit calls), so the old "free / no-credit" comment was wrong and a bare `payment-verified` conflated "nothing owed" with "charged out-of-band". Batch now stamps `payment-verified` **plus** the Nevermined marker `x402.payment.settlement: "deferred"` (exported as `X402_SETTLEMENT_DEFERRED_KEY`; spec-only clients ignore it).

**2a. `HttpRequestContext` invariant documented + cast removed.** `validation` is now optional, the two mutually-exclusive states (payment-required vs authorized) are documented, the `validation: undefined as any` cast is gone, and the settlement consumer asserts the invariant explicitly. (A full `mode`-discriminated union remains a reasonable future refactor.)

**2b. `getPaymentRequirements` runtime guard.** Structurally validates (numeric `x402Version` + `accepts` array) before asserting `X402PaymentRequired`, instead of a blind `as unknown as` over arbitrary metadata.

**3. `_fetchAgentCard` fallback hardening.** Warns on the canonical→legacy fallback (a silent fallback to a typosquattable path is a discovery risk) and preserves the canonical-path root cause as `{ cause }` when both paths fail (a TLS/DNS/5xx is otherwise masked by a clean legacy 404).

## Tests (point 4 + guards for the above)
- `recordPaymentDeferred` sets the deferred marker; plain verify does not.
- `getPaymentRequirements` rejects non-conforming objects.
- Client does **not** resubmit on a non-payment `input-required` response (no double-pay).
- `_fetchAgentCard`: canonical-fails→legacy-succeeds (warns) + both-fail throws with the canonical `cause`.

56 a2a tests pass (+5); `pnpm build` + `pnpm lint` clean; full suite 344 pass. `openclaw/`/`cli/` use no A2A APIs; all changes additive.

## Note
The staging-e2e item in #393 is a separate **pre-release validation** (run the in-band flow against a live facilitator), not a code change — it's the gate before the patch release.

### DIFF
--- a/markdown/a2a-integration.md
+++ b/markdown/a2a-integration.md
@@ -128,6 +128,8 @@ Lifecycle:
 
 Metadata keys: `x402.payment.status`, `x402.payment.required`, `x402.payment.payload`, `x402.payment.receipts`, `x402.payment.error`.
 
+> **Batch (deferred) settlement.** If the server settles credits in batch, a successful call returns `x402.payment.status: "payment-verified"` (not `payment-completed`) plus the Nevermined marker `x402.payment.settlement: "deferred"` — the payment was verified but settled out-of-band, so there is no in-band receipt. Spec-only clients ignore the unknown key.
+
 > ⚠️ **Header flow deprecated.** The legacy `payment-signature` HTTP header still works but is now a **deprecated fallback, kept for one release**. New integrations should rely on the in-band metadata flow above.
 
 ## Implement Executor
--- a/src/a2a/paymentsClient.ts
+++ b/src/a2a/paymentsClient.ts
@@ -88,10 +88,27 @@ export class PaymentsClient extends A2AClient {
     try {
       return await load(agentCardPath)
     } catch (err) {
-      if (agentCardPath !== LEGACY_AGENT_CARD_WELL_KNOWN_PATH) {
-        return await load(LEGACY_AGENT_CARD_WELL_KNOWN_PATH)
+      if (agentCardPath === LEGACY_AGENT_CARD_WELL_KNOWN_PATH) {
+        throw err
       }
-      throw err
+      // Canonical path failed — fall back to the legacy path. Warn, because a
+      // silent fallback to a typosquattable legacy path is a discovery risk.
+      console.warn(
+        `[PaymentsA2A] agent card not found at '${agentCardPath}', falling back to '${LEGACY_AGENT_CARD_WELL_KNOWN_PATH}':`,
+        err instanceof Error ? err.message : String(err),
+      )
+      try {
+        return await load(LEGACY_AGENT_CARD_WELL_KNOWN_PATH)
+      } catch (legacyErr) {
+        // Both failed. Surface the legacy error but preserve the canonical-path
+        // root cause (a TLS/DNS/5xx there is otherwise masked by a clean 404).
+        const message = legacyErr instanceof Error ? legacyErr.message : String(legacyErr)
+        const wrapped = new Error(
+          `Failed to fetch agent card from '${agentCardPath}' or legacy '${LEGACY_AGENT_CARD_WELL_KNOWN_PATH}': ${message}`,
+        )
+        ;(wrapped as Error & { cause?: unknown }).cause = err
+        throw wrapped
+      }
     }
   }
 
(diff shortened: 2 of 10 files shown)


eruizgar91 (commented):
## Code Review

Clean, well-scoped follow-up to #392's review. Each of the four fixes is focused and carries a targeted test. The doc comments are genuinely good — they explain *why*, not *what*. 15/15 touched unit tests pass locally.

### Issue worth flagging — streaming batch path stamps nothing

The new deferred marker is only applied on the **non-streaming** path (`handleTaskFinalization`, ~line 884). The **streaming** path (`processStreamingEventsWithFinalization`, ~lines 525-563) skips the entire settlement-recording block when `useBatch` is true:

```ts
if (!redemptionConfig.useBatch) {
  ...
  if (httpContext?.inBand) { this.recordInBandSettlement(task, ...) }
}
// useBatch === true -> no recordInBandSettlement call at all
```

So a **streaming** client on a batch-settlement agent gets a `completed` task with **no** `payment-verified` and **no** `deferred` marker — the exact "reading a completed task as 'nothing owed'" ambiguity this PR set out to remove, just on the streaming transport. A stream can't retract already-yielded content, but it *can* still stamp the persisted task (visible via `tasks/get`), which is what the failure path at ~line 579 already does. Likely out of scope for this patch, but recommend either fixing here or filing a follow-up so the asymmetry is tracked rather than silent.

### Minor nits

- The PR description's claim that "the `creditsToBurn` guard already excludes free/no-credit calls" is slightly off — the guard at `paymentsRequestHandler.ts:811-818` only excludes `undefined`/`null`, so `creditsUsed === 0` still passes. Keeping the `recordPaymentVerified` else-branch as a defensive default is the right call; just the reasoning is imprecise.
- `new Error(msg, { cause: err })` is the idiomatic form vs. the manual `(wrapped as Error & { cause?: unknown }).cause = err` cast — purely cosmetic, fine as-is.
- The `mode`-discriminated-union follow-up noted in `types.ts:161` is the correct long-term fix and worth an issue so it isn't lost.

### Correctness & conventions
- `recordInBandSettlement` branch ordering (failure → success → deferred → verified) is correct and mutually exclusive.
- `getPaymentRequirementsFromMessage` structural guard is sound — treats malformed metadata as absent instead of asserting a rich interface over untrusted input.
- `X402_SETTLEMENT_DEFERRED_KEY` exported from `src/index.ts`, markdown doc updated, no openclaw/cli A2A signatures touched. ✅

**Verdict:** Approve once the streaming-batch gap is either addressed or explicitly deferred with a tracking issue. Everything else is clean.

aaitor:
Thanks @eruizgar91 — addressed in 1cca2e2:

**Streaming batch gap (the main one): fixed, not deferred.** The streaming finalization now has an `else` branch for `useBatch` that stamps `payment-verified` + the `x402.payment.settlement: "deferred"` marker on the **persisted** task (visible via `tasks/get` / resubscribe), mirroring the non-streaming path — the stream still can't retract already-yielded content. Added integration **test 6** (`streaming + batch redemption -> persisted task stamps payment-verified + deferred`), which asserts `settleCallCount === 0` and the marker on the persisted task.

**Nits:**
- *creditsToBurn guard reasoning* — you're right, the guard is a `!= null` check so `creditsUsed === 0` passes. The `recordPaymentVerified` else-branch is a defensive default; I corrected the code comment to say so (no behavior change).
- *`new Error(msg, { cause })`* — agreed it's the idiomatic form, but it doesn't type-check on our ES2020 `lib` (Error is typed with 0-1 args), so I replaced the manual cast with `Object.assign(new Error(msg), { cause })` — cleaner and it compiles.
- *`mode`-discriminated union* — filed as #395 so the asymmetry isn't lost.

57 a2a tests pass (+1), build + lint clean. Ready for re-review.

eruizgar91 (approved):
Re-reviewed after `1cca2e20`. The streaming-batch gap I flagged is now fixed symmetrically: the streaming `useBatch` branch stamps `payment-verified` + the deferred marker on the persisted task (`recordInBandSettlement(task, httpContext, undefined, true)`), matching the non-streaming path, and is covered by a dedicated integration test (`6. streaming + batch redemption -> persisted task stamps payment-verified + deferred`).

CI green (Lint & Build, Unit & Integration, CLI/OpenClaw sync); E2E pending is the staging pre-release gate, not a code blocker. Remaining nits (cosmetic `{ cause }` form, the `mode`-discriminated-union follow-up in types.ts) are non-blocking.

LGTM — approving.

PR #398, opened 2026-06-18
docs(mcp): plan-centric MCP server config (agentId optional)

r-marques:
## What

Updates the SDK doc source `markdown/mcp-integration.md` (the source of truth for the docs site's `api-reference/typescript/**`, regenerated by the release bot) so the next release sync carries the remaining plan-centric / agentId-optional wording to the site.

The bulk of the agentId-optional doc changes already landed with the agentId-optional feature. This PR fills the two remaining gaps plus one intro line:

- **Handler-options `planId` row** — now describes it as a per-handler override of the *required* server-level `planId` (was the stale "inferred from token").
- **In-band "Request — payment payload"** — clarifies that the `Authorization: Bearer <accessToken>` header authenticates the OAuth-protected MCP **session** (`initialize` returns `401` without it), which is separate from the in-band `_meta["x402/payment"]` **payment**; header-only payment is a deprecated one-release fallback. Adds a short client example.
- **"Configure MCP" intro** — "agent details" → "plan details".

## Why

The docs-site `api-reference/**` is generated from this repo — hand-editing it in the docs repo gets overwritten on the next release sync. This puts the change at the source so it flows correctly.

### DIFF
--- a/markdown/mcp-integration.md
+++ b/markdown/mcp-integration.md
@@ -33,7 +33,7 @@ These endpoints are generated automatically—no manual configuration required.
 
 ## Configure MCP
 
-Initialize the MCP integration with your agent details:
+Initialize the MCP integration with your plan details:
 
 ```typescript
 import { Payments, EnvironmentName } from '@nevermined-io/payments'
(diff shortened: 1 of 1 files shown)


eruizgar91 (approved):
## Review

Docs-only change, source-of-truth for the generated docs site. Prose edits (intro wording, `planId` handler-option row, session-auth-vs-payment clarification) are accurate and well-aligned with the agentId-optional feature. Approving.

### Non-blocking suggestion: the new example will throw at runtime

```typescript
const { accessToken } = await payments.x402.getX402AccessToken(planId) // agentId is optional
```

`getX402AccessToken` now **requires** `tokenOptions.delegationConfig` (`src/x402/token.ts:76-82`) — called with only `planId` it throws `PaymentsError.validation("delegationConfig is required ...")`. The "agentId is optional" comment is correct (optional 2nd positional arg), but the example omits the now-mandatory 3rd arg, so it fails immediately — the same stale-example problem this PR is otherwise fixing.

Canonical form (see `token.ts:59`):

```typescript
const { delegationId } = await payments.delegation.createDelegation({ /* ... */ })
const { accessToken } = await payments.x402.getX402AccessToken(planId, undefined, {
  delegationConfig: { delegationId },
}) // agentId (2nd arg) is optional
```

At minimum add `, undefined, { delegationConfig: { delegationId } }`.

### Verified OK
- `decodeAccessToken` import — exported via `export * from './utils.js'` (`src/index.ts:3`).
- `decodeAccessToken(accessToken)` returns the decoded `PaymentPayload`, matching what the MCP server reads (`src/mcp/core/paywall.ts:300`); `_meta["x402/payment"]` shape is consistent.
- `planId` row and session-auth-vs-payment prose match actual paywall/auth behavior.

PR #410, opened 2026-07-21
fix(mcp): PRM advertises the real issuer + 401s carry WWW-Authenticate (RFC 9728)

aaitor:
## Summary

Two MCP OAuth-discovery fixes so an agent (ChatGPT/Cursor connector, etc.) can actually walk the RFC 9728 discovery chain from a Nevermined MCP server to the authorization server. Part of the agentic-authorization epic (nvm-monorepo#2375, **Phase 0** — the one workstream that's blocked by nothing and lives in this repo).

### 1. PRM named the MCP server as its own authorization server

`buildProtectedResourceMetadata` / `buildMcpProtectedResourceMetadata` set `authorization_servers: [config.baseUrl]` — i.e. the MCP server points at **itself** as its AS — while `getOAuthUrls(...)` computed the correct Nevermined issuer and `void`-discarded it. An agent following `authorization_servers` was sent back to the resource, not to the server that mints tokens.

Both builders now advertise `oauthUrls.issuer`. `resource` stays the MCP server (RFC 9728: the server *is* the protected resource; its AS is Nevermined).

### 2. 401s carried no `WWW-Authenticate`, so the PRM was undiscoverable

`createRequireAuthMiddleware`'s three 401s had no challenge, so nothing pointed a client at the PRM (RFC 9728 §5.1 — the first link in the chain). They now emit:

```
WWW-Authenticate: Bearer resource_metadata="<proto>://<host>/.well-known/oauth-protected-resource"
```

derived from the request host (resolves regardless of proxying), **`resource_metadata` only** — no RFC 6750 `error` parameter, which would leak absent-vs-invalid-token. **No signature change** — the `openclaw/` and `cli/` in-tree consumers are unaffected.

## Test plan

- [x] `oauth_metadata_builders.test.ts` — `authorization_servers` now asserted to equal the real issuer and NOT contain `baseUrl` (the old tests locked in the bug)
- [x] `oauth_router_require_auth_middleware.test.ts` — 4 new tests: 401 carries the challenge on the requested host; uses request proto+host; `resource_metadata`-only (no `error=`); a valid token emits no challenge
- [x] `pnpm build` — clean · `pnpm lint` — clean · `pnpm test:unit` — **320 passed** (31 suites)

## Notes

- Behaviour change only — no public interface/type/signature change, so `openclaw`/`cli` compile unchanged.
- Docs (`markdown/mcp-integration.md`) list the endpoints but show no PRM body example, so nothing stale to update.

### DIFF
--- a/src/mcp/core/server-manager.ts
+++ b/src/mcp/core/server-manager.ts
@@ -259,6 +259,9 @@ export class McpServerManager {
         mountMcpHandlers(this.expressApp as any, {
           sessionManager: this.sessionManager,
           requireAuth: true,
+          // Pin the challenge's resource_metadata origin to the operator's known
+          // baseUrl (never client Host/X-Forwarded-* headers) — see #410 review.
+          baseUrl,
           log: this.log,
         })
 
--- a/src/mcp/http/mcp-handler.ts
+++ b/src/mcp/http/mcp-handler.ts
@@ -29,6 +29,13 @@ export interface McpHandlerConfig {
   sessionManager: SessionManager
   /** Whether to require authentication (default: true) */
   requireAuth?: boolean
+  /**
+   * Public base URL of this MCP server. When provided, the 401
+   * `WWW-Authenticate` challenge's `resource_metadata` pointer is built from it
+   * (never from client-supplied Host/X-Forwarded-* headers) — the correct,
+   * spoof-proof choice behind a TLS-terminating proxy.
+   */
+  baseUrl?: string
   /** Logger function */
   log?: (message: string) => void
 }
@@ -196,8 +203,11 @@ export function mountMcpHandlers(router: Router, config: McpHandlerConfig): void
 
   if (requireAuth) {
     // These handlers protect the `/mcp` resource, so a 401 must advertise the
-    // `/mcp`-scoped PRM (resource: `<baseUrl>/mcp`), not the root document.
+    // `/mcp`-scoped PRM (resource: `<baseUrl>/mcp`), not the root document. Pin the
+    // origin to the operator baseUrl when known, so the challenge never depends on
+    // client-supplied Host/X-Forwarded-* headers.
     const authMiddleware = createRequireAuthMiddleware({
+      baseUrl: config.baseUrl,
       resourceMetadataPath: '/.well-known/oauth-protected-resource/mcp',
     })
     router.post('/mcp', authMiddleware, postHandler as any)
--- a/src/mcp/http/mcp-handler.ts
+++ b/src/mcp/http/mcp-handler.ts
@@ -195,7 +195,11 @@ export function mountMcpHandlers(router: Router, config: McpHandlerConfig): void
   const deleteHandler = createDeleteMcpHandler(config)
 
   if (requireAuth) {
-    const authMiddleware = createRequireAuthMiddleware()
+    // These handlers protect the `/mcp` resource, so a 401 must advertise the
+    // `/mcp`-scoped PRM (resource: `<baseUrl>/mcp`), not the root document.
+    const authMiddleware = createRequireAuthMiddleware({
+      resourceMetadataPath: '/.well-known/oauth-protected-resource/mcp',
+    })
     router.post('/mcp', authMiddleware, postHandler as any)
     router.get('/mcp', authMiddleware, getHandler as any)
     router.delete('/mcp', authMiddleware, deleteHandler as any)
--- a/src/mcp/http/oauth-metadata.ts
+++ b/src/mcp/http/oauth-metadata.ts
@@ -119,11 +119,16 @@ export function buildProtectedResourceMetadata(config: OAuthConfig): ProtectedRe
  */
 function resolveAuthorizationServer(config: OAuthConfig): string {
   const { tokenUri } = getOAuthUrls(config.environment, config.oauthUrls)
-  // Guard an explicit `{ tokenUri: undefined }` override slipping through the
-  // `{...baseUrls, ...overrides}` spread (exactOptionalPropertyTypes is off) —
-  // fall back to the environment default rather than `new URL(undefined)`.
-  const resolved = tokenUri || getOAuthUrls(config.environment).tokenUri
-  return new URL(resolved).origin
+  const envTokenUri = getOAuthUrls(config.environment).tokenUri
+  // Guard both a falsy `{ tokenUri: undefined }` override (slips through the
+  // `{...baseUrls, ...overrides}` spread — exactOptionalPropertyTypes is off) AND
+  // a malformed non-falsy one — fall back to the environment default rather than
+  // crash the metadata endpoint with `new URL(...)`.
+  try {
+    return new URL(tokenUri || envTokenUri).origin
+  } catch {
+    return new URL(envTokenUri).origin
+  }
 }
 
 /**
--- a/src/mcp/http/oauth-metadata.ts
+++ b/src/mcp/http/oauth-metadata.ts
@@ -94,18 +94,32 @@ export function getOAuthUrls(
  */
 export function buildProtectedResourceMetadata(config: OAuthConfig): ProtectedResourceMetadata {
   const scopes = config.scopes || [...DEFAULT_SCOPES]
-  // RFC 9728: `resource` is THIS server (the protected resource); the
-  // `authorization_servers` are the Nevermined issuer that actually mints tokens
-  // — NOT this MCP server. The issuer is derived from the environment.
-  const oauthUrls = getOAuthUrls(config.environment, config.oauthUrls)
 
   return {
     resource: config.baseUrl,
-    authorization_servers: [oauthUrls.issuer],
+    authorization_servers: [resolveAuthorizationServer(config)],
     scopes_supported: scopes,
     bearer_methods_supported: ['header'],
     resource_documentation: `${config.baseUrl}/`,
   }
+}
+
+/**
+ * Resolve the authorization-server identifier for `authorization_servers`.
+ *
+ * RFC 9728 §2 requires each entry to be an AS issuer identifier — a URL from
+ * which `/.well-known/oauth-authorization-server` (RFC 8414) actually resolves.
+ * For Nevermined that is the **backend** API origin (which serves the AS
+ * metadata and whose own published `issuer` is the backend URL), NOT:
+ *   - `config.baseUrl` — that is this MCP server, the protected *resource*; and
+ *   - `OAuthUrls.issuer` — that is the *frontend* (a client-routed SPA that 200s
+ *     on every path, so an RFC 8414 fetch there silently returns HTML).
+ * The backend origin is derived from `tokenUri` (`${backend}/oauth/token`), which
+ * respects any `config.oauthUrls.tokenUri` override.
+ */
+function resolveAuthorizationServer(config: OAuthConfig): string {
+  const { tokenUri } = getOAuthUrls(config.environment, config.oauthUrls)
+  return new URL(tokenUri).origin
 }
 
 /**
@@ -129,13 +143,12 @@ export function buildMcpProtectedResourceMetadata(
   config: OAuthConfig,
 ): McpProtectedResourceMetadata {
   const scopes = config.scopes || [...DEFAULT_SCOPES]
-  // See buildProtectedResourceMetadata: `authorization_servers` is the Nevermined
-  // issuer, not this MCP server.
-  const oauthUrls = getOAuthUrls(config.environment, config.oauthUrls)
 
   return {
     resource: `${config.baseUrl}/mcp`,
-    authorization_servers: [oauthUrls.issuer],
+    // See resolveAuthorizationServer: the AS is the backend API origin, not this
+    // MCP server (baseUrl) nor the frontend SPA (OAuthUrls.issuer).
+    authorization_servers: [resolveAuthorizationServer(config)],
     scopes_supported: scopes,
     scopes_required: scopes,
     bearer_methods_supported: ['header'],
--- a/src/mcp/http/oauth-metadata.ts
+++ b/src/mcp/http/oauth-metadata.ts
@@ -119,7 +119,11 @@ export function buildProtectedResourceMetadata(config: OAuthConfig): ProtectedRe
  */
 function resolveAuthorizationServer(config: OAuthConfig): string {
   const { tokenUri } = getOAuthUrls(config.environment, config.oauthUrls)
-  return new URL(tokenUri).origin
+  // Guard an explicit `{ tokenUri: undefined }` override slipping through the
+  // `{...baseUrls, ...overrides}` spread (exactOptionalPropertyTypes is off) —
+  // fall back to the environment default rather than `new URL(undefined)`.
+  const resolved = tokenUri || getOAuthUrls(config.environment).tokenUri
+  return new URL(resolved).origin
 }
 
 /**
--- a/src/mcp/http/oauth-metadata.ts
+++ b/src/mcp/http/oauth-metadata.ts
@@ -94,12 +94,14 @@ export function getOAuthUrls(
  */
 export function buildProtectedResourceMetadata(config: OAuthConfig): ProtectedResourceMetadata {
   const scopes = config.scopes || [...DEFAULT_SCOPES]
-  // oauthUrls calculated but not used in this metadata (kept for future use)
-  void getOAuthUrls(config.environment, config.oauthUrls)
+  // RFC 9728: `resource` is THIS server (the protected resource); the
+  // `authorization_servers` are the Nevermined issuer that actually mints tokens
+  // — NOT this MCP server. The issuer is derived from the environment.
+  const oauthUrls = getOAuthUrls(config.environment, config.oauthUrls)
 
   return {
     resource: config.baseUrl,
-    authorization_servers: [config.baseUrl],
+    authorization_servers: [oauthUrls.issuer],
     scopes_supported: scopes,
     bearer_methods_supported: ['header'],
     resource_documentation: `${config.baseUrl}/`,
@@ -127,12 +129,13 @@ export function buildMcpProtectedResourceMetadata(
   config: OAuthConfig,
 ): McpProtectedResourceMetadata {
   const scopes = config.scopes || [...DEFAULT_SCOPES]
-  // oauthUrls calculated but not used in this metadata (kept for future use)
-  void getOAuthUrls(config.environment, config.oauthUrls)
+  // See buildProtectedResourceMetadata: `authorization_servers` is the Nevermined
+  // issuer, not this MCP server.
+  const oauthUrls = getOAuthUrls(config.environment, config.oauthUrls)
 
   return {
     resource: `${config.baseUrl}/mcp`,
-    authorization_servers: [config.baseUrl],
+    authorization_servers: [oauthUrls.issuer],
     scopes_supported: scopes,
     scopes_required: scopes,
     bearer_methods_supported: ['header'],
--- a/src/mcp/http/oauth-router.ts
+++ b/src/mcp/http/oauth-router.ts
@@ -317,30 +317,35 @@ export function createJsonMiddleware() {
  */
 export function createRequireAuthMiddleware() {
   return function requireAuthMiddleware(req: Request, res: Response, next: NextFunction) {
+    // RFC 9728 §5.1: a 401 from a protected resource MUST advertise where its
+    // Protected Resource Metadata lives, so a client can discover the authorization
+    // server and complete the OAuth flow. We point at THIS resource's PRM on the
+    // host the client actually reached (so it resolves regardless of proxying).
+    // `resource_metadata` ONLY — no RFC 6750 `error` param, which would leak
+    // absent-vs-invalid token.
+    const send401 = (error_description: string): void => {
+      const host = req.get('host')
+      if (host) {
+        res.setHeader(
+          'WWW-Authenticate',
+          `Bearer resource_metadata="${req.protocol || 'https'}://${host}/.well-known/oauth-protected-resource"`,
+        )
+      }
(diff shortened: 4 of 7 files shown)


aaitor:
Pushed `b9659d7` after an adversarial security pass caught **two real defects in the first cut** (thanks to the review — my initial 'validated e2e' claim was wrong; the existing e2e test only checks internal consistency, not that the issuer resolves):

**1. `authorization_servers` pointed at the frontend, not the backend.** `getOAuthUrls().issuer` is the **frontend** SPA (200s HTML on every path), while the real RFC 8414 AS metadata is served by the **backend** API. Now derived from the backend origin (`tokenUri.origin`) via `resolveAuthorizationServer()`. Tests now pin the hardcoded backend value and assert it's neither `baseUrl` nor the frontend issuer.

**2. `WWW-Authenticate` had a Host-header injection.** The raw `Host` was interpolated into the quoted-string, so `Host: evil";x="` broke out and injected a second `resource_metadata` auth-param at an attacker origin. Now: an optional operator-pinned `baseUrl` (never client input) wins; otherwise the (`x-forwarded-`)Host is used **only if it matches a safe Host grammar**, else the header is omitted. Also honours `x-forwarded-proto/host` behind a TLS-terminating proxy, and adds RFC 6750 `error=invalid_request` on malformed-credential branches. Signature stays backward-compatible.

**324 unit tests** (incl. an injection regression guard), build + lint clean. The only red check is **`E2E`**, failing on 3 unrelated `X402 Card Delegation (Stripe)` tests where staging returns `network: braintree` — my diff touches zero X402/Stripe code (main passed these in June); it's a staging-config drift, not this PR.

r-marques (commented):

_Reviewed `origin/main...HEAD` (4 files) with the pr-review-toolkit panel (code-reviewer, silent-failure-hunter, pr-test-analyzer), plus live verification against the deployed Nevermined OAuth endpoints, execution of the metadata builders, and a trace through the vendored `@modelcontextprotocol/sdk` client._

The direction of this change is right — a protected resource must not advertise itself as its own authorization server, and the RFC 9728 §5.1 challenge was genuinely missing. But the issuer value it now advertises serves no authorization-server metadata in any environment, and we traced a hard client-side failure. One blocker, and it's a one-line value change plus a deliberate decision.

### 🔴 Blocker (1)

- **`authorization_servers` advertises the frontend URL, which serves no AS metadata → OAuth discovery hard-fails** — `src/mcp/http/oauth-metadata.ts:104`, `:138` — _(confidence: high — verified against live endpoints, by executing the builder, and by tracing the official SDK)_

  `getOAuthUrls().issuer` resolves to `envConfig.frontend` (`oauth-metadata.ts:31`, `issuer: frontend`). Executing the real builder:

  ```
  sandbox -> ["https://nevermined.app"]
  custom  -> ["http://localhost:4200"]
  ```

  Against the live deployment:

  | URL | Result |
  |---|---|
  | `https://nevermined.app/.well-known/oauth-authorization-server` | `200` **`text/html`** — the SPA `index.html` |
  | `https://nevermined.dev/.well-known/oauth-authorization-server` | `200` **`text/html`** — same |
  | `https://api.sandbox.nevermined.app/.well-known/oauth-authorization-server` | `200` **`application/json`**, `"issuer": "https://api.sandbox.nevermined.app"` |

  The **backend** is the authorization server and names itself as the issuer; the frontend only hosts `authorization_endpoint` (`https://nevermined.app/oauth/authorize`), which is a field *inside* the backend's document. Sending `Accept: application/json` + `MCP-Protocol-Version` still returns HTML, so no content negotiation rescues it.

  **Exact break path** in the official `@modelcontextprotocol/sdk` (vendored in the worktree):
  1. `client/auth.js:668` takes `authorization_servers[0]` → `https://nevermined.app`
  2. `discoverAuthorizationServerMetadata` (`auth.js:607`) fetches `/.well-known/oauth-authorization-server`
  3. The response is **200**, so `response.ok` is true → the `4xx → continue` fallback at `auth.js:627` is **skipped**
  4. `OAuthMetadataSchema.parse(await response.json())` at `auth.js:634` → **`response.json()` throws on HTML**

  The throw escapes the loop; the fallback at `auth.js:675` only covers a failed *PRM* fetch, not a failed AS-metadata fetch. No recovery.

  This is a **working → hard-failing regression**: pre-PR, `authorization_servers: [config.baseUrl]` resolved to this router's own valid-JSON AS document (`oauth-router.ts:140`). In fairness the old state was *also* non-compliant — that document declares `issuer: <frontend>`, which doesn't match its retrieval URL, so strict clients reject it on RFC 8414 §3.3 issuer mismatch. Both are broken; this PR fixes the wrong half.

  Aggravating case: **`environment: 'custom'`** (self-hosted, no `NVM_FRONTEND_URL`) now publishes **`http://localhost:4200`** as the authorization server to remote clients. `Environments` is a module-load-time const, so setting `NVM_FRONTEND_URL` after importing the SDK is silently ignored.

  Suggested fix: advertise `Environments[env].backend` (trailing slash stripped), not `.frontend`. Two things to decide deliberately rather than in passing:
  - `OAuthUrls.issuer` is itself mislabeled — `buildAuthorizationServerMetadata` (`:171`) and `buildOidcConfiguration` (`:207`) also emit `issuer: <frontend>`, disagreeing with the deployed backend. Fixing only the PRM leaves those inconsistent.
  - **DCR discoverability**: once clients follow an external issuer, this router's own `POST /register` (`oauth-router.ts:168`) becomes undiscoverable, and the backend AS advertises **no** `registration_endpoint` (verified). That needs a plan even after the issuer value is corrected.

### 🟡 Should fix (3)

- **The 401 challenge points at the root PRM, not the `/mcp`-scoped one** — `src/mcp/http/oauth-router.ts:331` — _(confidence: high)_

  The middleware is mounted exclusively on `/mcp` (`mcp-handler.ts:198-201`), so every 401 carrying this challenge concerns the resource `<baseUrl>/mcp`. The challenge always emits `/.well-known/oauth-protected-resource`, whose document has `resource: config.baseUrl`. The correctly-scoped document **already exists**, served at `oauth-router.ts:128` with `resource: ${baseUrl}/mcp` — it just isn't advertised.

  To be precise about severity: the official SDK does **not** hard-reject this — `checkResourceAllowed` (`shared/auth-utils.js`) is a prefix match, so `/` vs `/mcp/` passes. Real consequences anyway: (a) `selectResourceURL` returns `resourceMetadata.resource` (`auth.js:343`), so the client sends the **origin** as the RFC 8707 `resource` parameter → tokens end up audience-bound more broadly than intended; (b) the client never sees `scopes_required` / `mcp_capabilities`; (c) stricter clients may reject outright.

- **The proxy comment is inaccurate in both halves of the URL it builds** — `src/mcp/http/oauth-router.ts:322-333` — _(confidence: high, verified against `express@4.21.2 lib/request.js`)_

  The comment claims the challenge "resolves regardless of proxying". In fact:
  - `req.protocol` honors `X-Forwarded-Proto` **only under `trust proxy`** — and `grep -rn "trust proxy"` returns **nothing** in this repo, while the SDK creates the Express app itself (`managed-server.ts:63`, `:230`, `server-manager.ts:216`), so it owns the setting and doesn't set it. Behind Cloudflare/ALB/ingress this advertises `http://…`.
  - `req.get('host')` is a plain header lookup and is **not** `X-Forwarded-Host`-aware (only `req.hostname` is, and it strips the port).

  So the scheme and the host describe *different hops*. Behind a proxy that rewrites Host upstream (nginx's default `proxy_set_header Host $proxy_host`), the challenge silently advertises an internal, externally-unresolvable host. Also `req.protocol || 'https'` is dead code in real Express — the fallback only ever fires against the test mock, giving false comfort.

- **Client-controlled `Host` reflected into the challenge, while `baseUrl` was available** — `src/mcp/http/oauth-router.ts:327-331` — _(high on the design gap, medium-low on exploitability)_

  The raw Host header is interpolated into a quoted-string header value with no validation. Live-verified emitted headers:

  ```
  Host: a".com                     -> Bearer resource_metadata="https://a".com/.well-known/..."
  Host: evil.com/p", scope="admin  -> Bearer resource_metadata="https://evil.com/p", scope="admin/.well-known/..."
  ```

  Most pointedly, `Host: evil.com/", error="invalid_token` emits:

  ```
  Bearer resource_metadata="https://evil.com/", error="invalid_token/.well-known/oauth-protected-resource"
  ```

  — injecting exactly the `error=` param this diff's own comment at `:324` declares it must never emit. The test at `:247` asserts `not.toContain('error=')`, but only for a benign host, so it cannot see this.

  This is *not* CRLF response-splitting — we verified Node's llhttp parser rejects CR, bare LF, NUL and DEL in a Host header with a 400 **before** the handler runs, so `setHeader` never throws and there is no 401→500 path. Exploitability alone is low (a spoofed Host poisons only the attacker's own 401; no `Cache-Control` on 401 makes intermediary caching unlikely). It matters because the downstream shape of an attacker-influenced auth-discovery path is severe, and because the fix is shared: `createOAuthRouter(options)` already destructures `baseUrl` at `:69`, and `createRequireAuthMiddleware()` is the only zero-arg factory. Threading `baseUrl` through fixes this, the protocol bug, and the PRM-path bug in one change.

### 💡 Good to have (3)

- **The `error_description` comment asserts a security property the code doesn't provide — and doesn't need** — `src/mcp/http/oauth-router.ts:324-325`, `:339/343/348` — the comment says `resource_metadata` ONLY, "no RFC 6750 `error` param, which would leak absent-vs-invalid token", while the body returns three distinct `error_description` values carrying exactly that distinction. But this is **not** a leak: all three states are properties of the caller's *own* request, already known to them — no server-side secret, no cross-user oracle. Moreover this middleware never validates the token (`:352-354`, deferred to `withPaywall()`), so the absent-vs-*invalid* distinction the comment names cannot arise here at all. Omitting `error=` is fine per RFC 6750; just fix the **rationale**, since a false security justification in an auth path invites someone to "preserve" it later.

- **`if (host)` silently drops the challenge — reachable, not theoretical** — `src/mcp/http/oauth-router.ts:327-334` — verified with raw sockets against Node v22: `GET / HTTP/1.0` with no Host yields `host === undefined`, and an empty `Host:` value yields `''`; Node passes **both** to the handler. The 401 then goes out bare, with no log and no fallback — the exact failure this PR exists to fix, silently. Auth still rejects correctly, so this is fail-open on *discoverability* only. A real fallback needs the same `baseUrl` threading as above; meanwhile the file's existing `onLog` pattern makes a warning a one-liner.

- **No test in the repo can catch the blocker — proven, not asserted** — `tests/unit/mcp/oauth_metadata_builders.test.ts:34`, `:70` — `expect(metadata.authorization_servers).toEqual([getOAuthUrls(baseConfig.environment).issuer])` computes the expectation with the same call the implementation makes. We `jest.mock`-ed `src/environments.ts` to force `staging_sandbox.frontend = 'https://TOTALLY-WRONG.example'` and ran the PR's assertions verbatim: **both still pass**, as does the pre-existing `expect(urls.issuer).toBeDefined()` at `:289`/`:299`. The test pins nothing.

  The adjacent `.not.toContain(baseConfig.baseUrl)` is the only real signal — it catches an exact revert to `[config.baseUrl]` and nothing else; it passes even when the issuer is `undefined` or `''`. No hardcoded issuer assertion exists anywhere in `tests/`. The e2e check at `tests/e2e/test_mcp_oauth_e2e.test.ts:317-319` only asserts non-empty, so it wouldn't have caught the *original* bug either. Fix: pin the literal, ideally a `test.each` over all five environments.

  Complete list of branches this diff introduces that no test can reach: the `if (host)` false branch (`oauth-router.ts:328` — the mock's `opts.host ?? 'mcp.example.com'` means `{host: undefined}` silently yields the default, so **no call site can produce it**); the `|| 'https'` fallback (`:331`); proxied scheme (`X-Forwarded-Proto`); hostile `Host`; a `config.oauthUrls` override on the two changed builders (it works — `{issuer:'https://custom-issuer.com'}` → `['https://custom-issuer.com']` — but the only override tests hit *other* builders, and it became load-bearing in this diff); and any environment other than `staging_sandbox`, which every PRM test hardcodes at `:19` despite the fix's whole premise being that the issuer is environment-derived.

  Related, same area: an explicit `{ issuer: undefined }` survives the `{...baseUrls, ...overrides}` spread (`oauth-metadata.ts:75`; `tsconfig.json` sets `strict` but not `exactOptionalPropertyTypes`) and serves `authorization_servers: [null]` with a 200. Confirmed by execution; impossible before this PR, since the value was discarded. A `?? baseUrls.issuer` guard is cheap.

### ✅ Strengths

- The RFC 9728 §5.1 gap was real, and the challenge is exactly the mechanism the official SDK keys on (`extractWWWAuthenticateParams`, `auth.js:348`). Closing it is the right call.
- Correctly identifies the core defect: a protected resource naming *itself* as its authorization server is wrong under RFC 9728. The conceptual direction is sound and worth landing once the value is right.
- The `send401` closure is the right shape — it makes the header impossible to forget on a future branch, and the refactor introduces no behavior drift in the existing paths.
- The negative test at `tests/unit/mcp/oauth_router_require_auth_middleware.test.ts:249-262` ("a valid token does NOT emit a WWW-Authenticate header") is the assertion usually missing from work like this.
- Replaces two `void getOAuthUrls(...)` "kept for future use" dead calls with real uses, with a why-comment — strictly better than the placeholder.
- Comments cite specific RFC sections, which made this review substantially faster.
- No security fail-open anywhere: authentication still rejects correctly on every path examined; the issues are discovery/metadata correctness.
- Suite is green and nothing went stale — 44/44 in the changed files, 107 across `tests/unit/mcp`, 320 passed / 1 skipped across all 31 unit suites (verified locally). No pre-existing test asserted the old value, so there's no broken-test debt here.

### One-line summary for the author

Findings 2–4 share a single fix: thread the configured `baseUrl` into `createRequireAuthMiddleware` and build the challenge from it plus the protected path, instead of from `req.protocol` / `req.get('host')`. The blocker is separate — `.backend` not `.frontend`, plus a decision about what `OAuthUrls.issuer` means and about DCR discoverability.

aaitor:
Thanks @r-marques — thorough. Timing note first: your review landed at 13:38 stamped against `b9659d7`, but the findings describe **`b8491cc`** (the prior commit) — GitHub stamps the review with current HEAD. `b9659d7` (pushed ~2 min before) had already independently fixed the blocker and two of the three should-fixes. `aa9075b` now covers the rest.

**🔴 Blocker — authorization_servers = frontend → already fixed in `b9659d7`.** It no longer uses `getOAuthUrls().issuer` (frontend SPA); it derives the **backend** origin via `resolveAuthorizationServer()` (`new URL(tokenUri).origin`). Your live trace (backend serves the JSON AS doc, frontend serves HTML) is exactly right and is what drove the fix.

**🟡 Challenge points at root PRM, not /mcp — fixed now (`aa9075b`).** `createRequireAuthMiddleware` takes an optional `resourceMetadataPath`; `mcp-handler` passes `/.well-known/oauth-protected-resource/mcp` so the client gets the correctly-scoped `resource` (RFC 8707) + `scopes_required`/`mcp_capabilities`.

**🟡 Host injection + proxy — fixed in `b9659d7`.** The raw-Host interpolation is gone: a safe Host grammar (`SAFE_HOST_RE`) is required or the header is omitted (injection regression test added, incl. your `evil.com/", error="…` payload). It reads `x-forwarded-host`/`x-forwarded-proto` **explicitly** (not via `req.get('host')`/`req.protocol`), and an operator-pinned `baseUrl` option bypasses client input entirely.

**💡 error_description rationale — fixed in `b9659d7`.** You're right it wasn't a leak. Now grounded in RFC 6750: `error="invalid_request"` on the malformed-credential branches, omitted when no credential is presented — and the comment says so.

**💡 tautological tests / `{issuer:undefined}`→`[null]` — fixed.** Hardcoded backend literal + a `test.each` over all four stable environments (asserts backend, forbids frontend + baseUrl). Since the value now derives from `tokenUri`, I guarded an explicit `{ tokenUri: undefined }` override → env-default fallback (test added), so no `new URL(undefined)`/`[null]`.

**Deferred, with reasons** (flagged so they're not lost):
- **baseUrl caller-threading** — `server-manager.ts:259` has no ready `baseUrl` (would need it added to `McpHandlerConfig` + derived from port/host); request-derivation is now validated + `x-forwarded`-aware and the pin option exists, so this is robustness, not a hole. The silent `if(host)` drop-with-no-log rides on the same thread.
- **`OAuthUrls.issuer` mislabeled in `buildAuthorizationServerMetadata`/`buildOidcConfiguration`** — real and pre-existing, but out of this PR's scope (PRM only).
- **DCR discoverability** — genuine architectural gap once clients follow the backend issuer (which advertises no `registration_endpoint`). Flagging for the epic (nvm-monorepo#2375) rather than solving in passing.

330 unit tests, build + lint clean, backward-compatible (openclaw/cli/mcp-handler).

r-marques:
## 🤖 Follow-up review — ✅ Blocker resolved

_Re-checked `b8491cc..aa9075b` (commits `b9659d7`, `aa9075b`). Verified by executing the builders and re-querying the live Nevermined OAuth endpoints, not by reading the diff alone._

Thanks — this addresses everything, and in two places the fix is better than what I suggested. Deriving the AS origin from `tokenUri` via `new URL(tokenUri).origin` (rather than reading `Environments[env].backend` directly) means a `config.oauthUrls.tokenUri` override flows through correctly, which my suggestion would have silently ignored. Likewise, parsing `x-forwarded-proto`/`x-forwarded-host` explicitly plus the operator-pinned `baseUrl` is stronger than the `app.set('trust proxy')` I proposed — it doesn't depend on the consumer configuring Express at all.

### ✅ Blocker — fixed and verified end-to-end

Executed `buildProtectedResourceMetadata` per environment and checked each result against what that host actually serves:

| environment | now advertises | live `issuer` at that host |
|---|---|---|
| `sandbox` | `https://api.sandbox.nevermined.app` | ✅ exact match, `application/json` |
| `live` | `https://api.live.nevermined.app` | ✅ exact match, `application/json` |
| `staging_sandbox` | `https://api.sandbox.nevermined.dev` | ✅ exact match, `application/json` |
| `custom` | `http://localhost:3001` | configured backend — correct for self-hosted |

RFC 8414 §3.3 issuer-match now holds in every deployed environment, the `@modelcontextprotocol/sdk` `response.json()`-on-HTML failure path is gone, and the `custom` env no longer publishes `http://localhost:4200` as a public authorization server.

Also confirmed fixed: the `/mcp`-scoped PRM (`resourceMetadataPath`, wired at `mcp-handler.ts:200`); Host injection (`SAFE_HOST_RE` omits the header rather than emitting an injectable one — the `Host: evil.com/", error="invalid_token` case no longer produces a challenge); the `{ tokenUri: undefined }` fallback guard; and the `error=` rationale, which now correctly splits RFC 6750 §3 (no credential → no error param) from §3.1 (malformed credential → `error="invalid_request"`). The tautological assertions are replaced with pinned literals across a `test.each` over all four environments, plus new coverage for injection, forwarded headers, pinned-baseUrl-beats-forged-Host, scoped path, and the error-param split.

### 🟡 Three residuals — none blocking, no need to hold the merge on my account

1. **`baseUrl` isn't plumbed at the SDK's own call site** — `mcp-handler.ts:200` passes only `resourceMetadataPath`, so in the managed path the challenge origin still comes from request headers. Since `x-forwarded-host` is now *preferred* over `Host` with no trust-proxy gate, a client can steer its own challenge to any `SAFE_HOST_RE`-valid host via a spoofed `X-Forwarded-Host` — a header this code previously ignored entirely. Injection is blocked and it's self-poisoning only, so impact is low, but the documented mitigation (`options.baseUrl`) isn't actually wired up in the path the SDK controls.

2. **DCR discoverability still looks open** — the backend AS advertises no `registration_endpoint` (re-verified), so this router's own `POST /register` (`oauth-router.ts:168`) becomes undiscoverable now that clients follow an external issuer. Pre-PR they got it from the local AS document. Is that tracked separately, or intentionally dropped?

3. **`new URL(resolved).origin` throws on a malformed non-falsy `tokenUri` override** — only the falsy case is guarded. Very minor.

### Test status

315 passed / 1 skipped, 30 of 31 unit suites green. The single failing suite is `tests/unit/x402/langchain-decorator.test.ts` — `Cannot find module '@langchain/langgraph/prebuilt'`, a missing optional peer dep in my local checkout. I confirmed it fails identically on `origin/main`, so it is **pre-existing and unrelated to this PR**.

r-marques (approved):

aaitor:
Thanks for the approval and the end-to-end re-verification. Closed the two actionable residuals in `9a5d875` (the third — DCR — is design, tracked below):

1. **`baseUrl` now plumbed at the SDK's own call site.** `server-manager.start()` already computes `baseUrl` (`:203`, `config.baseUrl || http://localhost:${port}`); it's now threaded → `mountMcpHandlers` (`McpHandlerConfig.baseUrl`, optional) → `createRequireAuthMiddleware`. So in the managed path the challenge origin comes from the operator baseUrl, never `Host`/`X-Forwarded-*`. The spoofed-`X-Forwarded-Host` self-steer you flagged is gone in the path the SDK controls; request-derivation (still validated) only remains for a bare `createRequireAuthMiddleware()`.

3. **Malformed non-falsy `tokenUri`** — `resolveAuthorizationServer` now `try/catch`es `new URL()`, falling back to the env default (test.each covers both `undefined` and `'not a url'`).

2. **DCR discoverability** — not dropped; it's a real gap once clients follow the backend issuer (which advertises no `registration_endpoint`). Recorded on the agentic-auth epic [nvm-monorepo#2375](https://github.com/nevermined-io/nvm-monorepo/issues/2375#issuecomment-5034828317) as a concrete Phase-2/4a (RFC 7591 DCR) input, since it's an architectural decision rather than a one-liner here.

Public `createRequireAuthMiddleware` signature unchanged; `McpHandlerConfig.baseUrl` is optional → openclaw/cli/mcp-handler unaffected. **331 unit tests**, build + lint clean. The only red check remains the unrelated `X402 Card Delegation (Stripe)` E2E (staging `braintree`), not this diff.

PR #415, opened 2026-07-31
docs: retire docs.nevermined.app references

eruizgar91:
Payments-repo slice of the cross-repo sweep tracked in [nevermined-io/nevermined.ai-website#233](https://github.com/nevermined-io/nevermined.ai-website/issues/233).

## What changed

This repo had four source references to `docs.nevermined.app`, and all four hit the same dead path — `/api-reference/versioning` 404s on the old host *and* on `nevermined.ai/docs`. So this is issue case #2 (dead link, not just a stale host): a plain host substitution would have produced a working URL to a 404.

Repointed all four at the live page, `https://nevermined.ai/docs/development-guide/api-versioning` (verified 200):

| File | Context |
| --- | --- |
| `src/common/api-version.ts:15` | `@see` on `LOCKED_API_VERSION` |
| `src/common/types.ts:53` | `@see` on `PaymentOptions.version` |
| `CLAUDE.md:92` | API Version Pinning section |
| `openclaw/skills/nevermined/SKILL.md:70` | already used the correct `development-guide/api-versioning` path — only the host was stale |

## Deliberately not touched

`docs/interfaces/PaymentOptions.html` and `docs/variables/LOCKED_API_VERSION.html` are TypeDoc build output, not source — the false-positive category the issue calls out. They pick up the new URL on the next `pnpm doc`.

## Verification

```
$ grep -rn "docs\.nevermined\.app" CLAUDE.md src openclaw
(no matches)
```

`pnpm build` currently fails on `src/x402/langchain/agent.ts` with `TS2307: Cannot find module '@langchain/langgraph/prebuilt'`. That failure reproduces on a clean tree at `main` and is unrelated to this diff, which touches only doc comments and markdown.

Refs nevermined-io/nevermined.ai-website#233

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,7 +89,7 @@ Every HTTP call to the Nevermined backend carries a `Nevermined-Version` header 
 
 **Override path**: per `Payments` instance via the existing `options.version` (e.g. `Payments.getInstance({ ..., version: '1.0' })`). There is intentionally **no per-request override** — `Nevermined-Version` is not in the `ALLOWED_EXTRA_HEADERS` allowlist, so `extraHeaders` cannot clobber it.
 
-See https://docs.nevermined.app/api-reference/versioning and nvm-monorepo#1535/#1938.
+See https://nevermined.ai/docs/development-guide/api-versioning and nvm-monorepo#1535/#1938.
 
 ### x402 Protocol and Documentation Standards
 
--- a/openclaw/skills/nevermined/SKILL.md
+++ b/openclaw/skills/nevermined/SKILL.md
@@ -67,7 +67,7 @@ Log out from Nevermined and remove the stored API key.
 
 Every Nevermined-backend call this plugin makes goes through the `@nevermined-io/payments` SDK, which **pins the backend API version** (the platform `MAJOR.MINOR`) via the `Nevermined-Version` header automatically — so OpenClaw-orchestrated agents keep getting a stable wire shape across platform releases without any per-call work. The pinned version is the one the bundled SDK release was built and tested against; it moves only when the plugin upgrades its SDK dependency.
 
-If an agent makes a **direct** REST call to the Nevermined API (outside these tools), send `Nevermined-Version: <MAJOR.MINOR>`, default it to the `current` from `GET /api/v1/meta/versions`, and never silently change a key's stored pin. See <https://docs.nevermined.app/docs/development-guide/api-versioning>.
+If an agent makes a **direct** REST call to the Nevermined API (outside these tools), send `Nevermined-Version: <MAJOR.MINOR>`, default it to the `current` from `GET /api/v1/meta/versions`, and never silently change a key's stored pin. See <https://nevermined.ai/docs/development-guide/api-versioning>.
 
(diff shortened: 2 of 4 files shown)


aaitor (approved):
Clean, correctly-scoped doc-URL sweep — approving. All four `docs.nevermined.app` → `nevermined.ai/docs/development-guide/api-versioning` swaps land on a live page, and every claim in the description checks out. This is issue #233's "dead link, not stale host" case handled correctly: a plain host substitution would have produced a 404, so repointing at the correct live path was the right call.

**Verification I ran**
- `curl -L` on the affected URLs:
  - `https://nevermined.ai/docs/development-guide/api-versioning` → **200** ✓ (the new target)
  - `https://nevermined.ai/docs/api-reference/versioning` (what a host-swap-only fix of the old `/api-reference/versioning` path yields) → **404** ✓ — confirms the lazy shortcut would've been dead, justifying the path change over a plain host substitution
  - `https://docs.nevermined.app/...` now redirects to `nevermined.ai/docs/...`; the dead path still 404s through the redirect
  - SKILL.md's old URL (`docs.nevermined.app/docs/development-guide/api-versioning`) resolved to **200** via that redirect — i.e. it was host-stale only, not a dead link. The description says exactly this; accurate.
- `git grep docs.nevermined.app` over **all tracked files on the PR branch** (12af5ac): only two matches remain, both TypeDoc build output (see LOW below). No missed source reference.
- Branch diff (`git diff --name-only origin/main…HEAD`) is exactly the 4 files the description lists.
- CI: 7/7 green.
- Leanness pass (`/ponytail-review`): 4 one-for-one line swaps, net 0 removable. Lean already. Ship.
- Panel scaled to the change: pure doc-comment/markdown URL swap — no logic, security, DB, test, or architecture surface. The substantive verification here is URL liveness + tracked-source completeness (done above); a full sub-agent panel would each return empty, so it was skipped deliberately rather than run as ceremony.

**LOW — two committed TypeDoc files still carry the dead link (deferred, acceptable)**
`docs/interfaces/PaymentOptions.html` and `docs/variables/LOCKED_API_VERSION.html` still contain `https://docs.nevermined.app/api-reference/versioning` — the exact dead `/api-reference/versioning` path — on the PR branch. The description correctly classifies these as regenerated TypeDoc output (issue #233's false-positive category) that picks up the fix on the next `pnpm doc`. The only wrinkle: they're *committed* build output, so anyone browsing the repo's `docs/` still hits the dead link until a regen lands. Not a blocker, not worth gating on. If you want to close it out in this PR, `pnpm doc` + commit the two files; otherwise it's a defensible stated deferral.

**Breadcrumb for the cross-repo sweep (#233) — not this PR**
Local, gitignored build output `dist/api/nvm-apid.d.ts` carries a *different* stale reference: `docs.nevermined.app/docs/tutorials/integration/nvm-api-keys`. Out of scope here (gitignored; and it's a tutorials URL, not the versioning one). But that `.d.ts` is generated from the upstream API spec, so the `nvm-api-keys` tutorial link may still point at `docs.nevermined.app` wherever it's authored upstream — worth a glance as part of #233.

PR #420, opened 2026-08-28
fix(organizations): onboardCustomer returns the Bearer (hash), not the encrypted blob

aaitor:
## Why this matters

White-label customer onboarding through the SDK is currently broken: `onboardCustomer` hands callers a credential that cannot authenticate. Any integrator following our docs — initialize a client with the returned key and act for the customer — gets `401`s, because the value we return is an encrypted internal blob, not a usable Bearer token. This fixes the SDK so the documented flow actually works, and unblocks the docs PR that surfaced it ([docs#301](https://github.com/nevermined-io/docs/pull/301)).

## The bug

`onboardCustomer` returned `walletResult.nvmApiKey` — which the backend documents (for the `as=customer` outcome, #2924) as *"the encrypted SDK/server-side blob — NOT a Bearer; do not send it as a header. To authenticate an HTTP request use `hash`."* The usable credential is `walletResult.hash` (`<env>:<JWT>`).

The sibling `createMember` already does the right thing:

```ts
return { ...data.walletResult, nvmApiKey: data.walletResult.hash }
```

`onboardCustomer` simply never got that mapping, so it leaked the raw blob through as `nvmApiKey`. Following it produced `Authorization: Bearer <encrypted blob>`.

## The fix

- Return `wallet.hash` as `nvmApiKey` (mirroring `createMember`).
- Guard on `wallet.hash` (the field we now return) instead of the blob, so a partial payload still fails loudly.

## Alternatives considered

- **Document the current behavior instead** (have docs tell REST callers to use `hash` and demote the SDK path). Rejected: the SDK genuinely can't produce a working customer client today — `CustomerOnboardingResponse` exposes no `hash` — so that path is a real defect, not a doc gap. Fixing the SDK makes the whole flow correct rather than papering over it.

## Test plan

- Corrected the unit test, which had encoded the inverted premise (it called `hash` a "lookup hash" and asserted the raw blob was returned). It now asserts `hash` is surfaced as the Bearer and the blob never leaks.
- `pnpm jest tests/unit/organizations-api.test.ts` → 20/20 pass.
- `pnpm build` (tsc) clean; `pnpm lint` 0 errors.

https://claude.ai/code/session_01CZ3y8ptZEYfscAKfhyfs1h

### DIFF
--- a/src/api/organizations-api/organizations-api.ts
+++ b/src/api/organizations-api/organizations-api.ts
@@ -91,17 +91,20 @@ export class OrganizationsAPI extends BasePaymentsAPI {
     if (response.status === 202 || wallet.consentRequired) {
       return { consentRequired: true }
     }
-    // New account or the org's returning customer: a usable key MUST be present.
-    // A 2xx without one means a partial/unexpected payload — fail loudly rather
-    // than hand back a "success" carrying undefined credentials.
-    if (!wallet.nvmApiKey) {
+    // New account or the org's returning customer: the Bearer MUST be present.
+    // The backend returns the usable credential as `walletResult.hash`
+    // (`<env>:<JWT>`) — the sibling `nvmApiKey` is the encrypted server-side
+    // blob and is NOT a valid Bearer. Surface `hash` as `nvmApiKey`, exactly as
+    // createMember does. A 2xx without a hash means a partial/unexpected payload
+    // — fail loudly rather than hand back a "success" carrying no credential.
+    if (!wallet.hash) {
       throw PaymentsError.internal(
         'Customer onboarding did not return an API key (unexpected backend response)',
       )
     }
     return {
       consentRequired: false,
-      nvmApiKey: wallet.nvmApiKey,
+      nvmApiKey: wallet.hash,
       userId: wallet.userId,
       userWallet: wallet.userWallet,
       isCustomer: wallet.isCustomer,
--- a/src/api/organizations-api/organizations-api.ts
+++ b/src/api/organizations-api/organizations-api.ts
@@ -109,6 +109,7 @@ export class OrganizationsAPI extends BasePaymentsAPI {
       userWallet: wallet.userWallet,
       isCustomer: wallet.isCustomer,
       customerRecorded: wallet.customerRecorded,
+      expiresAt: wallet.expiresAt,
     }
   }
 
--- a/tests/unit/organizations-api.test.ts
+++ b/tests/unit/organizations-api.test.ts
@@ -315,7 +315,7 @@ describe('OrganizationsAPI — workspace surface', () => {
   })
 
   describe('onboardCustomer (#2418)', () => {
-    test('new/returning customer: sends as=customer and returns the REAL nvmApiKey', async () => {
+    test('new/returning customer: sends as=customer and returns hash as the Bearer', async () => {
       const payments = makePayments()
       const calls = installFetchStub(() => ({
         ok: true,
@@ -324,10 +324,14 @@ describe('OrganizationsAPI — workspace surface', () => {
           success: true,
           message: 'Customer onboarded',
           walletResult: {
-            hash: 'lookup-hash',
+            // `hash` is the Bearer credential (`<env>:<JWT>`) — the backend
+            // documents this as the value to send as `Authorization: Bearer …`.
+            hash: 'sandbox:jwt-bearer-token',
             userId: 'us-123',
             userWallet: '0xabc',
-            nvmApiKey: 'nvm-real-usable-key',
+            // `nvmApiKey` is the encrypted server-side blob — explicitly NOT a
+            // Bearer; sending it as a header does not authenticate.
+            nvmApiKey: 'encrypted-blob-not-a-bearer',
             isCustomer: true,
             customerRecorded: true,
             alreadyMember: false,
@@ -344,10 +348,11 @@ describe('OrganizationsAPI — workspace surface', () => {
         email: 'customer@example.com',
         as: 'customer',
       })
-      // The USABLE key is returned — not the (non-usable) lookup hash.
+      // The USABLE Bearer is surfaced as `nvmApiKey` (mirroring createMember) —
+      // the raw `nvmApiKey` blob must never leak through as a credential.
       expect(result).toEqual({
         consentRequired: false,
-        nvmApiKey: 'nvm-real-usable-key',
+        nvmApiKey: 'sandbox:jwt-bearer-token',
         userId: 'us-123',
         userWallet: '0xabc',
         isCustomer: true,
@@ -390,7 +395,7 @@ describe('OrganizationsAPI — workspace surface', () => {
       installFetchStub(() => ({
         ok: true,
         status: 201,
-        // Partial/regressed payload: not consent-pending, yet no nvmApiKey.
+        // Partial/regressed payload: not consent-pending, yet no `hash` (the Bearer).
         body: { success: true, walletResult: { userId: 'us-1', isCustomer: true } },
       }))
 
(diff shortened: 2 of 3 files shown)


aaitor:
@eruizgar91 small one-liner following up on the `onboardCustomer` you reviewed in #413 — it was returning `walletResult.nvmApiKey` (the encrypted blob) instead of `walletResult.hash` (the Bearer), so the SDK customer client never authenticated. Fix mirrors `createMember`; test corrected. Mind a quick look? Unblocks docs#301.

r-marques:
:eyes: Reviewing

src/api/organizations-api/organizations-api.ts
```diff
@@ -91,17 +91,20 @@ export class OrganizationsAPI extends BasePaymentsAPI {
...
+    // — fail loudly rather than hand back a "success" carrying no credential.
+    if (!wallet.hash) {
       throw PaymentsError.internal(
         'Customer onboarding did not return an API key (unexpected backend response)',
       )
     }
     return {
       consentRequired: false,
-      nvmApiKey: wallet.nvmApiKey,
+      nvmApiKey: wallet.hash,
```
r-marques, line 107: **🟡 SHOULD FIX — Surface `expiresAt` too — the paired docs PR tells integrators to track it, and this return drops it**

The credential half is now right. The **expiry** is still dropped, and that matters because the docs change shipping alongside this one (nevermined-io/docs#301, `solutions/organizations/white-label-onboarding.mdx`) tells integrators:

> The exact expiry is returned as `expiresAt` (ISO-8601) — track it and re-onboard to refresh before it lapses.

A TypeScript SDK user cannot do that. The backend returns it on **this exact outcome** — `organizations.controller.ts:173-178` documents `expiresAt` as *"ISO-8601 expiry of `hash` (#2924). Returned for the `as=customer` credential outcome"* — but `onboardCustomer` neither reads `wallet.expiresAt` nor exposes it, and `CustomerOnboardingResponse` (`types.ts:46-58`) has no such field. So the only way to follow the documented guidance is to bypass the SDK and call the REST endpoint directly, which is precisely the situation the docs note frames as the *fallback*.

The change is small and additive:

```ts
// types.ts — in the consentRequired: false branch
/** ISO-8601 expiry of the credential (#2924). */
expiresAt?: string

// organizations-api.ts
return {
  consentRequired: false,
  nvmApiKey: wallet.hash,
  expiresAt: wallet.expiresAt,
  ...
}
```

Optional rather than required, since `expiresAt` is documented for the credential outcome specifically and an older backend would omit it — that keeps this from becoming a second `!wallet.hash`-style hard guard.

I'm raising this as should-fix rather than a blocker because the credential bug — the one that made `getInstance` send an encrypted blob as a Bearer — is genuinely fixed here, and shipping that is worth more than waiting. But the two PRs land together, and as it stands the docs promise something the SDK does not provide.

tests/unit/organizations-api.test.ts
```diff
@@ -344,10 +348,11 @@ describe('OrganizationsAPI — workspace surface', () => {
         email: 'customer@example.com',
         as: 'customer',
       })
-      // The USABLE key is returned — not the (non-usable) lookup hash.
+      // The USABLE Bearer is surfaced as `nvmApiKey` (mirroring createMember) —
+      // the raw `nvmApiKey` blob must never leak through as a credential.
       expect(result).toEqual({
         consentRequired: false,
-        nvmApiKey: 'nvm-real-usable-key',
+        nvmApiKey: 'sandbox:jwt-bearer-token',
```
r-marques, line 356: **✅ NICE — Fixed the test that asserted the defect, rather than leaving it green**

Worth calling out, because this is the failure mode that would have let the bug survive the fix. The previous test asserted `nvmApiKey: 'nvm-real-usable-key'` under the comment *"The USABLE key is returned — not the (non-usable) lookup hash"* — it encoded the **inverted** belief as a safety property, so it passed precisely because the code was wrong, and flipping the code alone would have turned it red and invited someone to "fix" the test back.

This PR corrects the fixture's story instead: `hash` becomes `'sandbox:jwt-bearer-token'` (the real `<env>:<JWT>` shape) and `nvmApiKey` becomes `'encrypted-blob-not-a-bearer'`, with comments stating which is which. That makes the assertion fail for the right reason if anyone reintroduces the bug, and the naming now carries the contract rather than contradicting it.

r-marques (commented):

Fixes the defect this repo's own `createMember` never had: `onboardCustomer` was returning `walletResult.nvmApiKey` — the encrypted server-side blob — as the credential, while `Payments.getInstance` sends its argument verbatim as `Authorization: Bearer …` (`src/api/base-payments.ts:222`). **The fix is correct, complete for the credential path, and mirrors `createMember` exactly.** The one thing left open is `expiresAt`, which the backend returns on this same outcome and which the paired docs PR instructs integrators to track — detail inline.

**What I reviewed**
- Diff range `origin/main...HEAD` — 2 files, +19/−11, head `214c250f`.
- Panel: none — this is my own verification against `nvm-monorepo@fd5af9a76` (the backend contract) and the SDK source. No panel agent contributed.
- **Verified the direction of the fix against the backend, not just the commit message.** `organizations.controller.ts:155-176` documents `hash` as "The Bearer credential … Format `<env>:<JWT>`" and `nvmApiKey` as "Encrypted SDK/server-side blob — NOT a Bearer; do not send it as a header." `organizations.service.ts:556-559` returns them as two **distinct** values, so they were never interchangeable.
- **Traced the guard flip.** `!wallet.nvmApiKey` → `!wallet.hash` changes behaviour in both directions and both are right: a payload with `hash` but no blob now succeeds (correct — `hash` is the credential), and one with the blob but no `hash` now throws instead of handing back a useless "success" (correct — that is the bug).
- Checked every other reader of these fields in the repo: only `createMember` (line 55, already correct) and this function. The CLI (`cli/src/commands/organizations/onboard-customer.ts:27`) consumes the result by the unchanged field name, so it picks the fix up for free.
- `CustomerOnboardingResponse` (`types.ts:46-58`) is unchanged, and its comment *"The usable NVM API key (Bearer credential). Always present in this branch."* is now **true** — it was false before this PR.
- **I did not run the tests**: the review worktree has no `node_modules` and I did not install into it. The test changes are assessed by reading.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 1 should-fix, 0 nits.

### 🧭 Inline comments (2)

- 🟡 **SHOULD FIX** — Surface `expiresAt` too — the paired docs PR tells integrators to track it, and this return drops it — `src/api/organizations-api/organizations-api.ts:107`
- ✅ **NICE** — Fixed the test that asserted the defect, rather than leaving it green — `tests/unit/organizations-api.test.ts:355`

r-marques (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 0 nit(s) left inline. Auto-merge was disabled so these can be looked at first.

Inline review: https://github.com/nevermined-io/payments/pull/420#pullrequestreview-5049740419

src/api/organizations-api/organizations-api.ts
aaitor: Done in 87bb714 — added optional `expiresAt` to `CustomerOnboardingResponse` and return `wallet.expiresAt` (with a test asserting it). SDK users can now follow the docs guidance without dropping to the REST endpoint. Kept it optional (not a hard guard) so an older backend that omits it still works.

tests/unit/organizations-api.test.ts
aaitor: Thanks — yes, that inverted test was the whole reason the bug survived the original feature PR. Corrected the fixture's story so it now fails for the right reason if anyone reintroduces it.

PR #423, opened 2026-09-01
fix(mcp): build a fresh transport per request so spec-compliant MCP clients work

### LINKED ISSUE #422: Managed MCP server returns empty 500 after initialize when the client echoes Mcp-Session-Id (breaks spec-compliant MCP clients)
## Summary

The managed MCP server (`payments.mcp.start()`) returns an **empty-body HTTP 500** to every request after `initialize` when the client echoes the `Mcp-Session-Id` header. Since every spec-compliant MCP client (Claude, Cursor, the official `@modelcontextprotocol/sdk` client) echoes that header per the streamable-HTTP spec, `client.connect()` fails and the server is unusable from real MCP clients. Requests sent **without** the header succeed, which is why simple scripts and tutorials don't hit it.

Reported via a user gist: https://gist.github.com/mqklin/8e2d10156f94bee42f6b226d7beb099c

## Environment

- `@nevermined-io/payments` **1.11.1** (current latest)
- `@modelcontextprotocol/sdk` **>= 1.25.0** (reproduced on 1.29.0 and 1.30.0)

## Reproduction

1. `POST /mcp` `initialize` → **200**, response carries `Mcp-Session-Id: <uuid>`
2. `POST /mcp` `tools/list` with `mcp-session-id: <uuid>` → **500, empty body**
3. Same `tools/list` **without** the header → **200**

Reduced to the SDK alone (no Nevermined code): one `StreamableHTTPServerTransport({ sessionIdGenerator: undefined, enableJsonResponse: true })` connected once and reused for a second request → request 1 (`initialize`) 200, request 2 (`tools/list`) empty 500. The transport writes the 500 itself, so the wrapping handler's catch never produces an error body.

## Root cause

`src/mcp/http/session-manager.ts` builds transports in stateless mode (`sessionIdGenerator: undefined`) but **caches them per session id and reuses them** — `getOrCreateSession()` returns the cached instance on a cache hit. The SDK's stateless transport is single-request by design, so reuse yields the empty 500. Meanwhile `src/mcp/http/mcp-handler.ts` sets the `Mcp-Session-Id` response header on every response, so every compliant client echoes it back and is routed straight into the broken reuse path.

## Expected behavior

A spec-compliant client that echoes `Mcp-Session-Id` on subsequent requests should get valid responses (`tools/list`, tool calls, etc.), not a 500.

## Impact

Blocks any out-of-the-box Claude / Cursor / SDK-client connection to a `payments.mcp.start()` server. The only client-side workaround is to strip the `mcp-session-id` header (non-compliant direct JSON-RPC calls).

aaitor:
## Why this matters

A user reported that AI agents built on our managed MCP server can't be reached by the tools people actually use — Claude, Cursor, and the official MCP SDK client all fail to connect out of the box, getting a blank HTTP 500 error right after connecting. This makes `payments.mcp.start()` servers effectively unusable for real customers, even though quick test scripts appeared to work. This PR fixes that so any spec-compliant MCP client connects and calls tools normally.

## Description

The managed MCP server handed the client an `Mcp-Session-Id` on `initialize`, then failed with an empty-body **HTTP 500** on every subsequent request that carried that header back. Per the MCP streamable-HTTP spec, every compliant client echoes the session id, so `client.connect()` failed at the `notifications/initialized`/`tools/list` step. Requests **without** the header succeeded, which is why simple scripts didn't surface the bug.

**Root cause:** `SessionManager.getOrCreateSession()` cached the stateless `StreamableHTTPServerTransport` (`sessionIdGenerator: undefined`) per session id and reused it. The SDK's stateless transport is single-request by design, so reuse writes an empty 500 on the second request. Because `mcp-handler.ts` sets `Mcp-Session-Id` on every response, compliant clients were steered straight into that broken reuse path.

**Fix:** `getOrCreateSession()` now always builds a **fresh** transport instead of returning the cached one — the echoed-session-id path behaves exactly like the no-header path that already worked. The existing `mcpServer.close()` before each connect already reaps the previous transport (via `onclose`), so the session map stays bounded. The cached entry is kept only so GET (SSE) / DELETE can resolve the latest transport for a session id.

**Alternatives considered:**
- *Stop emitting `Mcp-Session-Id`.* Makes compliant clients stop echoing, but is fragile (a proxy or reconnect can still send the header) and doesn't fix the root cause — reuse would still 500.
- *Implement real stateful sessions* (`sessionIdGenerator` set, multi-request transports). Requires a separate `McpServer` per session; the current architecture shares one server and closes/reconnects it per request. Much larger change for no additional user-visible benefit given the stateless design.
- *Chosen:* rebuild a fresh transport per request — smallest diff, fixes the root cause, keeps both the compliant and the direct-JSON-RPC paths working.

Fixes #422

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] Tests Cover Changes
- [ ] Documentation

**Test plan:**
- New `tests/integration/mcp/mcp_session_reuse_regression.test.ts` drives `initialize → tools/list` through the real handler + session manager over HTTP, asserting **200** both when the client echoes `Mcp-Session-Id` (the regression) and when it omits it. Verified RED before the fix (echoed id → 500), GREEN after.
- `pnpm build && pnpm lint` clean (only pre-existing warnings in untouched files).
- `pnpm test:unit` (570 passed) and `pnpm test:integration` (54 passed) green.

### DIFF
--- a/src/mcp/http/session-manager.ts
+++ b/src/mcp/http/session-manager.ts
@@ -107,15 +107,19 @@ export class SessionManager {
 
   /**
    * Get or create a transport for a session.
+   *
+   * Always builds a FRESH transport, even when a transport is already cached for
+   * this session id. The transports are created in stateless mode
+   * (`sessionIdGenerator: undefined`), and the SDK's stateless transport is
+   * single-request by design — reusing a cached one yields an empty-body HTTP 500
+   * on the second request. Spec-compliant clients (Claude, Cursor, the official
+   * SDK) echo the `Mcp-Session-Id` we return on `initialize` back on every
+   * subsequent request, which used to route them straight into that broken reuse
+   * path. Rebuilding per request keeps the echoed-id path working exactly like the
+   * no-header path. The cached entry is only retained so GET (SSE) / DELETE can
+   * still resolve the latest transport for a session id.
    */
   async getOrCreateSession(sessionId: string): Promise<any> {
-    if (this.sessions.has(sessionId)) {
-      const existingSession = this.sessions.get(sessionId)
-      if (existingSession) {
-        return existingSession
-      }
-    }
-
     if (!this.mcpServer) {
       throw new Error('MCP server not set. Call setMcpServer() first.')
     }
--- a/src/mcp/http/session-manager.ts
+++ b/src/mcp/http/session-manager.ts
@@ -116,8 +116,17 @@ export class SessionManager {
    * SDK) echo the `Mcp-Session-Id` we return on `initialize` back on every
    * subsequent request, which used to route them straight into that broken reuse
    * path. Rebuilding per request keeps the echoed-id path working exactly like the
-   * no-header path. The cached entry is only retained so GET (SSE) / DELETE can
-   * still resolve the latest transport for a session id.
+   * no-header path.
+   *
+   * The session map still records the latest transport per id so DELETE can
+   * terminate it and requests for an unknown id can be rejected. This managed
+   * server runs in stateless JSON-response mode (`enableJsonResponse: true`) and
+   * does NOT support standalone GET/SSE streams — the stateless transport rejects
+   * them — so there is no long-lived stream for a later POST to disrupt.
+   *
+   * ponytail: one shared McpServer that every request `close()`s and reconnects —
+   * two in-flight requests still race (tracked in #424). Per-session server+transport
+   * pairs are the upgrade if concurrency ever matters.
    */
   async getOrCreateSession(sessionId: string): Promise<any> {
     if (!this.mcpServer) {
--- a/tests/integration/mcp/mcp_session_reuse_regression.test.ts
+++ b/tests/integration/mcp/mcp_session_reuse_regression.test.ts
@@ -1,0 +1,89 @@
+/**
+ * Regression test for the "empty 500 after initialize when the client echoes
+ * Mcp-Session-Id" bug (gist mqklin/8e2d10156f94bee42f6b226d7beb099c).
+ *
+ * A spec-compliant MCP client echoes the `Mcp-Session-Id` returned on
+ * `initialize` back on every subsequent request. The managed server used to
+ * return a cached single-use stateless transport for that echoed id, which the
+ * SDK answers with an empty-body HTTP 500. This test drives the real
+ * mcp-handler + session-manager over HTTP and asserts the echoed-id path works.
+ */
+import express from 'express'
+import type { Server } from 'http'
+import type { AddressInfo } from 'net'
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
+import { SessionManager } from '../../../src/mcp/http/session-manager.js'
+import { mountMcpHandlers } from '../../../src/mcp/http/mcp-handler.js'
+
+describe('MCP Session - transport reuse regression', () => {
+  let server: Server
+  let mcpServer: McpServer
+  let baseUrl: string
+
+  beforeAll(async () => {
+    mcpServer = new McpServer({ name: 'repro', version: '1.0.0' })
+    mcpServer.registerTool('ping', { description: 'ping' }, async () => ({
+      content: [{ type: 'text', text: 'pong' }],
+    }))
+
+    const sessionManager = new SessionManager()
+    sessionManager.setMcpServer(mcpServer)
+
+    const app = express()
+    app.use(express.json())
+    mountMcpHandlers(app as any, { sessionManager, requireAuth: false })
+
+    await new Promise<void>((resolve) => {
+      server = app.listen(0, () => resolve())
+    })
+    baseUrl = `http://127.0.0.1:${(server.address() as AddressInfo).port}/mcp`
+  })
+
+  afterAll(async () => {
+    await mcpServer?.close().catch(() => {})
+    await new Promise<void>((resolve) => server?.close(() => resolve()))
+  })
+
+  const post = (body: object, sessionId?: string) =>
+    fetch(baseUrl, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json, text/event-stream',
+        ...(sessionId ? { 'mcp-session-id': sessionId } : {}),
+      },
+      body: JSON.stringify(body),
+    })
+
+  const initialize = {
+    jsonrpc: '2.0',
+    id: 1,
+    method: 'initialize',
+    params: {
+      protocolVersion: '2025-06-18',
+      capabilities: {},
+      clientInfo: { name: 'r', version: '0' },
+    },
+  }
+  const toolsList = { jsonrpc: '2.0', id: 2, method: 'tools/list', params: {} }
+
+  test('tools/list succeeds when the client echoes the Mcp-Session-Id (compliant client)', async () => {
+    const initRes = await post(initialize)
+    expect(initRes.status).toBe(200)
+    const sessionId = initRes.headers.get('mcp-session-id')
+    expect(sessionId).toBeTruthy()
+
+    // Compliant clients echo the session id back — this used to yield an empty 500.
+    const listRes = await post(toolsList, sessionId!)
+    expect(listRes.status).toBe(200)
+    const body = await listRes.json()
+    expect(body.result.tools.map((t: any) => t.name)).toContain('ping')
+  })
+
+  test('tools/list still succeeds without the session header (non-compliant path)', async () => {
+    const listRes = await post(toolsList)
+    expect(listRes.status).toBe(200)
+    const body = await listRes.json()
+    expect(body.result.tools.map((t: any) => t.name)).toContain('ping')
+  })
+})
--- a/tests/integration/mcp/mcp_session_reuse_regression.test.ts
+++ b/tests/integration/mcp/mcp_session_reuse_regression.test.ts
@@ -86,4 +86,25 @@ describe('MCP Session - transport reuse regression', () => {
     const body = await listRes.json()
     expect(body.result.tools.map((t: any) => t.name)).toContain('ping')
   })
+
+  test('standalone GET/SSE is unsupported in stateless mode, so a later POST tears down nothing', async () => {
+    // Review question on #423: does rebuilding a transport on POST close a live
+    // GET/SSE stream on the same session? In this stateless config
+    // (sessionIdGenerator: undefined, enableJsonResponse) the SDK rejects the
+    // standalone SSE stream outright — it never establishes, so there is no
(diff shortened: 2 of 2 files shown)


r-marques:
:eyes: Reviewing

r-marques (commented):

The bug is real and the fix is the right shape: a stateless `StreamableHTTPServerTransport` is
single-request, so returning the cached one for an echoed `Mcp-Session-Id` produced the empty 500, and
rebuilding per request makes the echoed-id path behave exactly like the no-header path. The regression
test is the good kind — it drives real HTTP through the real `mountMcpHandlers` + `SessionManager`
rather than asserting on a double.

My two comments are about what removing the early return does to the *other* paths. That `return`
was also the only route through `getOrCreateSession` that skipped `await this.mcpServer.close()`
(`session-manager.ts:133`), so that close now runs on **every** request — and it closes whatever
transport is currently connected, including one another request is using. Concretely, a POST now tears
down the transport a GET/SSE stream is holding, which is the opposite of what the new comment at
`:119-120` says the cache is retained for.

**What I reviewed**
- Diff range `origin/main...HEAD` — 2 files, +100/−7, head `3cb7ea51`.
- Panel: **none** — the `pr-review-toolkit` agents have returned findings 0 times in 26 spawns this
  session, so rather than cite a panel that reports nothing, this is my own reading. Stating the
  deviation rather than implying coverage I did not have.
- Traced the full request path: `mcp-handler.ts:101-130` (session id derivation → `setRequestContext`
  → `getOrCreateSession` → `transport.handleRequest`), and the GET/DELETE handlers at `:157-192`.
- **Checked a suspected auth regression and it is not one.** `onclose` deletes
  `requestContexts[sessionId]` (`session-manager.ts:152`), and the handler sets that context one line
  *before* calling `getOrCreateSession` — so on a repeat request the new close path deletes the entry
  just written. It does not matter: `getRequestContext` has **no callers** anywhere in `src/`, and
  headers actually reach tool handlers through `requestContextStorage.run(...)` (`mcp-handler.ts:129`),
  which this change does not touch. Worth recording since the deletion looks alarming in isolation.
- **Checked for a transport/map leak and found none.** Each new connect is preceded by the close of the
  previously connected transport, whose `onclose` removes its own entry, so `sessions` stays at ~1 live
  entry rather than growing per request.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 2 should-fix, 1 nit.

### 🧭 Inline comments (3)

- 🟡 **SHOULD FIX** — A POST now closes the transport the GET/SSE stream is using — the opposite of what this sentence claims — `src/mcp/http/session-manager.ts:120`
- 🟡 **SHOULD FIX** — `mcpServer.close()` now runs on every request, so two concurrent requests tear down each other's transport — `src/mcp/http/session-manager.ts:122`
- 💡 **NIT** — The test pins the fix but cannot catch either consequence of removing the early return — `tests/integration/mcp/mcp_session_reuse_regression.test.ts:76`

src/mcp/http/session-manager.ts
```diff
@@ -107,15 +107,19 @@ export class SessionManager {
...
+   * Always builds a FRESH transport, even when a transport is already cached for
+   * this session id. The transports are created in stateless mode
+   * (`sessionIdGenerator: undefined`), and the SDK's stateless transport is
+   * single-request by design — reusing a cached one yields an empty-body HTTP 500
+   * on the second request. Spec-compliant clients (Claude, Cursor, the official
+   * SDK) echo the `Mcp-Session-Id` we return on `initialize` back on every
+   * subsequent request, which used to route them straight into that broken reuse
+   * path. Rebuilding per request keeps the echoed-id path working exactly like the
+   * no-header path. The cached entry is only retained so GET (SSE) / DELETE can
+   * still resolve the latest transport for a session id.
```
r-marques: **🟡 SHOULD FIX — A POST now closes the transport the GET/SSE stream is using — the opposite of what this sentence claims**

This says the cached entry is kept "so GET (SSE) / DELETE can still resolve the latest transport for a session id". Resolving it still works; **using** it does not, because the removed early return was the only path that skipped `await this.mcpServer.close()` at `:133`.

Concrete sequence, all within one compliant client and one session id `S`:

1. `POST` initialize → `getOrCreateSession(S)` connects `t1`, `sessions[S] = t1`.
2. `GET /mcp` → `mcp-handler.ts:159-165` resolves `t1` and calls `t1.handleRequest(req, res)` — the SSE stream is now live on `t1`.
3. `POST` tools/list with the echoed `S` → **`this.mcpServer.close()` closes `t1`**, whose `onclose` (`:150-155`) deletes `sessions[S]`; `t2` is built and stored.

The stream opened in step 2 is now closed underneath the client. Before this PR step 3 hit the cache and returned `t1` without closing anything, so the stream survived.

I am filing this as should-fix rather than a blocker deliberately: I verified the mechanism, but I did **not** verify that Claude/Cursor actually open the GET stream against this server — with `enableJsonResponse: true` the responses themselves are JSON, so the GET leg may be unused in practice. If it is unused, the honest fix is to say so here instead, because the sentence as written asserts a property the code no longer has. If it *is* used, the transport backing a stream needs to be exempt from the teardown.

Either way this comment should not survive merge in its current form — it is the sort of statement a future reader will rely on.

<sub>🤖 pr-review-toolkit · manual · high</sub>

src/mcp/http/session-manager.ts
```diff
@@ -107,15 +107,19 @@ export class SessionManager {
...
    */
   async getOrCreateSession(sessionId: string): Promise<any> {
```
r-marques, line 131: **🟡 SHOULD FIX — `mcpServer.close()` now runs on every request, so two concurrent requests tear down each other's transport**

With the cache hit gone, every call reaches `await this.mcpServer.close()` at `:133`. `mcpServer` is a single shared instance, and that close terminates whichever transport is currently connected — not "this request's" transport, which does not exist yet.

So for two overlapping requests A and B:

```
A: close() → connect(tA) → sessions[S]=tA → tA.handleRequest(...)  ← in flight
B:                          close()  ← closes tA underneath A
   → connect(tB) → tB.handleRequest(...)
```

A's response is being written by a transport that has just been closed.

**This is pre-existing, not introduced here** — the no-header path already generates a fresh id per request (`mcp-handler.ts:104-107`), so it always missed the cache and always closed. What this PR changes is that the *established-session* path, the one case that previously avoided the teardown, now closes too. So the window goes from "most requests" to "all requests", and it does so on precisely the traffic this PR is enabling: a compliant client that keeps one session and sends many calls, sometimes concurrently.

I would not hold the PR for this — the 500 it fixes is worse than the race it widens — but it is worth an issue, because the current design cannot serve two in-flight MCP requests at once by construction. The shape that fixes it is one `McpServer`+transport pair per in-flight request (or per session) rather than one global server that every request re-connects.

<sub>🤖 pr-review-toolkit · manual · high</sub>

tests/integration/mcp/mcp_session_reuse_regression.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+  const toolsList = { jsonrpc: '2.0', id: 2, method: 'tools/list', params: {} }
+
+  test('tools/list succeeds when the client echoes the Mcp-Session-Id (compliant client)', async () => {
+    const initRes = await post(initialize)
+    expect(initRes.status).toBe(200)
+    const sessionId = initRes.headers.get('mcp-session-id')
+    expect(sessionId).toBeTruthy()
+
+    // Compliant clients echo the session id back — this used to yield an empty 500.
```
r-marques, line 76: **💡 NIT — The test pins the fix but cannot catch either consequence of removing the early return**

Both cases here are sequential, single-client, `requireAuth: false`, and never open the GET stream — so the two effects of dropping the cache hit are invisible to it: the SSE teardown (`session-manager.ts:119`) and the concurrent-request race (`:122`).

A third case would pin the property the new comment actually claims, and is only a few lines given the harness you have already built:

```ts
test('an open SSE stream survives a subsequent POST on the same session', async () => {
  const initRes = await post(initialize)
  const sessionId = initRes.headers.get('mcp-session-id')!
  const sse = await fetch(baseUrl, {
    method: 'GET',
    headers: { Accept: 'text/event-stream', 'mcp-session-id': sessionId },
  })
  expect(sse.status).toBe(200)
  await post(toolsList, sessionId)          // currently closes the stream's transport
  // assert the stream is still readable here
})
```

If that test is red, it is the finding on `:119` reproduced; if it is green, my reading of the teardown is wrong and I would rather it be settled by a test than by either of us arguing from the code.

The rest of this file is well judged, incidentally — driving the real `mountMcpHandlers` over a real socket is what makes it a regression test rather than a restatement of the implementation.

<sub>🤖 pr-review-toolkit · manual · medium</sub>

r-marques (approved):

Inline review: https://github.com/nevermined-io/payments/pull/423#pullrequestreview-5078453426

src/mcp/http/session-manager.ts
aaitor: You're right that the comment over-claimed. I probed it against the real `mountMcpHandlers` over a socket: in this stateless config (`sessionIdGenerator: undefined`, `enableJsonResponse: true`) the standalone GET/SSE stream **never establishes** — GET returns 500 (`content-type: text/plain`) and the stream is DONE immediately, independent of this PR. So there is no live stream for a subsequent POST to close; the teardown mechanism you traced is real but never reached in practice here.

Since it's unused, I took the branch you offered: fixed the comment to state the truth instead of asserting a property the code no longer has. The session map is now described as retained for DELETE termination + unknown-id rejection, with an explicit note that this server does not support standalone GET/SSE. Done in 35ed568.

tests/integration/mcp/mcp_session_reuse_regression.test.ts
aaitor: Settled by test in 35ed568, and the answer changes the test's shape: your suggested case would be RED, but not from the teardown — the GET/SSE leg returns 500 and never opens a stream at all in this stateless mode (verified over a real socket). So I added a case asserting GET/SSE does **not** establish (`status != 200`) and that the session id stays usable for POSTs afterwards. That pins the actual behavior your question was probing rather than locking in a teardown that never happens. Thanks for pushing for the test over arguing from the code — the probe is what surfaced that GET is 500 here.

src/mcp/http/session-manager.ts
aaitor: Agreed on all counts — real, pre-existing, and this PR widens the window from 'most requests' to 'all requests'. Filed as #424 with your traced sequence and the one-server-per-in-flight-request/session fix as the upgrade path, and left a `ponytail:` marker at the `close()` site pointing to it. Keeping it out of this PR since the 500 it fixes is worse than the race it widens, exactly as you said.

PR #429, opened 2026-09-06
feat(orders): OrdersAPI client for browser-fiat Orders (createOrder / getOrder)

aaitor:
## Why this matters

Merchants can now charge a customer an arbitrary amount by card straight from the Nevermined SDK, without first publishing a payment plan or asking the buyer to create a Nevermined account. One call creates the order and returns the Stripe client secret the checkout page needs; a second call reads the order's status so the merchant can fulfil once it is paid. This delivers the TypeScript half of nvm-monorepo#3250 (P1.8) for the "dynamic pricing: goods & services" epic (#3238).

### What

`payments.orders` (new `OrdersAPI`) wrapping the Orders endpoints shipped in nvm-monorepo#3311:

- `createOrder(options)` → `POST /api/v1/orders` (merchant, authenticated). Returns `{ orderId, status, clientSecret? }`.
- `getOrder(orderId)` → `GET /api/v1/orders/:id` (anonymous, public HTTP options — the unguessable id is the access control). Returns the buyer-safe `Order` projection.

Modelled on `PlansAPI`: same `BasePaymentsAPI` request layer (`getBackendHTTPOptions` / `getPublicHTTPOptions`), same `Bearer` + `Nevermined-Version` headers, errors surfaced via `PaymentsError.fromBackend(...)` so `err.code` is the catalogue code (`BCK.ORDER.0001..0011`). No new error types, no hand-rolled HTTP, no server changes.

### Decisions (left to the implementer by #3250)

- **Typed options object, not positionals.** `createOrder` has 1 required + 7 optional fields; positionals would be unusable (the same pain `registerPlan`'s six positionals already cause). `OrganizationsAPI.getOrganizationActivity(orgId, filters)` already takes an options object, so this is in-convention. `getOrder(orderId)` stays positional like `getPlan(planId)`.
- **`currency` defaults to `'usd'`.** The wire requires it but Phase 1 accepts only `usd`; defaulting removes boilerplate without hiding anything. Typed as the literal `'usd'` so a later currency is an additive union change.
- **Full response typing.** `CreateOrderResult` and `Order` mirror `CreateOrderResponseDto` / `OrderResponseDto` field-for-field, preserving optional-vs-nullable (`clientSecret?` is absent when withheld; `description` / `buyerRef` / `paymentIntentId` / `expiresAt` are `string | null`). No merchant-identity or fee fields are modelled, per the buyer-safe contract. `OrderStatus` is the six-value wire union.
- **No client-side range validation.** The server owns the bounds (`BCK.ORDER.0001`; the max is env-configurable via `ORDER_MAX_AMOUNT_MINOR`), so duplicating them in the SDK would only drift.
- **Tests: unit, fetch-stub style** (same pattern as `organizations-api.test.ts`): URL, method, headers (Bearer present on create, absent on get, `X-Current-Org-Id` forwarded from the instance pin), the exact JSON body including verbatim opaque `metadata` / `lineItems`, and catalogue-code propagation for `BCK.ORDER.0003` / `0002`. No e2e: creating an order on staging needs an org-scoped key with an active org and a Connect account, which the CI test keys are not, and staging may not carry #3311 yet (argocd#611).

### Escalations (per the #3250 brief) — @aaitor

1. **No release / publish.** This PR stops at merge; no version bump. Cutting `@nevermined-io/payments` with this needs your explicit go-ahead.
2. **Org-scoped API key.** The API derives the merchant organization from the key's own org tag (`req.apiKeyOrgId`, set by `NvmApiHashKeyGuard`), not from the `X-Current-Org-Id` header. The SDK's key model is "any NVM API key string", so an org-scoped key is expressible today with no change: the merchant constructs `Payments` with it. What does **not** work is `setOrganizationId()` on a personal key — that only sets the header, which the org gate (#3317) ignores, so the call is refused with `BCK.ORDER.0003`. Documented in the class TSDoc; no workaround attempted. Flagging so "get an org-scoped key" lands in the merchant onboarding docs.

Related to nevermined-io/nvm-monorepo#3250 (P1.8 — SDK OrdersAPI), epic nevermined-io/nvm-monorepo#3238.

## Types of changes

## Checklist:

- [x] Tests Cover Changes
- [x] Documentation — `markdown/orders.md` guide page + `markdown/README.md` index entry; TSDoc on `OrdersAPI` and its types

### Test plan

- [x] `pnpm build` — exit 0
- [x] `pnpm lint` — 0 errors (3 pre-existing warnings in untouched files)
- [x] `pnpm test:unit` — 44 suites, 576 passed
- [x] Deep pre-PR review — 0 blockers; 2 should-fix addressed in the second commit, re-run clean (0/0/0)
- [ ] CI green: `lint_build`, `cli_sync_check`, `openclaw_check`, `unit_integration`, `e2e`
- [x] `./scripts/generate-docs.sh` — exit 0

Refs nevermined-io/nvm-monorepo#3250 · epic #3238 · server contract nvm-monorepo#3311 · org gate #3317. Companion PR: nevermined-io/payments-py#270 (Python SDK).

https://claude.ai/code/session_01CSHvoA2jHxfARTVwEmrR33

### DIFF
--- a/markdown/orders.md
+++ b/markdown/orders.md
@@ -30,11 +30,11 @@ const payments = Payments.getInstance({
 })
 
 const { orderId, status, clientSecret } = await payments.orders.createOrder({
-  amountMinor: 3437, // USD cents: $34.37 ($1.00 - $999,999.99)
+  amountMinor: 3437, // USD cents: $34.37
   description: 'Cart checkout - 3 items',
   buyerRef: 'merchant-order-4821',
   idempotencyKey: 'merchant-order-4821',
-  lineItems: [{ sku: 'PRO-PLAN', quantity: 1, amountMinor: 3437 }],
+  lineItems: [{ sku: 'PRO-PLAN', quantity: 1, unit_price: 3437 }],
   metadata: { channel: 'web' },
 })
 
@@ -45,13 +45,13 @@ console.log(clientSecret) // hand this to the browser (Stripe.js confirm)
 
 | Option | Type | Description |
 |---|---|---|
-| `amountMinor` | `number` | Charge amount in USD cents, `100` to `99_999_999`. |
+| `amountMinor` | `number` | Charge amount in USD cents (at least `100`, i.e. $1.00). The API validates the upper bound (`BCK.ORDER.0001`); a deployment may enforce a lower per-order cap (`BCK.ORDER.0003`). |
 | `currency` | `'usd'` | ISO currency, lower-cased. Default and only value in Phase 1. |
 | `description` | `string` | Optional. Human-readable description (max 1024 chars). |
 | `buyerRef` | `string` | Optional. Your own reference for the buyer or cart (max 255 chars). |
 | `idempotencyKey` | `string` | Optional. A retried create with the same key returns the same Order and `clientSecret`; a conflicting body is refused with `BCK.ORDER.0007`. |
-| `lineItems` | `Record<string, unknown>[]` | Optional. Recorded verbatim, opaque to the API. |
-| `metadata` | `Record<string, unknown>` | Optional. Recorded verbatim, opaque to the API. |
+| `lineItems` | `Record<string, unknown>[]` | Optional. Merchant-defined structure, recorded verbatim (keys are not transformed), opaque to the API. |
+| `metadata` | `Record<string, unknown>` | Optional. Merchant-defined structure, recorded verbatim (keys are not transformed), opaque to the API. |
 | `captureMode` | `'automatic'` | Optional. The only value in Phase 1. |
 | `paymentProvider` | `'stripe'` | Optional. The only value in Phase 1. |
 
--- a/markdown/orders.md
+++ b/markdown/orders.md
@@ -1,0 +1,110 @@
+---
+title: "Orders"
+description: "Charge an arbitrary amount by card without a payment plan: create a browser-fiat Order and read its status"
+icon: "cart-shopping"
+---
+
+# Orders
+
+This guide covers browser-fiat **Orders**: a merchant-initiated, off-plan charge for an arbitrary amount (the Stripe PaymentIntent analog) that the buyer's browser confirms client-side. No payment plan, no buyer Nevermined account, no delegation.
+
+## Overview
+
+1. `payments.orders.createOrder(...)` — the merchant creates the Order server-side and receives a Stripe `clientSecret`.
+2. The merchant hands the `clientSecret` to its checkout page, which confirms the payment with Stripe.js.
+3. `payments.orders.getOrder(orderId)` — anyone holding the unguessable `orderId` reads the buyer-safe status. No API key is sent on the wire.
+
+## Requirements
+
+`createOrder` needs an **organization-scoped** NVM API key: the API resolves the merchant organization from the key's own org tag. A personal key is refused with `BCK.ORDER.0003`, and `payments.setOrganizationId(...)` does not substitute for an org-scoped key. The organization must be active and have a Stripe Connect account able to receive card payments.
+
+## Create an Order
+
+```typescript
+import { Payments, EnvironmentName } from '@nevermined-io/payments'
+
+// Initialize with the merchant's organization-scoped API key
+const payments = Payments.getInstance({
+  nvmApiKey: process.env.NVM_API_KEY!,
+  environment: 'sandbox' as EnvironmentName,
+})
+
+const { orderId, status, clientSecret } = await payments.orders.createOrder({
+  amountMinor: 3437, // USD cents: $34.37 ($1.00 - $999,999.99)
+  description: 'Cart checkout - 3 items',
+  buyerRef: 'merchant-order-4821',
+  idempotencyKey: 'merchant-order-4821',
+  lineItems: [{ sku: 'PRO-PLAN', quantity: 1, amountMinor: 3437 }],
+  metadata: { channel: 'web' },
+})
+
+console.log(orderId) // ord_... - the buyer-facing access control
+console.log(status) // 'requires_payment'
+console.log(clientSecret) // hand this to the browser (Stripe.js confirm)
+```
+
+| Option | Type | Description |
+|---|---|---|
+| `amountMinor` | `number` | Charge amount in USD cents, `100` to `99_999_999`. |
+| `currency` | `'usd'` | ISO currency, lower-cased. Default and only value in Phase 1. |
+| `description` | `string` | Optional. Human-readable description (max 1024 chars). |
+| `buyerRef` | `string` | Optional. Your own reference for the buyer or cart (max 255 chars). |
+| `idempotencyKey` | `string` | Optional. A retried create with the same key returns the same Order and `clientSecret`; a conflicting body is refused with `BCK.ORDER.0007`. |
+| `lineItems` | `Record<string, unknown>[]` | Optional. Recorded verbatim, opaque to the API. |
+| `metadata` | `Record<string, unknown>` | Optional. Recorded verbatim, opaque to the API. |
+| `captureMode` | `'automatic'` | Optional. The only value in Phase 1. |
+| `paymentProvider` | `'stripe'` | Optional. The only value in Phase 1. |
+
+## Read an Order
+
+```typescript
+const order = await payments.orders.getOrder(orderId)
+
+console.log(order.status) // 'requires_payment' | 'paid' | 'refunded' | 'partially_refunded' | 'disputed' | 'failed'
+console.log(order.amountMinor) // 3437
+console.log(order.amountRefundedMinor) // 0
+console.log(order.clientSecret) // present only while the Order is payable
+
+if (order.status === 'paid') {
+  // fulfil the order
+}
+```
+
+The read is anonymous on the wire (the unguessable id is the access control) and returns a buyer-safe projection: it never includes the merchant identity, the Connect account or the fee.
+
+## Order lifecycle
+
+| Status | Meaning |
+|---|---|
+| `requires_payment` | Created; the browser has not confirmed yet. `clientSecret` is available. |
+| `paid` | Payment succeeded. |
+| `failed` | Payment failed, or the PaymentIntent could not be created. No money moved. |
+| `refunded` / `partially_refunded` | Refunded in full / in part (see `amountRefundedMinor`). |
+| `disputed` | A chargeback is open. |
+
+## Error codes
+
+Errors throw `PaymentsError` with `code` set to the backend catalogue code:
+
+| Code | HTTP | Meaning |
+|---|---|---|
+| `BCK.ORDER.0001` | 400 | Invalid request (amount out of range, unsupported currency / capture mode / provider). |
+| `BCK.ORDER.0002` | 404 | No Order with this id. |
+| `BCK.ORDER.0003` | 403 | The key is not an active organization, or the amount exceeds the per-order cap. |
+| `BCK.ORDER.0004` | 500 | The merchant has no Connect account able to receive card payments. |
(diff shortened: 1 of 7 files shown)


r-marques:
:eyes: Reviewing

src/api/orders-api.ts
```diff
@@ -0,0 +1,167 @@
...
+   *
+   * @remarks
+   * The endpoint is anonymous — the unguessable id is the sole access control —
+   * so this call sends no API key (the `Payments` instance still needs one to
+   * be constructed). The response never includes the merchant identity or the
+   * fee, and carries `clientSecret` only while the Order is payable.
+   *
+   * @param orderId - The unguessable Order id returned by {@link createOrder}.
+   * @returns @see {@link Order}
+   * @throws PaymentsError with code `BCK.ORDER.0002` when no Order has this id.
```
r-marques, line 158: **🟡 SHOULD FIX — `getOrder` can 429 from the read throttle, and that 429 carries no `code`**

The `@throws` names only `BCK.ORDER.0002`, but `GET /api/v1/orders/:id` is rate-limited on the API side:

```ts
// nvm-monorepo apps/api/src/orders/orders.controller.ts (main)
const ORDERS_GET_THROTTLE = { long: { limit: 60, ttl: 60_000 } }
...
@Get(':id')
@Public()
@Throttle(ORDERS_GET_THROTTLE)
```

and the comment above it notes that **all anonymous callers share one ingress-IP bucket** — so a merchant server polling several Orders is spending from the same 60-per-minute budget as every other reader behind that IP.

That 429 is a Nest `ThrottlerException`, not an `NVMException`, so its body has no `code` field. `PaymentsError.fromBackend` reads `error.code` straight through:

```ts
return new PaymentsError(`${message}. ${error.message}`, error.code)
```

so the caller gets a `PaymentsError` with `code === undefined`. Failure scenario: a merchant follows `markdown/orders.md`, polls `getOrder` in a loop, trips the throttle, and their `if (error.code === 'BCK.ORDER.0010') backOff()` branch is false — because the doc's only 429 row is the **create-side** velocity cap. They fall through to `throw error` and the poll dies rather than backing off.

Suggested: extend the `@throws` to say the read can also return `429` from the endpoint throttle with **no catalogue code**, and add a matching row/sentence to `markdown/orders.md` distinguishing it from `BCK.ORDER.0010`. Branching on `response.status === 429` rather than on the code is the reliable test for callers.

<sub>🤖 pr-review-toolkit · self · high</sub>

src/api/orders-api.ts
```diff
@@ -0,0 +1,167 @@
...
+   * @example
+   * ```
+   *  const order = await payments.orders.getOrder(orderId)
+   *  if (order.status === 'paid') fulfil(order)
+   * ```
+   */
+  public async getOrder(orderId: string): Promise<Order> {
+    const query = API_URL_GET_ORDER.replace(':orderId', orderId)
```
r-marques: **💡 NIT — `orderId` is interpolated into the path without `encodeURIComponent`**

`API_URL_GET_ORDER.replace(':orderId', orderId)` puts caller-supplied text straight into the path, and `new URL()` then resolves it. An id carrying `?`, `#` or `..` silently retargets the request instead of producing a clean `BCK.ORDER.0002` — e.g. `getOrder('ord_x?foo=1')` requests `/api/v1/orders/ord_x` with a query string, and a `../`-bearing value normalizes to a different endpoint entirely.

No credential leaks (this call deliberately goes out via `getPublicHTTPOptions`, so no `Authorization` header rides along), which is why this is a nit and not a blocker. But the id here is buyer-facing and often arrives from a URL parameter, which is exactly where a stray character comes from.

`organizations-api.ts:223` already sets the precedent in this codebase:

```ts
const path = API_URL_ORG_ACTIVITY.replace(':orgId', encodeURIComponent(orgId))
```

Same one-word change here.

<sub>🤖 pr-review-toolkit · self · high</sub>

markdown/orders.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Order lifecycle
+
+| Status | Meaning |
+|---|---|
+| `requires_payment` | Created; the browser has not confirmed yet. `clientSecret` is available. |
+| `paid` | Payment succeeded. |
+| `failed` | Payment failed, or the PaymentIntent could not be created. No money moved. |
+| `refunded` / `partially_refunded` | Refunded in full / in part (see `amountRefundedMinor`). |
+| `disputed` | A chargeback is open. |
```
r-marques: **💡 NIT — `disputed` also covers a lost (closed) dispute, and a won one reads back as `paid`**

"A chargeback is open" describes only one of the two internal states that surface as `disputed`. The API's wire serializer collapses them:

```ts
// nvm-monorepo apps/api/src/orders/order-status.serializer.ts (main)
case 'paid':
case 'won':
  return 'paid'
...
case 'disputed':
case 'lost':
  return 'disputed'
```

So `disputed` means "a chargeback is open **or** was lost", and a **won** dispute returns the Order to `paid` — which the table's `paid` row ("Payment succeeded.") doesn't hint at either. A merchant treating `disputed` as "still being decided" will wait for a resolution that, on a win, arrives as `paid`.

Suggested: "`disputed` — A chargeback is open, or was lost. A **won** dispute returns the Order to `paid`."

<sub>🤖 pr-review-toolkit · self · high</sub>

tests/unit/orders-api.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    test('GETs /api/v1/orders/:id anonymously and returns the buyer-safe view', async () => {
+      const calls = installFetchStub(() => ({ ok: true, body: order }))
+      const result = await makePayments().orders.getOrder(order.id)
+
+      expect(result).toEqual(order)
+      const [{ url, init }] = calls
+      expect(url.pathname).toBe(`/api/v1/orders/${order.id}`)
+      expect(init.method).toBe('GET')
+      // The id IS the access control — never leak the merchant key on the read.
+      expect(init.headers.Authorization).toBeUndefined()
```
r-marques, line 155: **✅ NICE — The anonymous-read discipline is asserted, not just assumed**

This assertion is the one that makes the whole "the id is the access control" design safe to ship, and it's easy to leave out — the happy-path test would pass identically with the merchant key attached. I verified it holds for the right reason: `getPublicHTTPOptions` builds headers from scratch (`Accept`, `Content-Type`, `API_VERSION_HEADER`) and never reaches for the key, so `getOrder` cannot leak it even if someone later changes the default options.

Worth keeping as a regression guard if the two option builders are ever merged. `safeParseJson` is also doing real work on this path — it always returns an object (`{}` on an empty body, a truncated snippet for non-JSON), so the `error.message` read in `PaymentsError.fromBackend` can't throw on an HTML gateway page.

<sub>🤖 pr-review-toolkit · self · high</sub>

r-marques (commented):

A thin, well-scoped `OrdersAPI` client over the two browser-fiat Orders endpoints, wired into `Payments` alongside the other sub-APIs, with a guide and unit tests. The shape matches the rest of the SDK and the types are faithful to the backend DTOs. The one thing I'd change before merge is a documentation gap rather than a code defect: the read endpoint is throttled, and neither the JSDoc nor `orders.md` prepares a caller for that 429.

**What I reviewed**
- Diff range `origin/main...HEAD` — 7 files, +460/−0, head `84bbe6aa`. (Note the head moved from `be16d711` while this review was queued; everything below is against `84bbe6aa`.)
- Panel: **none** — the operator's standing session instruction forbids spawning subagents, so every check below is my own reading, not a panel's.
- Ran the new suite: `npx jest --config ./tests/jest.config.json --testPathPattern orders-api` in the review worktree — **6 passed, 6 total**. (Runner: jest + ts-jest. The worktree needed `node_modules`; I symlinked the base repo's rather than installing.)
- Checked every type against the backend DTOs on `nvm-monorepo@main`: `OrderStatus` matches `ORDER_STATUS_VALUES` exactly (all six, same spellings); `CreateOrderOptions` matches `RegisterOrderDto` field-for-field including the 1024/255/255 caps and the `usd` / `automatic` / `stripe` literals; `CreateOrderResult` and `Order` match `CreateOrderResponseDto` / `OrderResponseDto`, with `clientSecret` optional in both and `description` / `buyerRef` / `paymentIntentId` / `expiresAt` nullable-but-present. Deliberately omitting any Connect-account or fee field is correct and matches the DTO's own `#3136` comment.
- Confirmed the anonymous-read claim holds at the source: `getPublicHTTPOptions` sets only `Accept`, `Content-Type` and the version header — no `Authorization` — so `getOrder` genuinely sends no key, and the test asserts it.
- Confirmed `safeParseJson` always returns an object, so the `error.message` read inside `PaymentsError.fromBackend` cannot throw on an empty or non-JSON error body.
- Checked the `Payments` wiring: `orders` is declared, constructed in `initializeApi`, and included in the `setOrganizationId` fan-out; exported from `src/index.ts` with its four types.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 1 should-fix, 2 nits.

### 🧭 Inline comments (4)

- 🟡 **SHOULD FIX** — `getOrder` can 429 from the read throttle, and that 429 carries no `code` — `src/api/orders-api.ts:151`
- 💡 **NIT** — `orderId` is interpolated into the path without `encodeURIComponent` — `src/api/orders-api.ts:159`
- 💡 **NIT** — `disputed` also covers a lost (closed) dispute, and a won one reads back as `paid` — `markdown/orders.md:83`
- ✅ **NICE** — The anonymous-read discipline is asserted, not just assumed — `tests/unit/orders-api.test.ts:155`

r-marques (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 2 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/429#pullrequestreview-5124745634

src/api/orders-api.ts
aaitor: Fixed in 9a905c1. Two things, since the SDK has no way to hand callers the HTTP status: (1) `getOrder` / `createOrder` now build the error envelope with an `http_<status>` fallback `code` — a body without a catalogue code (the Nest `ThrottlerException`, a gateway page) surfaces as `http_429` / `http_502` … instead of collapsing to the generic `payments_error`; a catalogue `code` in the body always wins. This is the same convention the Python SDK's `PaymentsError.from_response` already uses, so both SDKs now agree. (2) The `@throws` on `getOrder` and `markdown/orders.md` describe the 60/min shared read throttle, name `http_429` explicitly and distinguish it from the create-side `BCK.ORDER.0010` (separate row in the error table + a 'back off on http_429' note under Read an Order). Unit test added: a 429 with a Nest body rejects with `code === 'http_429'`. Mirrored in payments-py#270.

src/api/orders-api.ts
aaitor: Fixed in 9a905c1 — `encodeURIComponent(orderId)`, same as the `API_URL_ORG_ACTIVITY` precedent. Test added: `getOrder('ord_x?foo=1')` requests `/api/v1/orders/ord_x%3Ffoo%3D1` with an empty query string. Mirrored in payments-py#270 (`urllib.parse.quote(order_id, safe='')`).

markdown/orders.md
aaitor: Fixed in 9a905c1 — the lifecycle table now reads: `paid` — Payment succeeded; also the state after a **won** dispute. `disputed` — A chargeback is open, or was lost; a won dispute returns the Order to `paid`. Same wording in the Python guide.

tests/unit/orders-api.test.ts
aaitor: Thanks — kept as-is, and the new `http_429` test exercises the same public-options path, so the no-Authorization guard now has two callers watching it.

PR #432, opened 2026-09-08
feat(x402): surface billingModel so a pay-as-you-go settle can be read as paid

r-marques:
> **This is no longer a docs-only PR.** It now carries an **additive behaviour change** to the MCP paywall (`_meta["nevermined/credits"]` gains `billingModel` and `orderTx`) alongside the type + documentation work. Nothing is renamed, retyped or removed, and both new fields are omitted when absent — but it warrants a **minor**, not a patch. Titled `feat(...)` accordingly.

## The problem, in plain terms

On a **pay-as-you-go** plan, a settle that charged the buyer $1.00 comes back like this:

```json
{ "success": true, "network": "stripe", "billingModel": "pay-as-you-go",
  "creditsRedeemed": "0", "remainingBalance": "0",
  "orderTx": "pi_3U6tgrBYvSRKcV421ehH4bnX" }
```

Those plans hold no credit balance — each settle is one direct charge, referenced by `orderTx` (fiat rails) or `transaction` (crypto rails). So the success check a credits integration already has, `success === true && creditsRedeemed > 0`, is **unreachable** there: it reports a real charge as a decline.

That is not hypothetical. During company testing of the card rails a tester's pass condition was exactly that check, the charge succeeded, and the agent running the purchase had to reason from the backend source to work out that a healthy payment was being reported as a decline. A less careful consumer would have retried — and on a card rail a needless retry is the one thing we tell people never to do, because repeated attempts feed issuer fraud scoring.

The API has always returned the discriminator `billingModel`. **`SettlePermissionsResult` did not declare it**, so it was invisible to TypeScript consumers, and the correct check could only be worked out by reading our source. This PR makes it visible and documents the criterion.

## What changed

**No wire change.** `billingModel` was already on the response; the type just did not mention it.

- `SettlePermissionsResult` gains `billingModel?: X402BillingModel` (`'credits' | 'pay-as-you-go'`), exported from `@nevermined-io/payments` and `@nevermined-io/payments/x402`.
- The interface doc carries the per-model criterion, including the string-vs-number trap: these are **strings**, so `'0'` is truthy while `Number('0') > 0` is false — two plausible-looking checks disagree.
- `creditsRedeemed` / `remainingBalance` now say they are always `'0'` on pay-as-you-go. `orderTx` and `network` descriptions were corrected against the API's own DTO (`orderTx` is the per-request charge on fiat pay-as-you-go, not only an auto-top-up hash; `network` is the rail, not the billing model).
- The class-level JSDoc example printed `creditsRedeemed` unconditionally — the exact confusion the issue is about. It now demonstrates the correct check.
- `markdown/x402.md`, `markdown/validation-of-requests.md` and `markdown/mcp-integration.md`: a "Was the buyer charged?" section, a pay-as-you-go receipt example, and `billingModel` passed through the `PAYMENT-RESPONSE` examples.

## The criterion, as documented

| `billingModel` | Success criterion | Credit fields |
| --- | --- | --- |
| `credits` | `success === true && Number(creditsRedeemed) > 0` | `creditsRedeemed` is the amount burned, `remainingBalance` what is left |
| `pay-as-you-go` | `success === true` **and** a non-empty `orderTx` (fiat) / `transaction` (crypto) | always the string `'0'` — no balance exists on this plan shape |

## Additive behaviour change: the MCP receipt

The MCP paywall attaches two `_meta` keys. `x402/payment-response` spreads the **whole** settle result, so it already carried `billingModel` and `orderTx`. The condensed `nevermined/credits` key did not — it reported `creditsRedeemed` with nothing to read it by. On a pay-as-you-go plan that field is always `'0'`, so a settle that **did** charge the buyer was indistinguishable, on that key, from a credits settle that burned nothing.

Both `_meta` builders now carry the two fields — the non-streaming site and the streaming final-chunk site, which are separate code paths with a test each.

Constraints held to:

- **Purely additive.** Existing keys keep their names, types and values.
- **Omitted, not empty,** when the settle carried neither, so a consumer can distinguish "absent" from "present and empty". Pinned by its own test.
- `markdown/mcp-integration.md` is corrected — it previously stated the condensed key does *not* carry these fields, which this change makes false.

### Mutation-tested, recording which tests go red

| Mutation | Tests killed |
| --- | --- |
| `billingModel` removed from the **non-streaming** site | the non-streaming PAYG test, alone |
| `billingModel` removed from the **streaming** site | the streaming PAYG test, alone |
| `orderTx` removed from **both** sites | both PAYG tests |
| *(control: restored)* | none — 581 passed |

Each site fails independently, so neither test is masking the other.

## Verification

- `pnpm build` (tsc) — clean.
- `pnpm test:unit` — **44 suites, 581 passed, 1 skipped, 582 total** (3 new tests).
- `pnpm lint` — 0 errors, 3 warnings, all pre-existing (`src/utils.ts`, `src/x402/express/middleware.ts`).
- `python3 scripts/lint_doc_links.py` — 14 pages, no escaping or repo-source links.
- `./scripts/check_synced_doc_links.sh` — mintlify `broken-links` over the staged `api-reference/typescript/` tree: no broken links.

## Notes for the reviewer

- `markdown/` is the **source** for the docs site's generated `api-reference/typescript/**`; the generated copy is untouched, as it must be.
- The condensed `_meta["nevermined/credits"]` gap was first raised as a finding and deliberately held; the repo owner then approved changing it, which is the second commit here.

Refs nevermined-io/nvm-monorepo#2999

### DIFF
--- a/src/mcp/core/paywall.ts
+++ b/src/mcp/core/paywall.ts
@@ -237,10 +237,16 @@ export class PaywallDecorator {
           ...(creditsResult && { [X402_PAYMENT_RESPONSE_META_KEY]: creditsResult }),
           [NEVERMINED_CREDITS_META_KEY]: {
             ...(creditsResult?.transaction && { txHash: creditsResult.transaction }),
+            // `creditsRedeemed` is meaningless without this: a pay-as-you-go plan
+            // holds no balance, so it reads '0' on a settle that DID charge the
+            // buyer. Omitted (not '') when the settle carried no discriminator,
+            // so a consumer can tell "absent" from "present and empty".
+            ...(creditsResult?.billingModel && { billingModel: creditsResult.billingModel }),
             creditsRedeemed: creditsResult?.success
               ? (creditsResult.creditsRedeemed ?? credits.toString())
               : '0',
             remainingBalance: creditsResult?.remainingBalance,
+            ...(creditsResult?.orderTx && { orderTx: creditsResult.orderTx }),
             planId: authResult.planId,
             subscriberAddress: authResult.subscriberAddress,
             success: creditsResult ? creditsResult.success : true,
@@ -396,10 +402,14 @@ function wrapAsyncIterable<T>(
         // Nevermined-namespaced observability (NOT part of the x402 spec).
         [NEVERMINED_CREDITS_META_KEY]: {
           ...(settlement?.transaction && { txHash: settlement.transaction }),
+          // See the non-streaming site: without `billingModel`, `creditsRedeemed`
+          // cannot be read — it is '0' on a pay-as-you-go settle that charged.
+          ...(settlement?.billingModel && { billingModel: settlement.billingModel }),
           creditsRedeemed: settlement?.success
             ? (settlement.creditsRedeemed ?? credits.toString())
             : '0',
           remainingBalance: settlement?.remainingBalance,
+          ...(settlement?.orderTx && { orderTx: settlement.orderTx }),
           planId,
           subscriberAddress,
           success: settlement ? settlement.success : true,
--- a/src/x402/facilitator-api.ts
+++ b/src/x402/facilitator-api.ts
@@ -219,6 +219,12 @@ export type X402BillingModel = 'credits' | 'pay-as-you-go'
  * Mind the type as well: these fields are **strings**. `'0'` is truthy while
  * `Number('0') > 0` is false, so two plausible-looking checks disagree.
  *
+ * If `billingModel` is **absent**, you are talking to a Nevermined API that
+ * predates the discriminator: apply the `credits` rule, and never read a missing
+ * discriminator as pay-as-you-go. All three fields are optional; if
+ * `creditsRedeemed` is absent too there is no balance information to check, and
+ * `success === true` is the whole answer.
+ *
  * @example
  * ```typescript
  * const settled =
@@ -250,9 +256,11 @@ export interface SettlePermissionsResult {
   /**
    * Which billing model this settle was priced under (Nevermined extension).
    *
-   * Present regardless of `success` — check `success` before treating it as
-   * evidence of a charge. Read it before `creditsRedeemed` / `remainingBalance`:
-   * see the interface docs above for the per-model success criterion.
+   * Reported whether or not the settle succeeded — so check `success` before
+   * treating it as evidence of a charge. It is absent entirely against a
+   * Nevermined API that predates the discriminator, which is why it is optional;
+   * treat that case as `credits`. Read it before `creditsRedeemed` /
+   * `remainingBalance`: see the interface docs above for the per-model criterion.
    */
   billingModel?: X402BillingModel
   /**
--- a/src/x402/facilitator-api.ts
+++ b/src/x402/facilitator-api.ts
@@ -221,9 +221,7 @@ export type X402BillingModel = 'credits' | 'pay-as-you-go'
  *
  * If `billingModel` is **absent**, you are talking to a Nevermined API that
  * predates the discriminator: apply the `credits` rule, and never read a missing
- * discriminator as pay-as-you-go. All three fields are optional; if
- * `creditsRedeemed` is absent too there is no balance information to check, and
- * `success === true` is the whole answer.
+ * discriminator as pay-as-you-go.
  *
  * @example
  * ```typescript
--- a/src/x402/facilitator-api.ts
+++ b/src/x402/facilitator-api.ts
@@ -30,13 +30,22 @@
  * })
  *
  * if (verification.isValid) {
- *   // Settle (burn) the credits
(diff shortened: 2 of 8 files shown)


src/x402/facilitator-api.ts
```diff
@@ -189,15 +235,47 @@ export interface SettlePermissionsResult {
   errorReason?: string
   /** Address of the payer's wallet */
```
aaitor, line 240: **nit**: The interface docs say `billingModel` is "Present regardless of `success`", but all three of `billingModel`/`creditsRedeemed`/`remainingBalance` are optional — and on a backend older than nvm-monorepo#2754 none of them arrive, so every settle falls to the credits rule (`creditsRedeemed > 0`), the pre-existing behavior. The "Was the buyer charged?" table would be complete with a one-line footnote naming that case ("absent `billingModel` → old backend → apply the credits rule") — the deploy-e2e harness in nvm-monorepo#3343 had to make exactly this call explicit (`settle_nomodel_zero`).

src/mcp/core/paywall.ts
```diff
@@ -237,10 +237,16 @@ export class PaywallDecorator {
           ...(creditsResult && { [X402_PAYMENT_RESPONSE_META_KEY]: creditsResult }),
           [NEVERMINED_CREDITS_META_KEY]: {
             ...(creditsResult?.transaction && { txHash: creditsResult.transaction }),
+            // `creditsRedeemed` is meaningless without this: a pay-as-you-go plan
```
aaitor, line 240: **praise** (mutation-verified): `tests/unit/mcp.test.ts` passes 22/22 at the PR head, and two targeted mutations — dropping the `billingModel` spread in the non-streaming `_meta` builder, then the same in the streaming builder — each kill exactly their paired test. That pairing matters: the two builders are separate code and the suite pins both. Also confirmed `settlePermissions` already returns `await response.json()` wholesale, so `billingModel`/`orderTx` flow from the API with no wire change — the additive surface is types, docs, and the two `_meta` builders, exactly as claimed. (Note for the record: `tests/unit/x402/langchain-decorator.test.ts` fails to resolve `@langchain/langgraph` locally — it fails identically on the PR base, so it is my sandbox missing an optional dep, not this PR; CI Unit & Integration is green.)

aaitor (commented):
Manual pass + mutation-verified both _meta builders. The additive-only approach is right — the wire already carries the fields; this makes them typed and discoverable. No blockers — one nit on the old-backend footnote, one praise inline.

aaitor (approved):
🟡 **Approved with comments** — no blockers; 0 should-fix, 1 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/432#pullrequestreview-5144152716

src/x402/facilitator-api.ts
r-marques: Folded in `990a105` — the footnote is on all six criterion tables plus both SDK type docs and the three skill files (eleven locations, one wording):

> **No `billingModel` in the response?** You are talking to a Nevermined API that predates the discriminator. Apply the **credits** rule — never read a missing discriminator as pay-as-you-go.

I also reconciled the "Present regardless of `success`" line you flagged as inconsistent, so both halves hold.

**One correction to your premise, because it changes the shape of the advice.** You wrote that on a backend older than #2754 *none* of the three fields arrive. That is true of `billingModel` only — `creditsRedeemed` has been in the settle response since **2026-03-25** (`3744bc388`), `billingModel` since **2026-08-06** (`66ed3614f`), four and a half months apart (`git log -S` on `card-delegation-scheme.handler.ts`).

It mattered: the first draft carried an extra clause for "what if `creditsRedeemed` is missing too", on the reasoning that a fix *for* old backends must not reintroduce the decline-on-success bug this PR removes. Once the dates were checked that clause described a deployment that has not existed since March, so it is gone — a conditional that can never fire is noise in a doc whose whole job is telling someone which check to run. Flagging it since it is your model of the deployed fleet, not just this PR.

And the corroboration you drew to the deploy-e2e harness was the useful part: nvm-monorepo#3343 pins exactly this in `settle_nomodel_zero`/`settle_nomodel_ok`, so the harness enforced a rule the SDK did not document. It does now.

src/mcp/core/paywall.ts
r-marques: Thank you for mutation-verifying **both** `_meta` builders separately — that they are distinct code with a test each is the thing a reader would most reasonably assume rather than check, and "each mutation kills exactly its paired test" is the only way to know neither is masking the other.

The `settlePermissions` → `await response.json()` confirmation is the one that makes the PR description honest rather than merely plausible: it is what makes "the fields already flow from the API, this adds types and docs" a checked fact instead of my claim.

Noted on `@langchain/langgraph` failing identically on the base — agreed that is a missing optional dep locally and not this PR. No action taken.

Issue #431, opened 2026-09-08
getAgentPlans pagination throws on a plain object; environment option warns but docs still pass it


robindoble:
## What happens

1. `getAgentPlans(agentId, { page, offset })` throws:

```
TypeError: pagination.asQueryParams is not a function
```

It only works with an actual `new PaginationOptions({ ... })` instance. The signature (`pagination?: PaginationOptions`) reads as a plain options object, so callers pass one and fail at runtime rather than at compile time.

2. `Payments.getInstance({ nvmApiKey, environment })` now emits a deprecation warning because the environment is derived from the key prefix. Every documented snippet and example (and the pattern copied into downstream repos) still passes `environment`, so every new integration starts with a warning.

## Repro

SDK 1.11.1, sandbox, 2026-08-26, while enumerating a merchant's plans for the nvm-pilot-demo catalog audit.

## Asks

1. Accept a plain object in `getAgentPlans` (or narrow the type so TypeScript rejects one).
2. Drop `environment` from docs and examples, or suppress the warning when the value matches the key prefix.

Tracked as release-testing PICKUP finding 13.

aaitor:
Fixed in #436 (merged as `4495be8`).

- `getAgentPlans` / `getAgentsAssociatedToAPlan` now accept a plain `{ page, offset }` object (or a `PaginationOptions` instance, or nothing) — no more `asQueryParams is not a function`.
- The `environment` deprecation warning now fires only when the passed value is actually overridden by the API key prefix, so documented snippets that pass a matching `environment` no longer warn.

(Also fixed a byproduct: the CLI command generator was emitting a broken `--pagination` string flag for the widened signature; root-caused in `isComplexType`.)

PR #436, opened 2026-09-10
fix: accept plain-object pagination and quiet redundant environment warning (#431)

aaitor:
## Why this matters

Builders integrating the SDK hit two rough edges as soon as they follow our own docs. Listing a merchant's plans with `getAgentPlans(agentId, { page, offset })` — exactly how the option object reads — crashed at runtime, so a routine "show me the plans" call blew up instead of paginating. And every new integration that copied our documented setup snippet was greeted by a scary `[DEPRECATED]` warning on startup, even though it was doing nothing wrong. Both make the first five minutes with Nevermined feel broken. This fixes them so the documented, obvious usage just works, quietly.

## Description

Resolves two DX issues reported in #431 (release-testing PICKUP finding 13):

**1. `getAgentPlans` / `getAgentsAssociatedToAPlan` throw on a plain object.**
The pagination argument was typed as a `PaginationOptions` instance but the signature reads as a plain options object, so callers passing `{ page, offset }` hit `TypeError: pagination.asQueryParams is not a function` at runtime — only an actual `new PaginationOptions({...})` worked. The parameter is now `Partial<PaginationOptions>` and normalized through the constructor, so a plain object, a class instance, or nothing all work. Backward compatible.

**2. The `environment` deprecation warning fired on every documented snippet.**
The environment is derived from the API key prefix, and passing `environment` was deprecated — but the warning fired even when the passed value *matched* the derived one (which every doc example does), nagging each new integration. It now fires only when a passed `environment` is actually overridden by the key prefix (i.e. silently ignored). A matching value, or the fallback that's genuinely used for a keyless/custom prefix, stays quiet.

### Alternatives considered
- *Pagination:* narrow the type so a plain object fails to compile (the issue's other option). Rejected — it makes the API harder to use and doesn't help JS callers; accepting the plain object is the friendlier fix.
- *Warning:* strip `environment` from all docs/examples. Rejected — endless doc churn that still leaves already-shipped downstream code warning; suppressing the redundant warning fixes every caller at once.

Related to Issue #431

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] Tests Cover Changes
- [x] Documentation

## Test plan

- `pnpm build && pnpm lint && pnpm test:unit` — all green (635 passed).
- New `tests/unit/agent-plans-pagination.test.ts`: both methods accept a plain object, still accept a `PaginationOptions` instance, and default to page 1 / pageSize 10. The plain-object case reproduces the reported `TypeError` before the fix.
- Updated `tests/unit/environment-from-key-prefix.test.ts`: matching `environment` → no warning; conflicting key prefix → exactly one warning; keyless fallback → no warning.

#### Funny gif

No more mystery crashes

https://claude.ai/code/session_01L75q2EqZS86fJxHiXz9k4q

### DIFF
--- a/cli/src/generator/command-generator.ts
+++ b/cli/src/generator/command-generator.ts
@@ -364,12 +364,18 @@ ${runMethod}
     // Strip '| undefined' suffix from optional types before checking
     let baseType = type.replace(/\s*\|\s*undefined$/, '').trim()
 
-    // Unwrap a single generic utility wrapper (e.g. `Partial<PaginationOptions>`,
+    // Unwrap a generic wrapper (e.g. `Partial<PaginationOptions>`,
     // `Readonly<FooConfig>`) so the wrapped type is what the checks below see —
     // otherwise the trailing `>` defeats the suffix match and an optional
     // options object leaks out as a raw string flag that's silently ignored.
     const wrapped = baseType.match(/^\w+<(.+)>$/)
-    if (wrapped) baseType = wrapped[1].trim()
+    if (wrapped) {
+      const inner = wrapped[1].trim()
+      // Multi-argument generics (`Record<K,V>`, `Pick<T,K>`, `Map<K,V>`, …)
+      // always denote an object type — complex, no further unwrapping needed.
+      if (inner.includes(',')) return true
+      baseType = inner
+    }
 
     // Complex objects have literal braces or are explicitly 'object'
     // Exclude template literal types like `0x${string}` which are just strings
--- a/cli/src/generator/command-generator.ts
+++ b/cli/src/generator/command-generator.ts
@@ -362,7 +362,14 @@ ${runMethod}
    */
   private isComplexType(type: string): boolean {
     // Strip '| undefined' suffix from optional types before checking
-    const baseType = type.replace(/\s*\|\s*undefined$/, '').trim()
+    let baseType = type.replace(/\s*\|\s*undefined$/, '').trim()
+
+    // Unwrap a single generic utility wrapper (e.g. `Partial<PaginationOptions>`,
+    // `Readonly<FooConfig>`) so the wrapped type is what the checks below see —
+    // otherwise the trailing `>` defeats the suffix match and an optional
+    // options object leaks out as a raw string flag that's silently ignored.
+    const wrapped = baseType.match(/^\w+<(.+)>$/)
+    if (wrapped) baseType = wrapped[1].trim()
 
     // Complex objects have literal braces or are explicitly 'object'
     // Exclude template literal types like `0x${string}` which are just strings
--- a/cli/test/unit/generator-complex-type.test.ts
+++ b/cli/test/unit/generator-complex-type.test.ts
@@ -25,6 +25,8 @@ describe('CommandGenerator.isComplexType', () => {
     'Partial<PaginationOptions> | undefined',
     'FooConfig',
     'Readonly<BarMetadata>',
+    'Record<string, string>',
+    "Pick<PaginationOptions, 'page'>",
     'object',
     '{ page: number; offset: number }',
   ])('treats %s as a complex type', (type) => {
--- a/cli/test/unit/generator-complex-type.test.ts
+++ b/cli/test/unit/generator-complex-type.test.ts
@@ -1,0 +1,40 @@
+/**
+ * Unit tests for CommandGenerator.isComplexType — the check that decides whether
+ * an SDK method parameter is an options/DTO object (skipped for optional params,
+ * or JSON-string flag otherwise) versus a plain scalar flag.
+ *
+ * Regression for nevermined-io/payments#431: a generic-wrapped options type such
+ * as `Partial<PaginationOptions>` must still be recognized as complex. Before the
+ * fix the trailing `>` defeated the `Options$` suffix match, so the param leaked
+ * out as a raw `Flags.string` that was silently ignored at runtime.
+ */
+import { expect, test, describe } from '@jest/globals'
+import { CommandGenerator } from '../../src/generator/command-generator.js'
+
+// isComplexType is private; exercise it directly (its output drives whether a
+// broken flag is emitted).
+const isComplexType = (type: string): boolean =>
+  (new CommandGenerator('/tmp/unused') as unknown as { isComplexType(t: string): boolean }).isComplexType(
+    type,
+  )
+
+describe('CommandGenerator.isComplexType', () => {
+  test.each([
+    'PaginationOptions',
+    'Partial<PaginationOptions>',
+    'Partial<PaginationOptions> | undefined',
+    'FooConfig',
+    'Readonly<BarMetadata>',
+    'object',
+    '{ page: number; offset: number }',
+  ])('treats %s as a complex type', (type) => {
+    expect(isComplexType(type)).toBe(true)
+  })
+
+  test.each(['string', 'number', 'boolean', 'string | undefined', '`0x${string}`'])(
+    'treats %s as a scalar type',
+    (type) => {
+      expect(isComplexType(type)).toBe(false)
+    },
+  )
+})
--- a/src/api/agents-api.ts
+++ b/src/api/agents-api.ts
@@ -249,7 +249,7 @@ export class AgentsAPI extends BasePaymentsAPI {
    *
    * @example
    * ```
-   *  const result = payments.agents.getAgentPlans(planId, { page: 1, offset: 10 })
+   *  const result = payments.agents.getAgentPlans(agentId, { page: 1, offset: 10 })
    *  // {
    *  //  total: 10,
    *  //  page: 1,
--- a/src/api/agents-api.ts
+++ b/src/api/agents-api.ts
@@ -242,13 +242,14 @@ export class AgentsAPI extends BasePaymentsAPI {
    * Gets the list of plans that can be ordered to get access to an agent.
    *
    * @param agentId - The unique identifier of the agent.
-   * @param pagination - Optional pagination options to control the number of results returned.p
+   * @param pagination - Optional pagination options. Accepts a plain object
+   * (`{ page, offset, sortBy, sortOrder }`) or a `PaginationOptions` instance.
    * @returns A promise that resolves to the list of all different plans giving access to the agent.
    * @throws PaymentsError if the agent is not found.
    *
    * @example
    * ```
-   *  const result = payments.agents.getAgentPlans(planId)
+   *  const result = payments.agents.getAgentPlans(planId, { page: 1, offset: 10 })
    *  // {
    *  //  total: 10,
    *  //  page: 1,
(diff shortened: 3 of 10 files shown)


r-marques:
:eyes: Reviewing

tests/unit/environment-from-key-prefix.test.ts
```diff
@@ -80,33 +80,40 @@ describe('environment resolution at Payments.getInstance', () => {
...
+    expect(deprecationWarnings()).toHaveLength(0)
   })
 
-  test('falls back to the environment option for an unrecognized key prefix', () => {
+  test('falls back to the environment option for an unrecognized key prefix without warning', () => {
     const payments = Payments.getInstance({
       nvmApiKey: keyWithPrefix('local'),
       environment: 'custom',
     })
     expect(payments.getEnvironmentName()).toBe('custom')
```
r-marques: **🟡 SHOULD FIX — This test cannot fail — `environment: 'custom'` is indistinguishable from the default**

This is the test that changed meaning in this PR (it used to assert a warning), and the comment you attached to it — *"The option is actually used (not overridden), so it is doing its job"* — is the one claim it cannot demonstrate.

`resolveEnvironmentName` is `fromKey ?? options.environment ?? 'custom'`. With an unrecognized prefix and `environment: 'custom'`, **both** branches produce `'custom'`, so the assertion holds whether or not the fallback exists.

Mutation-tested it:

```js
// base-payments.ts
-    const resolved = fromKey ?? options.environment ?? 'custom'
+    const resolved = fromKey ?? 'custom'
```

→ **all 13 tests in this file still pass**, this one included. The fallback that the PR's whole rationale rests on ("stay quiet rather than telling the caller to remove a load-bearing option") is currently unpinned: someone could delete it tomorrow and CI would be green.

**Fix** — pick an environment that is *not* the default, so the value can only have come from the option:

```js
test('falls back to the environment option for an unrecognized key prefix without warning', () => {
  const payments = Payments.getInstance({
    nvmApiKey: keyWithPrefix('local'),
    environment: 'staging_sandbox',
  })
  expect(payments.getEnvironmentName()).toBe('staging_sandbox')
  expect(deprecationWarnings()).toHaveLength(0)
})
```

I verified this version with an unmutated control: it **passes** on the current code and **fails** under the mutation above, so it genuinely pins the fallback. The no-warning assertion still holds, because `options.environment === resolved` on that path.

The sibling test on line 119 (`neither key prefix nor option resolves`) already covers the real `'custom'` default, so nothing is lost by moving this one off it.

<sub>🤖 pr-review-toolkit · self-verified · high</sub>

cli/src/generator/command-generator.ts
```diff
@@ -362,7 +362,14 @@ ${runMethod}
...
   private isComplexType(type: string): boolean {
     // Strip '| undefined' suffix from optional types before checking
-    const baseType = type.replace(/\s*\|\s*undefined$/, '').trim()
+    let baseType = type.replace(/\s*\|\s*undefined$/, '').trim()
+
+    // Unwrap a single generic utility wrapper (e.g. `Partial<PaginationOptions>`,
+    // `Readonly<FooConfig>`) so the wrapped type is what the checks below see —
+    // otherwise the trailing `>` defeats the suffix match and an optional
+    // options object leaks out as a raw string flag that's silently ignored.
+    const wrapped = baseType.match(/^\w+<(.+)>$/)
```
r-marques: **💡 NIT — The unwrap only handles single-argument generics — `Record<K, V>` and `Pick<T, K>` still read as scalars**

Fixing this in `isComplexType` rather than at the two call sites was the right call, and the comment explains the trailing-`>` mechanism well. One boundary worth pinning while you are here.

`/^\w+<(.+)>$/` with a greedy `.+` unwraps to the *whole* argument list, so a multi-argument generic comes out as something like `string, unknown`, which then matches none of the checks below. Probed against the real implementation at this head:

```
COMPLEX  Partial<PaginationOptions>
COMPLEX  Readonly<BarMetadata>
COMPLEX  Array<PaginationOptions>
scalar   Record<string, unknown>
scalar   Pick<PaginationOptions, "page">
scalar   Map<string, string>
scalar   PaginationOptions[]
```

A `Record<string, unknown>` parameter would therefore leak out as a raw `Flags.string` and be silently ignored at runtime — the exact failure this commit exists to close, one generic away.

**This is latent, not live:** I grepped every public method in `src/api/*.ts` and none currently takes a `Record<>`, `Map<>`, `Pick<>`, `Omit<>` or complex-array parameter, so nothing is broken today. It only bites whoever adds the first one, and it will bite the same silent way.

Cheapest hardening is to treat *any* generic wrapper as complex rather than trying to unwrap it precisely — `if (/^\w+<.+>$/.test(baseType)) return true` would cover all of the above — or, if you prefer the current shape, take the first type argument (`.split(',')[0]`) so `Record<string, unknown>` resolves on its value type.

Either way, the natural place to lock it is the `test.each` list you just added in `generator-complex-type.test.ts`: adding `Record<string, unknown>` to whichever list you decide it belongs in makes the boundary a decision rather than an accident.

<sub>🤖 pr-review-toolkit · self-verified · high</sub>

cli/test/unit/generator-complex-type.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    'Partial<PaginationOptions>',
+    'Partial<PaginationOptions> | undefined',
+    'FooConfig',
+    'Readonly<BarMetadata>',
+    'object',
+    '{ page: number; offset: number }',
+  ])('treats %s as a complex type', (type) => {
+    expect(isComplexType(type)).toBe(true)
+  })
+
```
r-marques, line 35: **✅ NICE — Fixed the root cause and wired the regression test into the script that actually runs**

Worth calling out, because the tempting fix here was the local one — keep the hand-written flag and `JSON.parse` it — and that would have been quietly undone by the next `cli-sync-and-test` auto-sync, since the generator rewrites these files unconditionally and neither `MANUALLY_MAINTAINED_COMMANDS` nor a `customFlags` metadata entry covered them.

Going to `isComplexType` instead fixes it for every future optional options-object parameter, not just `pagination`. Three things I checked that make it stick:

- The two command files are back to **byte-identical generated form** — `sync-check` reports all 44 commands in sync, so the flag cannot reappear on the next sync.
- The test exercises the private method directly through a cast rather than asserting on generated output, which is the level the bug actually lives at.
- **`test:unit` in `cli/package.json` was updated to include the new file.** That script names its test files explicitly, so a new spec that is not added there is collected by nothing — adding it is what makes this a regression test rather than a file that exists.

The negative cases (`string`, `` `0x${string}` ``) are there too, so the template-literal carve-out above stays pinned.

<sub>🤖 pr-review-toolkit · self-verified · high</sub>

r-marques (commented):

Fixes `getAgentPlans` / `getAgentsAssociatedToAPlan` to accept a plain `{ page, offset }` object, and narrows the `environment` deprecation warning to the case where the passed value is actually overridden. Both changes are right, and the SDK fix is complete — I confirmed those two are the *only* public methods taking a pagination param, so there is no third one still on the old signature.

**Head moved during the review, and the new commit is a better fix than the one I was going to ask for.** I started at `af9abcb9`, where the CLI carried a hand-added `'pagination': Flags.string(...)`. That had two independent defects, both since resolved by `5330f9cc` — recorded here only so the reasoning is on the record, not as anything to action:
- It was **inert**. `flags['pagination']` is a string, and `new PaginationOptions("…")` reads `.page`/`.offset` off a string → `undefined` → defaults. Measured every input form (`{"page":2,"offset":25}`, `page=2`, `2`, unset): all produced `page=1 offset=10`. The type system could not catch it either — `this.parse(this.constructor as any)` widens `flags` to `any`, which I confirmed with a probe (assigning `flags['pagination']` to `null` raised no error).
- It sat in a **generated** file. `CommandGenerator.writeCommand` does an unconditional `writeFile`, the two commands are not in `MANUALLY_MAINTAINED_COMMANDS` (only `x402token/get-x402-access-token` is), and both `generator-metadata.json` files are empty — so no `customFlags` entry protected it. The next auto-sync would have deleted the flag, and `cli-sync-and-test.yml` commits that itself.

`5330f9cc` fixes the **root cause** in `isComplexType` instead, reverts both commands to generated form, and pins it with a regression test. That is the right shape.

**What I reviewed**
- Diff range `origin/main...HEAD` — 8 files, +178/−29, head `5330f9cc` (re-anchored after the mid-review push; the delta `af9abcb9..5330f9cc` was read in full).
- Panel: none — no `pr-review-toolkit` agents were spawned in this session, so **this is my own verification only**, not multi-agent coverage.
- Ran `sync-check`: **all 44 commands match** — the CLI is back in sync with the generator, so the reverted flag cannot reappear or be clobbered.
- Probed the real `isComplexType` at this head: `Partial<PaginationOptions>`, `Readonly<X>`, `Array<X>`, `Partial<X> | undefined` all classify COMPLEX; see the one nit for what still slips through.
- Mutation-tested the environment fallback: deleting `?? options.environment` from `resolveEnvironmentName` leaves **all 13 tests in `environment-from-key-prefix.test.ts` green**. That is the should-fix below.
- Checked the new warning text is honest: it fires only when `fromKey` is set *and* differs, so "derived from the API key" is accurate on every path that reaches it.
- Tests: SDK 620 passed (1 suite fails locally on a missing optional `@langchain/langgraph` dep in my environment — unrelated to this PR); CLI `test:unit` 4 suites / 31 passed, including the new `generator-complex-type.test.ts` now wired into the script.
- ⚠️ **No CI has run on this PR yet** — `check-runs` on `5330f9cc` is 0 and the combined status is `pending`; the runs for the previous head were still `in_progress`. Worth confirming green before merge rather than reading my local numbers as CI.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 1 should-fix, 1 nit.

### 🧭 Inline comments (3)

- 🟡 **SHOULD FIX** — This test cannot fail — `environment: 'custom'` is indistinguishable from the default — `tests/unit/environment-from-key-prefix.test.ts:113`
- 💡 **NIT** — The unwrap only handles single-argument generics — `Record<K, V>` and `Pick<T, K>` still read as scalars — `cli/src/generator/command-generator.ts:371`
- ✅ **NICE** — Fixed the root cause and wired the regression test into the script that actually runs — `cli/test/unit/generator-complex-type.test.ts:33`

r-marques (approved):

Inline review: https://github.com/nevermined-io/payments/pull/436#pullrequestreview-5167517952

tests/unit/environment-from-key-prefix.test.ts
aaitor: Fixed in b0a201f — the test now passes `environment: 'sandbox'` (distinct from the ultimate `custom` default) and asserts `getEnvironmentName() === 'sandbox'`, so dropping `?? options.environment` from `resolveEnvironmentName` now fails it. Confirmed via the same mutation.

cli/src/generator/command-generator.ts
aaitor: Fixed in b0a201f — `isComplexType` now returns true for a multi-argument generic (top-level comma in the type args), so `Record<K,V>` / `Pick<T,K>` / `Map<K,V>` classify as complex object types. Added `Record<string, string>` and `Pick<PaginationOptions, 'page'>` to the generator test.

cli/test/unit/generator-complex-type.test.ts
aaitor: Thanks — and thanks for the mutation-testing the fallback path; that caught a genuinely inert test.

src/api/agents-api.ts
```diff
@@ -242,13 +242,14 @@ export class AgentsAPI extends BasePaymentsAPI {
...
-   * @param pagination - Optional pagination options to control the number of results returned.p
+   * @param pagination - Optional pagination options. Accepts a plain object
+   * (`{ page, offset, sortBy, sortOrder }`) or a `PaginationOptions` instance.
    * @returns A promise that resolves to the list of all different plans giving access to the agent.
    * @throws PaymentsError if the agent is not found.
    *
    * @example
    * ```
-   *  const result = payments.agents.getAgentPlans(planId)
+   *  const result = payments.agents.getAgentPlans(planId, { page: 1, offset: 10 })
```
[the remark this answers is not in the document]
aaitor: Fixed in b0a201f — the JSDoc example now calls `getAgentPlans(agentId, ...)`.

PR #437, opened 2026-09-10
test: constrain settle/verify doubles to the model, and typecheck the test tree (#433)

### LINKED ISSUE #433: PaymentsMock takes `any`, so untyped settle/verify test doubles can drift from the model unchecked
Sibling of nevermined-io/payments-py#273 — same defect class, different shape, and here it is the type system being opted out of rather than absent.

Found while adding `billingModel` to `SettleResponse` for nevermined-io/nvm-monorepo#2999 (PR #432). In the Python SDK the equivalent change turned 28 tests red because its doubles had drifted. In this repo it turned **none** red — and that is not because the doubles are honest.

## Two populations

**Typed, and genuinely guarded (4 doubles).** The langsmith/langchain decorator and span tests annotate their doubles `SettlePermissionsResult`, so `tsc` checks them. Adding a *required* field to the model breaks them loudly, which is the behaviour we want.

**Untyped, and unguarded (the rest).** `PaymentsMock(settleResult?: any)` takes `any`, which erases the check entirely — an object literal passed through it can omit every field on the model and nothing complains, at compile time or run time. **25 test files reference `creditsRedeemed`**, and none of the untyped ones would notice a model change.

PR #432 needed no double fixes here, but only because the SDK reads the new fields off a conditional spread that tolerates `undefined`. A future field read unconditionally has the same latent failure the Python SDK just demonstrated — it simply surfaces as `undefined` flowing onward rather than as a thrown `AttributeError`, which is arguably worse because it is silent.

## Suggested fix

Two small changes, either of which restores the check:

1. Narrow `PaymentsMock`'s `settleResult` from `any` to `SettlePermissionsResult` (or `Partial<>` of it if partial fixtures are genuinely wanted — `Partial` still rejects a *misspelled* key, which `any` does not).
2. Have the object-literal doubles assert with `satisfies SettlePermissionsResult`, which type-checks the literal without widening its inferred type.

Both turn "a fixture has drifted from the model" into a compile error, which is where this belongs. Note the repo's `typecheck` is a separate gate from its tests, so this only has teeth if the doubles are in a typechecked path — worth confirming while fixing.

## Related

- nevermined-io/payments-py#273 — the Python half: 12 of 12 doubles drifted, 28 tests red
- nevermined-io/payments#432 — where this surfaced
- nevermined-io/nvm-monorepo#2999 — the originating issue

r-marques:
Closes #433.

`PaymentsMock` took `settleResult?: any`. That erases the check outright — an object literal passed through it could omit every field on the model and nothing complained, at compile time or run time. This restores it, and then makes the restored check actually run.

## What was wrong, beyond the `any`

The issue asks to confirm whether the doubles sit in a typechecked path. **They did not — nothing in this repo typechecks a test file.** Two independent causes, both measured:

| gate | sees a type error in `tests/`? | why |
|---|---|---|
| `pnpm build` (`tsc`) | **no** | root `tsconfig.json` has `"exclude": [… "tests"]` |
| `pnpm test:unit` (ts-jest) | **no** | root `isolatedModules: true` → ts-jest uses `transpileModule` |

Method: a file under `tests/unit/` containing `const bad: SettlePermissionsResult = { success: 123, …, nope: 'drift' }` compiled clean under `tsc -p tsconfig.json` **and passed as a jest test**. Injecting `isolatedModules: false` into the ts-jest transform makes that same file fail with `TS2322` — the control that identifies the cause rather than guessing it.

`tests/tsconfig.json` was already in the repo, already correct, and referenced by no script and no workflow.

So option 1 or 2 from the issue, shipped alone, would have been documentation. Hence commits 2 and 3.

## How to read this

Four commits, ordered so the load-bearing part is separable from the bulk:

| | commit | size | what it is |
|---|---|---|---|
| 1 | `test:` constrain the doubles | 20 files | the fix the issue asks for |
| 2 | `test:` fix the 36 | 5 files | **mechanical**; skim it |
| 3 | `ci:` typecheck the test tree | 2 files | **the part that changes what your PRs must pass** |
| 4 | `docs:` CLAUDE.md | 1 file | why `pnpm build` can't see `tests/` |
| 5 | `test:` review fold | 3 files | closes an escape hatch this branch had opened |

**Commit 2 is mechanical and every one of its 36 errors predates this PR** — they are on `origin/main` right now, and 32 of them are the same one-line omission repeated. Nothing in it changes what a test asserts; the suite's numbers are identical either side. Read commit 3 first if you are short of time.

## Commit 1 — constrain the doubles

Two mechanisms, each where it fits, which is what the issue proposes:

- **Narrowed parameters** where a mock or helper took `any` — `PaymentsMock(settleResult)`, `makeDecorator({ settle })`, `PaymentsMock(verifySequence)`. This checks every call site at once.
- **`satisfies`** on the object literals handed to `mockResolvedValue`, and explicit return annotations on mock methods that returned an inferred or hand-rolled shape.

**`SettlePermissionsResult`, not `Partial<>`.** `Partial` still rejects a misspelled key, so it is not a cop-out — but it forfeits the property the issue names as the goal: *"adding a required field to the model breaks them loudly, which is the behaviour we want."* Under `Partial` it would not. The cost of the full type was two shorthand fixtures that omitted `transaction`/`network`, and since the real facilitator never returns that shape, completing them is a fidelity improvement rather than a concession.

Every constraint targets a model type imported from `src/` — `SettlePermissionsResult`, `VerifyPermissionsResult`, `MppSettleResult`, `StartAgentRequest`, `CreateDelegationPayload`. None is inferred from the fixture it checks. Two of these doubles previously *were* hand-rolled inline shapes (`Promise<{ success: boolean; transaction: string; network: string; creditsRedeemed: string }>`), which is that failure mode written out; those are gone.

Three doubles had already drifted and are corrected:

- `x402-middleware-settlement` — `baseSettlement` had no `transaction`/`network`
- `mpp/middleware-x402-fallback` — the settle stub had no `transaction`/`network`
- `mpp/middleware-challenge` — the settle stub was `{ success: true }` and nothing else

A fourth, `mpp/middleware-hooks-parity`, stood in for `StartAgentRequest` with two of its seven fields. Completing it puts a real `bigint` balance into `req.paymentContext`, which `JSON.stringify` refuses — so that file's default handler now narrows bigints, as any seller relaying a payment context must.

## Commit 2 — make the check run

Adds `pnpm typecheck:tests` (`tsc --noEmit -p tests/tsconfig.json`) and a step in `testing.yml`'s `lint_build` job. Turning it on surfaced **36 pre-existing errors** — the same drift, undetected for the same reason:

- **32 ×** `mcp.configure({ agentId, serverName })` with no `planId`, which `McpConfig` has declared **required** since #391 (*"make agentId optional, require planId"*). The test call sites were never updated because nothing checked them. Each now passes the planId its own paywalled handlers already use. `withPaywall` reads the per-tool option first, so behaviour is unchanged — the suite's numbers are byte-identical before and after.
- **3 ×** `a2a/payments-request-handler-stream` — an `AgentCard` stub with one of nine fields, a `DummyExecutor` missing `cancelTask`, a `validation` standing in for `StartAgentRequest` with one of seven.
- **1 ×** `x402-visa` — a `currency` that widened to `string` because the payload literal was declared apart from its use.

### Were those 32 fixtures wrong, or is `McpConfig` over-demanding?

Worth asking, because they pass today only because `withPaywall` accepts a per-tool `planId` — so a type demanding a server-level one could be the thing that is inaccurate, and teaching 32 fixtures to satisfy a wrong type would be the wrong fix.

**Checked, and the fixtures were simply incomplete.** No test anywhere uses more than one planId per server — `unit/mcp` uses `plan123` throughout, `mcp_paywall_invalid_token_flow` `plan-basic`, `mcp-integration` `plan-123`; even the spec named `multi-tool-server` varies *credits*, not plans. So none of them was describing a per-tool-only pattern that `McpConfig` forbids. They omitted a value that is constant across each file, which is the drift this issue is about, and #391 made it required deliberately.

**But there is a real inconsistency next door, and it is not this PR's to fix.** `McpConfig.planId` is required, while `withPaywall`'s own docblock says the opposite — *"A `planId` must already be resolvable (per-tool `options.planId` **or** a server-level one set via `configure()`/`start()`)"* — and the runtime agrees with the docblock (`if (!opts.planId && !extendedConfig.planId) throw`). A server wanting a different plan per tool is a coherent shape that the implementation and its documentation both permit and the type refuses. `McpConfig` is production and stays untouched here; flagging it for a separate call.

## Mutation test

A guard never seen red proves nothing, so both mutations were run against the gate that now runs in CI, with an unmutated control required to pass.

**Control** — clean tree, `pnpm typecheck:tests` → exit **0**.

**M1 — add a required `settledAt: string` to `SettlePermissionsResult`.** `pnpm typecheck:tests` → exit **2**. Per-file, subtracting each side's own baseline so the count is attributable to the mutation:

| | test files red | new errors |
|---|---|---|
| `origin/main` | **4** | 9 |
| this branch | **21** | 56 |

The 4 on main are exactly the already-typed doubles the issue named (`langsmith-spans`, `langchain-decorator`, `langsmith-decorator`) plus `a2a/x402-a2a`. The **17 newly guarded**: `unit/mcp`, `unit/mcp/x402_inband`, `unit/mpp/{middleware-redeem, middleware-x402-fallback, middleware-challenge, middleware-challenge-error, middleware-body-digest, middleware-concurrent-credential, middleware-hooks-parity}`, `unit/x402-middleware-settlement`, `unit/x402-middleware-v3-relay`, `integration/mcp-integration`, `integration/mcp/mcp_paywall_invalid_token_flow`, `integration/a2a/{complete-message-send-flow, x402-inband-flow, payments-server, payments-server-additional}`.

**M2 — misspell one key in one fixture** (`creditsRedeemed` → `creditsRedeemd`). This one touches no `src` at all, so it isolates the doubles guard from the fact that M1 also breaks two `src` files:

- `pnpm typecheck:tests` → exit **2**: `TS2561: 'creditsRedeemd' does not exist in type 'SettlePermissionsResult'. Did you mean to write 'creditsRedeemed'?`
- the same misspelling applied to `origin/main`'s copy of that fixture → **no error**
- `pnpm build` → exit **0**, and `pnpm test:unit` → 635 passed. Both blind, which is the point of commit 2.

**M3 — does the CI job itself go red?** A gate that runs but cannot fail is the same false green as a suite reporting `total: 0`, so local `tsc` is not sufficient evidence. One of the 36 was reintroduced (the cheapest — a single `mcp.configure` losing its `planId`) and pushed to a throwaway branch:

```
JOB Lint & Build: failure
     - Build:           success     <- blind, as expected
     - Typecheck tests: failure
     - Lint:            skipped
JOB OpenClaw Plugin Check: skipped
JOB CLI Sync Check:        skipped
JOB Unit & Integration:    skipped
JOB E2E:                   skipped
```

```
##[error]tests/unit/mcp.test.ts(113,21): error TS2345: Argument of type
'{ agentId: string; serverName: string; }' is not assignable to parameter of
type 'ExtendedMcpConfig'.
##[error]Process completed with exit code 2.
```

One file red, which is right — exactly one omission was reintroduced — and every downstream job is gated on it. The throwaway branch has been deleted.

All three mutations were reverted; the tree is clean.

## Scope of the gate — it does NOT cover the in-tree consumers

`pnpm typecheck:tests` covers `tests/` only. **`cli/` and `openclaw/` remain unchecked**, and a reviewer would reasonably read "typecheck the test tree" as covering everything, so stating it: `cli/tsconfig.json:24` excludes `test` and `openclaw/tsconfig.json:20` excludes `tests`; neither has a test-specific tsconfig; and both have real suites (`cli/test/{unit,integration}`, `openclaw/tests/plugin.test.ts`). The same drift can still land there. Extending to them needs a tsconfig per package and whatever errors that surfaces, which is a separate change.

Also pre-existing and unchanged: `testing.yml` is `on: [push]`, so it never fires on fork PRs. Untouched here.

## What the review round changed

Commit 5 folds six review findings. The first is the one that mattered:

**An escape hatch this branch had opened.** `payments-request-handler-stream` widened `as AgentCard` to `as unknown as AgentCard` — the one assertion the typechecker cannot see through — which would have left that fixture's drift permanently invisible, in the PR whose purpose is making such drift visible. It needed no escape: `AgentCard` has nine required fields, eight of them primitives. The fixture is now complete and carries **no assertion at all**. Verified by mutation: deleting `defaultOutputModes` from it now fails with `TS2741 … but required in type 'AgentCard'`, where under the double assertion it was silent. No `as unknown as`, `as any`, `@ts-ignore` or `@ts-expect-error` remains anywhere in this branch's diff.

**Three invented fixture values, corrected.** The new `PaymentsMock` default paired `success: true` with `transaction: ''`, but the model documents an empty `transaction` as the **failure** marker — a response the facilitator never sends, teaching every fixture built from it a contradiction. It is now an ordinary successful crypto settle, with `creditsRedeemed` still absent so inheritors keep exercising the paywall's fallback. `x402-middleware-settlement` had gained `network: 'stripe'`, pinning a fiat rail inside a test about response methods; now a crypto rail, coherent with the hex `orderTx` already there.

**And a silently-changed invariant.** That default had also moved what a pre-existing assertion proved — from *"no `transaction` key ⇒ no txHash"* to *"empty `transaction` ⇒ no txHash"*, since the paywall gates on truthiness. Both are real invariants; only one was written down, and the test stayed green either way. It now passes its own explicit fixture and says which it means, so a future change to the default cannot move it again. Verified by mutation: giving that fixture a real hash makes the test **fail** (1 failed / 21 passed), so the assertion is load-bearing rather than vacuous.

**Ordering.** The `Typecheck tests` step now runs **after** `Lint`, so a test type error no longer fail-fasts the job and masks lint errors that used to surface in the same run.

## Deliberately not typed

Raw `fetch` / `Response` stubs (`x402-token-v3`, `x402-visa`, `mpp-api`, the two `integration/mpp` specs) model **what a server sends**, not an SDK return value. One test in `x402-token-v3` sends a body with **no `success` field on purpose**, to prove the SDK reports rather than trusts it — constraining that literal to the model would make the test unwritable. Untrusted wire input is a different thing from a typed API surface, and only the latter is in scope here.

## One thing found and deliberately left out of scope

`tests/unit/a2a/payments-request-handler.test.ts` settles with `{ txHash: '0xabc', amountOfCredits: 5n }` — neither field exists on `SettlePermissionsResult`. It is not accidental drift: `paymentsRequestHandler.ts` reads `response.txHash ?? response.transaction`, so the double covers a fallback the model does not describe. Typing it would delete that coverage, so it is left alone.

Traced since: `txHash` is **not** on the backend's settle response (nvm-monorepo's `x402-responses.dto.ts` declares `transaction?: string` and no `txHash`; every `txHash` in the API's x402 tree is an internal DB column), and `amountOfCredits` is on no type at all. Both are read off `executeRedemption`, which returns `Promise<any>`, and a later `response as SettlePermissionsResult` asserts a type the value was never checked against — **this issue's own defect, in production code rather than tests**, undetected by the same mechanism.

The fix is to type that return and let the compiler force the reconciliation, which is a production change needing its own review. Out of scope here; being handled separately.

## Verification

Run locally on this branch; the same numbers as `origin/main` @ `4495be8` except where noted.

| | baseline (`origin/main`) | this branch |
|---|---|---|
| `pnpm test:unit` | 48 suites, 635 passed, 1 skipped | **48 suites, 635 passed, 1 skipped** |
| `pnpm test:integration` | 11 suites, 55 passed | **11 suites, 55 passed** |
| `pnpm build` | exit 0 | **exit 0** |
| `pnpm lint` | 0 errors, 3 warnings | **0 errors, 3 warnings** (all pre-existing, in `src/`) |
| `pnpm typecheck:tests` | *(did not exist)* — 36 errors | **exit 0** |

No production code is changed.

## Related

- nevermined-io/payments-py#273 — the Python half of the same defect class, where 12 of 12 doubles had drifted and the equivalent model change turned 28 tests red. Being handled separately; nothing in this PR touches that repo.
- #432 — where this surfaced
- nevermined-io/nvm-monorepo#2999 — the originating issue

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -31,14 +31,18 @@ jobs:
       - name: Build
         run: pnpm build
 
+      - name: Lint
+        run: pnpm lint
+
       # `pnpm build` compiles src only — tsconfig.json excludes `tests`, and
       # ts-jest transpiles without checking (the root `isolatedModules: true`),
       # so nothing else in CI ever typechecks a test file.
+      #
+      # Last in the job on purpose: it is the newest and noisiest gate, and
+      # running it ahead of Lint would fail-fast past lint errors that used to
+      # surface in the same run.
       - name: Typecheck tests
         run: pnpm typecheck:tests
-
-      - name: Lint
-        run: pnpm lint
 
   cli_sync_check:
     name: CLI Sync Check
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -31,6 +31,12 @@ jobs:
       - name: Build
         run: pnpm build
 
+      # `pnpm build` compiles src only — tsconfig.json excludes `tests`, and
+      # ts-jest transpiles without checking (the root `isolatedModules: true`),
+      # so nothing else in CI ever typechecks a test file.
+      - name: Typecheck tests
+        run: pnpm typecheck:tests
+
       - name: Lint
         run: pnpm lint
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -105,14 +105,42 @@ When writing examples or documentation:
 Before submitting changes, run these checks:
 
 ```bash
-pnpm build && pnpm lint && pnpm test:unit
+pnpm build && pnpm lint && pnpm typecheck:tests && pnpm test:unit
 ```
 
 For full validation including integration tests:
 
 ```bash
-pnpm build && pnpm lint && pnpm test:unit && pnpm test:integration && pnpm test:e2e
+pnpm build && pnpm lint && pnpm typecheck:tests && pnpm test:unit && pnpm test:integration && pnpm test:e2e
 ```
+
+### `pnpm build` does NOT typecheck the tests — `pnpm typecheck:tests` does
+
+Two separate reasons, so neither existing gate catches a type error under `tests/`:
+
+- `pnpm build` is `tsc` against the root `tsconfig.json`, whose `exclude` lists `tests`.
+- **ts-jest transpiles without checking**, because the root config sets `isolatedModules: true`
+  and ts-jest honours that by using `transpileModule`. A test file asserting
+  `const x: SettlePermissionsResult = { success: 123, nope: 'drift' }` **passes**.
+
+`pnpm typecheck:tests` (`tsc --noEmit -p tests/tsconfig.json`) is the only thing that reads
+them, and it runs in CI's `lint_build` job as of #437. Run it after touching a test file —
+particularly a fixture — because neither a green build nor a green suite says anything about
+whether that fixture still matches the model.
+
+### Test doubles must be constrained to a model type imported from `src/`
+
+`PaymentsMock` used to take `settleResult?: any`, which let every settle/verify double drift
+from `SettlePermissionsResult` unnoticed (#433). Doubles now either narrow the *parameter*
+(`PaymentsMock(settleResult: SettlePermissionsResult)`) or `satisfies` the literal. Two rules:
+
+- **Constrain against the model, never a shape written next to the fixture.** A
+  `Promise<{ success: boolean; transaction: string; … }>` annotation is not a check; it is the
+  fixture repeated, and it goes stale in exactly the same way.
+- **Raw `fetch`/`Response` stubs stay untyped on purpose.** They model what a *server* sends,
+  and some specs post a deliberately malformed body — `x402-token-v3` sends one with no
+  `success` field to prove the SDK reports rather than trusts it. Constraining those to the
+  model would make such a test unwritable.
 
 ### After Modifying Source Files
 
@@ -122,7 +150,8 @@ When you modify source files, especially:
 - Authentication/authorization logic
 
 You MUST:
-1. Run `pnpm build` to verify TypeScript compilation
+1. Run `pnpm build` to verify TypeScript compilation, and `pnpm typecheck:tests` if you touched
+   anything under `tests/` — `pnpm build` cannot see those files (see above)
 2. Run `pnpm test:unit` to verify unit tests pass
 3. If modifying E2E-related code, run `pnpm test:e2e`
 4. If changing public interfaces (function signatures, options, types, response fields), update the corresponding documentation in `markdown/` to reflect the changes. Validate with `./scripts/generate-docs.sh`. On merge to main, the `update-docs.yml` workflow runs `generate-docs.sh` and auto-creates a PR with any changes. On tag push, `publish-docs.yml` publishes to docs_mintlify. However, `generate-docs.sh` only validates structure — it does NOT auto-update code examples, so those must be updated manually.
@@ -174,7 +203,7 @@ CI is configured in `.github/workflows/testing.yml` and runs on every push:
 
 | Job | Description | Depends On |
 |-----|-------------|------------|
-| `lint_build` | Install, build, lint | - |
+| `lint_build` | Install, build, typecheck tests, lint | - |
 | `cli_sync_check` | Build + test CLI against local SDK | lint_build |
 | `openclaw_check` | Build + test OpenClaw against local SDK | lint_build |
 | `unit_integration` | Unit + integration tests | lint_build |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -203,7 +203,7 @@ CI is configured in `.github/workflows/testing.yml` and runs on every push:
 
 | Job | Description | Depends On |
 |-----|-------------|------------|
-| `lint_build` | Install, build, typecheck tests, lint | - |
+| `lint_build` | Install, build, lint, typecheck tests | Typecheck tests runs LAST on purpose — see the comment in `testing.yml`: it is the newest and noisiest gate, and running it before Lint would fail-fast past lint errors that used to surface in the same run. Do not reorder to match a doc; fix the doc. |
 | `cli_sync_check` | Build + test CLI against local SDK | lint_build |
 | `openclaw_check` | Build + test OpenClaw against local SDK | lint_build |
 | `unit_integration` | Unit + integration tests | lint_build |
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   "scripts": {
     "clean": "rm -rf ./dist/ ./doc/ ./.nyc_output",
     "build": "tsc",
+    "typecheck:tests": "tsc --noEmit -p tests/tsconfig.json",
     "build:all": "pnpm build && pnpm cli:build",
     "lint": "eslint ./src",
     "test": "jest --verbose --config ./tests/jest.config.json",
--- a/tests/unit/mcp/x402_inband.test.ts
+++ b/tests/unit/mcp/x402_inband.test.ts
@@ -16,6 +16,7 @@
 import { PaywallDecorator } from '../../../src/mcp/core/paywall.js'
 import { PaywallAuthenticator } from '../../../src/mcp/core/auth.js'
 import { buildPaymentRequiredForPlans } from '../../../src/x402/facilitator-api.js'
+import type { SettlePermissionsResult } from '../../../src/x402/facilitator-api.js'
 import { PaymentRequiredError, SettlementFailedError } from '../../../src/mcp/utils/errors.js'
 import {
   NEVERMINED_CREDITS_META_KEY,
@@ -50,7 +51,7 @@ const SAMPLE_PR = {
 
 /** Build a PaywallDecorator wired with mocks for a single call. */
 function makeDecorator(opts: {
-  settle?: any
+  settle?: SettlePermissionsResult
   authenticate?: jest.Mock
   credits?: bigint
   agentId?: string
@@ -59,7 +60,8 @@ function makeDecorator(opts: {
     getEnvironmentName: () => 'staging_sandbox',
     facilitator: {
       settlePermissions: jest.fn(
-        async () => opts.settle ?? { success: true, transaction: '', network: '' },
+        async (): Promise<SettlePermissionsResult> =>
+          opts.settle ?? { success: true, transaction: '', network: '' },
       ),
     },
     agents: { getAgentPlans: jest.fn(async () => ({ plans: [] })) },
@@ -182,7 +184,7 @@ describe('x402 in-band: paywall wrapper', () => {
       network: 'eip155:84532',
       payer: '0x123',
       creditsRedeemed: '5',
-    }
+    } satisfies SettlePermissionsResult
     const { decorator } = makeDecorator({ settle })
     const handler = async () => ({ content: [{ type: 'text', text: 'ok' }] })
     const wrapped = decorator.protect(handler, { kind: 'tool', name: 'premium' })
@@ -204,7 +206,7 @@ describe('x402 in-band: paywall wrapper', () => {
       network: 'eip155:84532',
       payer: '0x123',
       creditsRedeemed: '5',
-    }
+    } satisfies SettlePermissionsResult
     // Configure with a planId but NO agentId — must not throw the
     // "missing agentId" misconfiguration; the plan-centric path settles fine.
     const { decorator } = makeDecorator({ settle, agentId: undefined })
@@ -348,7 +350,12 @@ describe('x402 in-band: paywall wrapper', () => {
     payments.facilitator.settlePermissions = jest.fn(async () => {
       calls += 1
       if (calls === 1) throw new Error('primary settle failed')
-      return { success: true, transaction: '0xok', network: 'eip155:84532', creditsRedeemed: '5' }
+      return {
+        success: true,
+        transaction: '0xok',
+        network: 'eip155:84532',
+        creditsRedeemed: '5',
+      } satisfies SettlePermissionsResult
     })
     const handler = async () => ({ content: [{ type: 'text', text: 'ok' }] })
     const wrapped = decorator.protect(handler, { kind: 'tool', name: 'premium' })
@@ -465,12 +472,14 @@ describe('x402 in-band: real MCP-SDK dispatch', () => {
     const payments: any = {
       getEnvironmentName: () => 'staging_sandbox',
       facilitator: {
-        settlePermissions: jest.fn(async () => ({
-          success: true,
-          transaction: '0xok',
-          network: 'eip155:84532',
-          creditsRedeemed: '5',
-        })),
+        settlePermissions: jest.fn(
+          async (): Promise<SettlePermissionsResult> => ({
+            success: true,
+            transaction: '0xok',
+            network: 'eip155:84532',
+            creditsRedeemed: '5',
+          }),
+        ),
       },
       agents: { getAgentPlans: jest.fn(async () => ({ plans: [] })) },
     }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ development workflow (build, lint, tests, releases) and
 
 ```bash
 pnpm install
-pnpm build && pnpm lint && pnpm test:unit   # the pre-PR check
+pnpm build && pnpm lint && pnpm typecheck:tests && pnpm test:unit   # the pre-PR check
 ```
 
 ## Documentation
--- a/tests/integration/a2a/complete-message-send-flow.test.ts
+++ b/tests/integration/a2a/complete-message-send-flow.test.ts
@@ -7,6 +7,10 @@ import request from 'supertest'
 import { PaymentsA2AServer } from '../../../src/a2a/server.js'
 import type { AgentCard, ExecutionEventBus, PaymentsAgentExecutor } from '../../../src/a2a/types.js'
 import type { Payments } from '../../../src/payments.js'
+import type {
+  SettlePermissionsResult,
+  VerifyPermissionsResult,
+} from '../../../src/x402/facilitator-api.js'
 
 jest.mock('../../../src/utils.js', () => ({
   decodeAccessToken: jest.fn(() => ({
@@ -36,7 +40,7 @@ class MockFacilitatorAPI {
   shouldFailValidation = false
   shouldFailSettle = false
 
-  async verifyPermissions(_: any): Promise<{ isValid: boolean; invalidReason?: string }> {
+  async verifyPermissions(_: any): Promise<VerifyPermissionsResult> {
     this.validationCallCount++
     if (this.shouldFailValidation) {
       return { isValid: false, invalidReason: 'Insufficient credits' }
@@ -44,7 +48,7 @@ class MockFacilitatorAPI {
     return { isValid: true }
   }
 
-  async settlePermissions(_: any): Promise<{ success: boolean; transaction: string; network: string; creditsRedeemed: string }> {
+  async settlePermissions(_: any): Promise<SettlePermissionsResult> {
     this.settleCallCount++
     this.lastSettleAmount = _.maxAmount
     if (this.shouldFailSettle) {
--- a/tests/integration/a2a/payments-server-additional.test.ts
+++ b/tests/integration/a2a/payments-server-additional.test.ts
@@ -6,6 +6,10 @@ import request from 'supertest'
 import { PaymentsA2AServer } from '../../../src/a2a/server.js'
 import type { AgentCard } from '../../../src/a2a/types.js'
 import type { Payments } from '../../../src/payments.js'
+import type {
+  SettlePermissionsResult,
+  VerifyPermissionsResult,
+} from '../../../src/x402/facilitator-api.js'
 import type { AgentExecutor } from '../../../src/a2a/types.js'
 
 function createNoopExecutor(): AgentExecutor {
@@ -63,13 +67,13 @@ describe('PaymentsA2AServer Middleware', () => {
             isValid: true,
             agentRequestId: 'REQ',
             payer: '0x1234567890abcdef',
-          }),
+          } satisfies VerifyPermissionsResult),
         settlePermissions: jest.fn().mockResolvedValue({
           success: true,
           transaction: '0xabcdef1234567890',
           network: 'eip155:84532',
           creditsRedeemed: '1',
-        }),
+        } satisfies SettlePermissionsResult),
       },
     }
 
--- a/tests/integration/a2a/payments-server.test.ts
+++ b/tests/integration/a2a/payments-server.test.ts
@@ -6,6 +6,10 @@ import request from 'supertest'
 import { PaymentsA2AServer } from '../../../src/a2a/server.js'
 import type { AgentCard, AgentExecutor } from '../../../src/a2a/types.js'
 import type { Payments } from '../../../src/payments.js'
+import type {
+  SettlePermissionsResult,
+  VerifyPermissionsResult,
+} from '../../../src/x402/facilitator-api.js'
 import * as utils from '../../../src/utils.js'
 
 const mockDecodeToken = (_token: string) => ({
@@ -66,15 +70,13 @@ describe('PaymentsA2AServer', () => {
       facilitator: {
         verifyPermissions: jest.fn().mockResolvedValue({
           isValid: true,
-        }),
-        settlePermissions: jest
-          .fn()
-          .mockResolvedValue({
-            success: true,
-            transaction: '0x1234567890abcdef',
-            network: 'eip155:84532',
-            creditsRedeemed: '1',
-          }),
+        } satisfies VerifyPermissionsResult),
+        settlePermissions: jest.fn().mockResolvedValue({
+          success: true,
+          transaction: '0x1234567890abcdef',
+          network: 'eip155:84532',
+          creditsRedeemed: '1',
+        } satisfies SettlePermissionsResult),
       },
       agents: {
         getAgentPlans: jest.fn().mockResolvedValue({ plans: [] }),
--- a/tests/integration/a2a/x402-inband-flow.test.ts
+++ b/tests/integration/a2a/x402-inband-flow.test.ts
@@ -22,6 +22,10 @@ import * as utils from '../../../src/utils.js'
 import { encodeAccessToken } from '../../../src/utils.js'
 import type { AgentCard, ExecutionEventBus, PaymentsAgentExecutor } from '../../../src/a2a/types.js'
 import type { Payments } from '../../../src/payments.js'
+import type {
+  SettlePermissionsResult,
+  VerifyPermissionsResult,
+} from '../../../src/x402/facilitator-api.js'
 
 /**
  * A spec-shaped x402 PaymentPayload object (the decoded form of an access
@@ -55,7 +59,7 @@ class MockFacilitatorAPI {
   shouldFailVerify = false
   shouldFailSettle = false
 
-  async verifyPermissions(params: any): Promise<{ isValid: boolean; invalidReason?: string }> {
+  async verifyPermissions(params: any): Promise<VerifyPermissionsResult> {
     this.verifyCallCount++
     this.lastVerifyToken = params.x402AccessToken
     if (this.shouldFailVerify) {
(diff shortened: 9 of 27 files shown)


r-marques:
## Self-review before flipping ready: the gate was measured, not assumed

The claim this PR rests on is that the test tree is typechecked by **nothing**, so a double can drift from the model it impersonates with every check green. I verified both halves of that on this head (`0a6df70e`), because a gate nobody has watched go red proves nothing.

**Method.** Corrupt one settle double — `tests/unit/mcp.test.ts`'s `settlePermissions`, returning `{ totallyNotAField: 'x' }`, i.e. exactly the drift #433 describes — then run each check against the *same* broken tree.

| Check | Broken fixture | Restored |
| --- | --- | --- |
| `pnpm typecheck:tests` | **exit 2** — `error TS2353: Object literal may only specify known properties, and 'totallyNotAField' does not exist in type 'SettlePermissionsResult'` (`mcp.test.ts:83`) | exit 0 |
| `pnpm test:unit` | **635 passed**, 48 suites — fully green | 635 passed |

So the gate catches it and jest cannot see it. That is the premise, reproduced.

**The gate is stronger than the description claims.** I also tried hiding the drift behind `as SettlePermissionsResult` — an escape hatch a future author would reach for. It **also** fails (exit 2): TypeScript refuses an `as` cast between types with no overlap. Worth knowing, because it means the honest fix is the only one available.

## What else I checked

- **Coverage has no hole.** `tests/tsconfig.json` predates this PR and includes `**/*.ts` plus `../src/**/*.ts`, excluding only `coverage` and `logs` — the whole test tree, not a subset.
- **No suppressions were added to make it pass.** Zero `@ts-nocheck`, `@ts-ignore` or `@ts-expect-error` anywhere in the diff. The PR **removes 19** `as any` / `: any`; the `any`s it adds are on double **inputs** (`params: any`), while every **return** is constrained to `SettlePermissionsResult` / `VerifyPermissionsResult`. That is the correct axis: the payload a double hands back is what has to match the model; what it is handed is not what #433 is about.
- **CI placement.** The step runs after Lint, deliberately, so it cannot fail-fast past lint errors that used to surface in the same run.

## Related

Once #441 lands (`PlanBalance.balance` declared as the string the wire really sends), the `jsonSafe` helper in `middleware-hooks-parity` becomes removable — it exists solely to let a fixture carry the bigint that fiction demanded.

Runner: jest via `pnpm test:unit`; gate via `pnpm typecheck:tests` (`tsc --noEmit -p tests/tsconfig.json`). Working tree restored and re-confirmed clean after the mutation.

aaitor (commented):
Reviewed: CI green (lint/build/unit/integration/e2e/CLI+OpenClaw sync). Diff is test-only + CI/docs — no `src/` changes. Return types on the facilitator doubles now match `SettlePermissionsResult`/`VerifyPermissionsResult` from `src/x402/facilitator-api.js`, and `typecheck:tests` wired into `lint_build` as claimed. No blockers found.

Note: this session's PR-review bot tooling has a persistent "sensitive file" permission gate on its run-dir artifact path, so it cannot complete the deterministic verdict step (act.py) right now — leaving this for a human approve or a later tick.

eruizgar91:
👀 Reviewing this now.

CLAUDE.md
```diff
@@ -105,15 +105,43 @@ When writing examples or documentation:
 Before submitting changes, run these checks:
 
 ```bash
-pnpm build && pnpm lint && pnpm test:unit
+pnpm build && pnpm lint && pnpm typecheck:tests && pnpm test:unit
```
eruizgar91, line 108: **LOW** — CONTRIBUTING.md still documents the old pre-PR check, which no longer covers the new CI gate

This PR updated the agent-facing command chain here, but the human-facing one was not updated with it.

[`CONTRIBUTING.md#L8-L13`](https://github.com/nevermined-io/payments/blob/992447aad9262c0653d80baa411f762c5050eca5/CONTRIBUTING.md#L8-L13) still reads:

```bash
pnpm build && pnpm lint && pnpm test:unit   # the pre-PR check
```

That file opens by saying it "covers conventions that CI enforces", and as of this PR `typecheck:tests` is one of them and can fail `lint_build`. A contributor who runs the documented check will now push a green-locally / red-in-CI branch, and the failure will name a file the three commands they ran never looked at.

One-line fix in an adjacent file:

```bash
pnpm build && pnpm lint && pnpm typecheck:tests && pnpm test:unit   # the pre-PR check
```

eruizgar91 (commented):
Test-only change (no `src/` file is touched) that adds `pnpm typecheck:tests`, wires it into CI's `lint_build` job, and fixes the pre-existing test-tree type errors the new gate surfaces.

## The gate genuinely gates — reproduced here, not taken from the description

- **Control.** `tsc --noEmit -p tests/tsconfig.json` at `992447a` → exit **0**. `tsc --showConfig` confirms `strict` / `noImplicitAny` / `strictNullChecks` are all inherited `true` from the root config, so nothing is checked in a relaxed mode.
- **Scope is exactly what is claimed.** `--listFiles` puts **79** files in the program from `tests/`; `find tests -name '*.ts'` also returns 79, and `comm` between the two is empty. Unit, integration, **e2e**, `e2e/helpers` and `setup` are all covered — not a narrower slice than "the test tree". There is no `.tsx` / `.mts` / `.cts` / `.d.ts` under `tests/` that the `**/*.ts` glob could be missing.
- **The violation reproduced, with both existing gates as negative controls.** Appending `const n: number = 'a string'` to `tests/unit/x402-middleware-settlement.test.ts`:
  - `tsc -p tests/tsconfig.json` → **exit 2**, `TS2322: Type 'string' is not assignable to type 'number'`
  - `jest --testPathPattern=tests/unit/x402-middleware-settlement` → **7 passed / 7 total**
  - `tsc -p tsconfig.json` (i.e. `pnpm build`) → **exit 0**

  That is precisely the false green this PR exists to close, and only the new script sees it. The same line in `tests/e2e/fixtures.ts` → exit 2, so the e2e tree is covered too.
- **The doubles guard specifically, not just a generic typecheck.** Dropping `network` and misspelling `creditsRedeemed` in the `satisfies SettlePermissionsResult` fixture at `tests/unit/mcp.test.ts:194` → exit 2 with `TS2561 … Did you mean to write 'creditsRedeemed'?` **and** `TS2345 … Property 'network' is missing`.
- **Model drift propagates.** Adding a required `settledAt: string` to `SettlePermissionsResult` in `src/x402/facilitator-api.ts` turns **21 test files** red (56 errors), matching the PR's own M1 measurement. Independently re-run in a throwaway worktree.
- **It can fail the build.** The step is a plain `run:` in `lint_build` with no `continue-on-error`; `cli_sync_check`, `openclaw_check` and `unit_integration` all declare `needs: lint_build` (and `e2e` needs `unit_integration`). On this head SHA the job's step list shows `8  Typecheck tests  success` — it ran, it was not skipped. `tsc` also fails loudly rather than silently on a bad config path (`TS5058`, exit 1), so there is no silent-pass mode.
- **No emit side effects**: despite the inherited `outDir`/`declaration`, `noEmit` holds — no `dist/` appears and the worktree stays clean across every run.

## Other things checked

- `tests/tsconfig.json` is **byte-identical** in base and head (blob `093fd98`), confirming the PR body's claim that it was already in the repo and referenced by no script and no workflow.
- The MPP doubles are constrained against the **right** models: `MppAPI.verifyCredential` really returns `Promise<VerifyPermissionsResult>` and `settleCredential` returns the discriminated `MppSettleResult` (`src/mpp/mpp-api.ts:184,193`), so `success: true` fixtures are forced to carry `paymentReceipt` and the `success: false` one is correctly allowed to omit it.
- The ~32 server-level `planId` additions cannot change behaviour. `PaywallDecorator` resolves `options?.planId ?? this.config.planId` and throws `Server misconfiguration: missing planId` when both are absent (`src/mcp/core/paywall.ts:98-103`), so any touched test whose per-tool options lacked a `planId` would already have been red on `main`. The one test that does omit it (`mcp.test.ts:403`) already had a server-level `planId` before this PR, and the untouched `tests/unit/mcp/paywall_planid_resolution.test.ts` still pins the missing-planId path.
- Suite parity at head: `tests/unit` → 49 suites, 640 passed, 1 skipped; `tests/integration` → 11 suites, 55 passed. No assertion in the touched files reads `transaction` / `network` off the fixtures that gained them, so the added fields are inert to what is asserted.
- The PR body's "no `as unknown as` / `as any` / `@ts-ignore` / `@ts-expect-error` remains in this branch's diff" holds literally — grepping the `+` lines of the full diff returns only the new `cancelTask(...args: any[])` (which mirrors the pre-existing `execute` signature in the same dummy) and a comment that mentions the phrase.

## Considered and deliberately not raised

- `testing.yml` is `on: [push]`, so the gate never fires on a fork PR. Pre-existing, disclosed, and a trigger change is its own review with its own secrets question.
- `tests/tsconfig.json` overrides `module` / `moduleResolution` to `ESNext` / `Bundler` where the build uses `Node16`, and pulls `../src/**/*.ts` into the program. No strictness is lost (see above); the looser resolver only affects specifier resolution, which `pnpm build` still checks under `Node16` in the same job.
- Long lines in the new fixtures: `prettier --check` and `eslint` both cover `./src` only, and the existing test tree is already full of >100-char lines, so nothing here is out of compliance.
- `release.yml` runs `pnpm build` + tests without `typecheck:tests`, but the published artifact is compiled from `src/` and `testing.yml` runs on the same pushes, so that is not a gate gap.

Four `LOW` findings inline — all one-line fixes, none of them blocking. The substance of the change is sound and the guard does what it says.

CLAUDE.md
```diff
@@ -174,7 +203,7 @@ CI is configured in `.github/workflows/testing.yml` and runs on every push:
 
 | Job | Description | Depends On |
 |-----|-------------|------------|
-| `lint_build` | Install, build, lint | - |
+| `lint_build` | Install, build, typecheck tests, lint | - |
```
eruizgar91: **LOW** — CI table row states a step order that contradicts the deliberate one in testing.yml

This row now reads `Install, build, typecheck tests, lint`, but the job runs **Build -> Lint -> Typecheck tests** (`.github/workflows/testing.yml:32-45`).

That ordering is not incidental — the comment added in this same PR makes it load-bearing:

> Last in the job on purpose: it is the newest and noisiest gate, and running it ahead of Lint would fail-fast past lint errors that used to surface in the same run.

So the table is the one place in the repo that describes the order, and it describes the order the YAML comment explicitly argues against. Someone reconciling the two later has a 50% chance of "fixing" the YAML to match the doc and silently undoing the reasoning.

Swap the last two words: `Install, build, lint, typecheck tests`.

tests/unit/mcp/x402_inband.test.ts
```diff
@@ -59,7 +60,8 @@ function makeDecorator(opts: {
     getEnvironmentName: () => 'staging_sandbox',
     facilitator: {
       settlePermissions: jest.fn(
-        async () => opts.settle ?? { success: true, transaction: '', network: '' },
+        async (): Promise<SettlePermissionsResult> =>
+          opts.settle ?? { success: true, transaction: '', network: '' },
```
eruizgar91, line 64: **LOW** — makeDecorator's default settle keeps the `success: true` + empty `transaction` contradiction this PR de-invented in PaymentsMock

This default is `{ success: true, transaction: '', network: '' }` — the exact pairing commit 5 removed from `PaymentsMock` for a reason that applies here verbatim:

> the model documents an empty `transaction` as the **failure** marker — a response the facilitator never sends, teaching every fixture built from it a contradiction.

`SettlePermissionsResult.transaction` is documented in `src/x402/facilitator-api.ts` as "empty string if settlement failed", and `network: ''` is neither a CAIP-2 id nor a PSP name. The two explicit failure fixtures in this same file (lines 239 and 321) use `success: false` with the empty `transaction`, so they are coherent; only the shared default is not — and it is the one every test that does *not* pass `settle` inherits.

The `satisfies`/annotation work here is correct and the line was rewritten by this PR, so the coherence fix belongs with it. No test in this file reads `transaction` or `network` off the default, so an ordinary successful crypto settle costs nothing:

```ts
opts.settle ?? { success: true, transaction: '0xdefaultsettletx', network: 'eip155:84532' },
```

(That is exactly what the new `PaymentsMock` default became.)

.github/workflows/testing.yml
```diff
@@ -34,6 +34,16 @@ jobs:
...
         run: pnpm lint
 
+      # `pnpm build` compiles src only — tsconfig.json excludes `tests`, and
+      # ts-jest transpiles without checking (the root `isolatedModules: true`),
+      # so nothing else in CI ever typechecks a test file.
+      #
+      # Last in the job on purpose: it is the newest and noisiest gate, and
+      # running it ahead of Lint would fail-fast past lint errors that used to
+      # surface in the same run.
+      - name: Typecheck tests
```
eruizgar91, line 44: **LOW** — The cli/ and openclaw/ test trees stay unchecked, disclosed in the PR body but with no tracking issue

Verified, and the PR body's account is accurate: [`cli/tsconfig.json#L24`](https://github.com/nevermined-io/payments/blob/992447aad9262c0653d80baa411f762c5050eca5/cli/tsconfig.json#L22-L25) excludes `test` and [`openclaw/tsconfig.json#L20`](https://github.com/nevermined-io/payments/blob/992447aad9262c0653d80baa411f762c5050eca5/openclaw/tsconfig.json#L18-L21) excludes `tests`; neither package has a test-specific tsconfig or a `typecheck` script, and both have real suites (`cli/test/{unit,integration,helpers,__mocks__}`, `openclaw/tests/plugin.test.ts`). `cli_sync_check` and `openclaw_check` run `pnpm test` there but nothing typechecks those files, so the drift class this PR closes for `tests/` can still land in them.

Not a change for this PR — it needs a tsconfig per package plus whatever errors that surfaces, which is a design decision and an unbounded diff. **But it currently has no home:** the analysis lives only in this PR description, which the squash-merge turns into a commit message nobody greps. Please open a follow-up issue and link it from the comment at `.github/workflows/testing.yml:37-43`, so the next person who reads "typecheck the test tree" learns which test tree from the repo rather than from git archaeology.

CLAUDE.md
r-marques: Folded in `d691823`. You are right that the human-facing chain is the one that matters here — the agent-facing one in CLAUDE.md is read by something that will be told about the new gate anyway; `CONTRIBUTING.md` is read by someone who then pushes.

```bash
pnpm build && pnpm lint && pnpm typecheck:tests && pnpm test:unit   # the pre-PR check
```

And I ran the whole chain end to end rather than just editing the string, since a pre-PR check that does not pass on a clean tree is worse than one that is out of date: build 0, lint 0, typecheck:tests 0, 640 unit tests. Green.

The detail that makes this more than tidiness is the one you put last: the CI failure would name **a file none of those three commands ever looked at**. That is the confusing shape — not "you forgot to run something", but "the thing you ran could not have seen this".

CLAUDE.md
r-marques: Fixed in `d691823` — and this is the better of the two findings, because the failure mode is a later contributor undoing the reasoning while believing they are fixing an inconsistency.

The row now reads `Install, build, lint, typecheck tests`, matching `testing.yml:28/31/34/44`.

I also put the **reason** in the row rather than only the order, and an explicit instruction about which side to change:

> Typecheck tests runs LAST on purpose — see the comment in `testing.yml`: it is the newest and noisiest gate, and running it before Lint would fail-fast past lint errors that used to surface in the same run. **Do not reorder to match a doc; fix the doc.**

Your 50% framing is exactly why the bare correction was not enough. A doc and a YAML that disagree is a coin flip for whoever reconciles them, and the YAML comment alone does not travel to someone reading the table. Now both ends say the same thing and name each other.

package.json
```diff
@@ -44,6 +44,7 @@
   "scripts": {
     "clean": "rm -rf ./dist/ ./doc/ ./.nyc_output",
     "build": "tsc",
+    "typecheck:tests": "tsc --noEmit -p tests/tsconfig.json",
```
eruizgar91, line 47: **MEDIUM** — `tests/utils.ts` carries `// @ts-nocheck` (line 5), which exempts the whole file from `typecheck:tests` — and the suppression is stale, so this is dead weight blinding the very gate this PR adds.

Reproduced directly against this PR's head:

1. With `@ts-nocheck` present, appending `export const x: number = "not a number"` to `tests/utils.ts` → `pnpm typecheck:tests` exits **0**, zero errors reported.
2. Deleting the `@ts-nocheck` (and its `eslint-disable` line) with the file otherwise unchanged → `pnpm typecheck:tests` exits **0** with zero errors — the file already typechecks clean without the suppression, so the comment's justification ("rootDir restriction doesn't apply to test utilities importing from src") is stale, most likely superseded by `tests/tsconfig.json`'s own `rootDir: ".."` override.
3. With the suppression removed and the same planted error reintroduced → correctly caught: `TS2322: Type 'string' is not assignable to type 'number'`.

`tests/utils.ts` isn't a throwaway file: it backs `retryWithBackoff` / `waitForCondition` / `makeWaitForAgent` / `makeWaitForPlan`, imported by ~11 files under `tests/e2e/` including `tests/e2e/helpers/a2a-setup-helpers.ts`. A future type-level drift there — exactly the class of bug this PR exists to catch — would go unseen by the new gate for as long as the suppression stays.

Given it's already verified to be a no-op today, this is a two-line deletion (permalink: https://github.com/nevermined-io/payments/blob/d691823da1209d7abf63f40f6c6e10208dcc5d88/tests/utils.ts#L4-L5) with no design decision left open — worth landing in this PR rather than a follow-up.

eruizgar91 (commented):
Follow-up from deeper verification after my first pass (thanks for the quick turnaround on the two doc findings — both confirmed fixed here). One more finding below, independently reproduced twice.

package.json
r-marques: Folded in `fa0f3c6`. I reproduced all three of your states independently before deleting anything, because "this suppression is a no-op" is exactly the claim that should not be taken on trust:

| state | result |
| --- | --- |
| `@ts-nocheck` present + planted `const x: number = "not a number"` | exit **0** — blind |
| suppression removed, file otherwise unchanged | exit **0** — already a no-op |
| suppression removed + the same planted error | exit **2**, `TS2322` at `tests/utils.ts(178,14)` |

Identical to yours. The stated justification is stale, almost certainly superseded by `tests/tsconfig.json`'s own `rootDir: ".."`.

**I also swept the whole test tree rather than fixing only the file you named**, since the general form of your finding is that a gate's reach is its glob *minus every suppression already inside it*. Two other hits, neither a problem, and the distinction is worth recording:

- `tests/unit/x402-visa.test.ts:308` — an `@ts-expect-error` asserting a deliberately removed `scheme:'visa'`. **Self-policing**: it fails if the error stops occurring, so it cannot rot into a blind spot.
- one prose mention inside a comment, not a directive.

So `@ts-nocheck` and `@ts-ignore` are the silent ones; `@ts-expect-error` is not. Only the file you found was actually blinding anything.

Your point that this is not a throwaway file is what makes it more than hygiene — `retryWithBackoff` / `waitForCondition` / `makeWaitForAgent` / `makeWaitForPlan` are imported by ~11 files under `tests/e2e/`, so a type-level drift there is precisely the bug class this PR exists to catch, and it would have been invisible to the gate for as long as the comment stayed. A gate that ships with a hole in the middle of its most-imported helper is worse than no gate, because it reads as coverage.

Full chain green: build 0, lint 0, typecheck:tests 0, 640 unit tests.

aaitor (commented):
Reviewed anyway despite the bot tooling blocker (see Discord) — this session's PR-review bot cannot write `findings.json` under `~/.claude/pr-reviews/runs/` ("sensitive file" permission gate), so it cannot run the deterministic verdict step.

Read the full diff: the `typecheck:tests` CI gate (placed last in `lint_build`, after Lint, on purpose), the `tests/tsconfig.json` addition, and the sweep constraining every settle/verify test double to `SettlePermissionsResult`/`VerifyPermissionsResult` imported from `src/` rather than an ad hoc inline shape. Confirmed the `@ts-nocheck` on `tests/utils.ts` — the suppression that was blinding this very gate on a file ~11 e2e specs import — is deleted, and that raw `fetch`/`Response` stubs are deliberately left untyped where a spec needs to post a malformed body. CI green.

No blockers found.

aaitor (commented):
Reviewed independently (fresh context): all 8 commits read end-to-end, including the final `@ts-nocheck` removal on `tests/utils.ts` — verified against the three reproduced states in the commit message (silently blind with the suppression + a planted type error, a no-op without it, TS2322 caught once removed). `typecheck:tests` CI wiring, doc updates, and the model-constrained test doubles all check out. No unresolved threads, CI green. No blockers found.

aaitor (approved):
✅ **Approved** — no blockers, nothing to address.

Inline review: https://github.com/nevermined-io/payments/pull/437#pullrequestreview-5189868090

PR #176, opened 2025-12-22
feat: deprecation of access token in favour of x402 access token

eruizgar91:
## Description

Add a description of your changes here.

## Is this PR related with an open issue?

Related to Issue #

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] Follows the code style of this project.
- [ ] Tests Cover Changes
- [ ] Documentation

#### Funny gif

Put a link of a funny gif inside the parenthesis-->

PR #439, opened 2026-09-10
feat(a2a): report the credits actually redeemed, per billing model (#438)

### LINKED ISSUE #438: executeRedemption returns `Promise<any>` and casts, hiding two undeclared field reads on the A2A redemption path
Found while fixing #433 (`PaymentsMock` takes `any`, so untyped test doubles drift unchecked). This is the **same defect in production code**, and it is why nobody has ever been forced to reconcile two undeclared fields on a money-adjacent path.

## What is wrong

`src/a2a/paymentsRequestHandler.ts:250`:

```ts
private async executeRedemption(…): Promise<any> {
```

Two call sites then read fields off that `any` which no type declares, and one casts the result to a type it was never checked against:

```ts
// :556 and :870
event.metadata.txHash = response.txHash ?? response.transaction
event.metadata.creditsCharged = response.amountOfCredits ? Number(response.amountOfCredits) : …

// :865
settlement = response as SettlePermissionsResult
```

## What I verified

- **`txHash` is not on the settle response.** In nvm-monorepo `origin/main`, `apps/api/src/x402/dto/x402-responses.dto.ts` declares `transaction?: string` and no `txHash`. Every `txHash` in the API's x402 tree is an internal `Purchase` DB column, not a wire field.
- **`amountOfCredits` is on no type in this SDK at all** — it appears in exactly four lines, all in this one file.
- **`SettlePermissionsResult` declares neither.** So the cast at `:865` asserts a shape the value has not been checked against.

## Why the fix is "type the return", not "pick one"

It is tempting to ask *"should the model declare `txHash`, or should the fallback be removed?"* — but that is a choice made without evidence, and it is avoidable. **Type `executeRedemption`'s return and the compiler forces the reconciliation**: either the real shape carries those fields, in which case the model is incomplete and should declare them; or it does not, in which case the reads are dead and go. The answer is determinate once the `any` is gone.

So:

1. Give `executeRedemption` its real return type (derive it from what it actually returns — do not guess, and do not widen the model to make the reads compile).
2. Remove the `as SettlePermissionsResult` cast at `:865`, or make it unnecessary.
3. Resolve `txHash` / `amountOfCredits` per whatever step 1 reveals, and say in the PR which of the two cases held.

⚠️ **`event.metadata.txHash` is what records the transaction id into A2A event metadata**, which is what a buyer is shown for payment support. Do not remove the fallback without establishing that nothing on the wire supplies `txHash` — a defensive read that turns out to be load-bearing is a worse outcome than an untyped one.

## Note on scope

`tests/` in this repo is typechecked by nothing today — `tsconfig.json` excludes it and ts-jest transpiles without checking (`isolatedModules: true`). #433 / PR #437 adds `pnpm typecheck:tests` and wires it into CI. That gate covers the test tree; **`src/` is already covered by `pnpm build`**, so this defect is visible to `tsc` the moment the `any` is replaced.

## Related

- #433 / PR #437 — the same defect class in the test doubles
- nevermined-io/payments-py#273 — the Python sibling

r-marques:
Closes #438.

> ### ⚠️ Read this first: buyer-facing behaviour change
>
> `event.metadata.creditsCharged` used to report the credits a request **asked** to burn. It now reports what the settle **actually redeemed** — and on a plan with no credit balance it is **omitted entirely** rather than set to a number that means nothing.
>
> This PR was scoped as a typing fix (#438). The behaviour change was added on the repo owner's ruling after the typing work exposed the question; it is not incidental, and it is the part to review hardest.

## What a consumer sees, before and after

`event.metadata.creditsCharged`, on the A2A status-update event and the persisted task:

| billing model | before | after |
| --- | --- | --- |
| `credits` | the credits **requested** (`creditsUsed`) | the credits **redeemed** (`creditsRedeemed`) — differs under margin-based pricing |
| `credits`, nothing redeemed | the credits requested | `0` |
| `pay-as-you-go` | the credits requested — a figure the plan has no basis for | **key absent** |
| `billingModel` absent (older backend) | the credits requested | the credits redeemed; a missing discriminator is read as `credits`, never as pay-as-you-go |
| no `creditsRedeemed` at all (older still) | the credits requested | **key absent** |

**Nothing is lost by the omissions.** The credits a request asked to burn remain on the same event as `creditsUsed`, untouched, and a pay-as-you-go charge is referenced by `orderTx` (fiat) or `transaction` (crypto). A consumer that genuinely wants the requested figure reads `creditsUsed`, which is what it always meant.

### Why pay-as-you-go is absent rather than `0`

On a plan with no credit balance, `creditsRedeemed` is the string `'0'` **even on a settle that charged the buyer** — that is the published contract (#432). Reporting that `0` as `creditsCharged` would tell a buyer they were charged nothing for a payment that really happened. There is no honest number to report, so nothing is reported; a consumer can then distinguish "not applicable on this plan" from a credits plan that genuinely redeemed nothing, which reports `0`.

Mind the types: these are **strings**. `'0'` is truthy while `Number('0') > 0` is false, so two plausible-looking checks disagree on exactly the value that matters.

The rule lives in one helper, `resolveCreditsCharged`, used by both call sites (`handleTaskFinalization` and `processStreamingEventsWithFinalization`) so the two copies cannot drift.

---

## The typing fix (#438 as filed)

`executeRedemption`'s real return type was never a judgement call. Its only non-throwing exit is `return await this.paymentsService.facilitator.settlePermissions(...)`, already declared `Promise<SettlePermissionsResult>` — so naming it is a pure narrowing of the `any`, not a new claim about the wire. With the `any` gone, `tsc` settles both undeclared reads and both casts. **Both fields were dead, so the reads go and the model is untouched.**

### `txHash` — nothing on the wire supplies it

Checked, rather than inferred from the SDK's own types:

1. The Nevermined API's settle response is `X402SettleResponseDto` (nvm-monorepo `origin/main` @ `fef4e854c`, `apps/api/src/x402/dto/x402-responses.dto.ts`). It declares `transaction?: string`. There is no `txHash`.
2. The whole settle path is typed as that DTO end to end — `x402.controller.ts` → `x402.service.ts` → both scheme handlers — and every settle `return { … }` is an object literal, so excess-property checking forbids an undeclared `txHash` reaching the wire. The controller returns the service result verbatim.
3. Every `txHash` in the API's x402 tree is an internal `Purchase` / delegation DB column.
4. `amountOfCredits` occurs in the entire API only as a parameter name in `observability.service.ts` and `protocol.service.ts` — no HTTP surface returns it.

The A2A metadata **key** `txHash` is unchanged — it is what a buyer is shown for payment support, and it still carries the transaction id, now read from the wire field that actually holds it.

### Why the fallback looked load-bearing and was not

It was inherited, not defensive:

| commit | date | what happened |
| --- | --- | --- |
| `5ce1981` | 2025-10-09 | reads written against `payments.requests.redeemCreditsFromRequest`, whose response really did carry `txHash` and `amountOfCredits` |
| `a204d9a` | 2026-01-08 (#176) | call switched to `facilitator.settlePermissions`; both `redeemCreditsFrom*` methods deleted — the reads start returning `undefined` |
| `e8ec566` | 2026-06-17 (#392) | `?? response.transaction` added **behind** the already-dead `txHash` instead of replacing it |

That last step is what made a dead read look load-bearing, and nobody could see it because the return was `any`.

### The loop closes on #433

**The only thing that ever supplied `txHash` / `amountOfCredits` on this path was its own test double** — `{ txHash: '0xabc', amountOfCredits: 5n }`, a shape `settlePermissions` cannot return. So #433's defect class (untyped fixtures drifting from the model) is the direct cause of #438's: an untyped double kept two dead production reads alive through two refactors, and the assertions pinned the drift rather than the behaviour.

The doubles now return the real wire shape. This touches `tests/unit/a2a/payments-request-handler.test.ts`, which is **not** among the files PR #437 (#433) changes; the streaming coverage lives there too, rather than in `payments-request-handler-stream.test.ts`, for the same reason.

## Verification

| gate | result | baseline on `main` |
| --- | --- | --- |
| `pnpm build` (`tsc`, covers `src/`) | exit 0 | exit 0 |
| `pnpm test:unit` | 48 suites / **643** passed / 1 skipped | 48 / 635 / 1 |
| `pnpm test:integration` | 11 suites / 55 passed | 11 / 55 |
| `pnpm lint` | 0 errors, 3 warnings | 0 errors, 3 warnings |

Note `pnpm lint` runs `eslint ./src` — it does not cover `tests/`.

### Mutation-tested, against a passing control

Control: `pnpm build` exit 0; `payments-request-handler.test.ts` 16 passed.

**The typing:**

| mutant | result |
| --- | --- |
| reintroduce `response.txHash ?? response.transaction` | `pnpm build` **red** — `paymentsRequestHandler.ts(562,50): error TS2339: Property 'txHash' does not exist on type 'SettlePermissionsResult'` |
| reintroduce `response.amountOfCredits` | `pnpm build` **red** — TS2339 `'amountOfCredits'` |

**The billing-model rule** — every mutant names the tests that died:

| mutant | tests killed |
| --- | --- |
| drop the pay-as-you-go guard (report the `'0'`) | 2 — both PAYG tests, plain and streaming |
| read a missing discriminator as pay-as-you-go | 1 — absent-`billingModel` |
| truthiness (`creditsRedeemed ? …`) instead of numeric | 1 — unparseable-figure |
| drop the `Number.isFinite` guard (publish `NaN`) | 2 — unparseable-figure, no-usable-figure |
| a `redeemed > 0` check | 1 — credits-plan-redeemed-nothing |
| streaming site reports the requested burn | 2 |
| non-streaming site reports the requested burn | 8 |

Two of those tests exist **because** of the mutation round: the pay-as-you-go guard returns before the numeric read, so it masked the `'0'`-vs-numeric trap entirely and the truthiness mutant survived the first pass. Two overlapping guards each masked the other.

### DIFF
--- a/markdown/a2a-integration.md
+++ b/markdown/a2a-integration.md
@@ -243,13 +243,25 @@ eventBus.publish({
 ### What the SDK adds to `metadata` after settlement
 
 Your executor sets `creditsUsed`. After the request settles, the SDK adds two more
-keys to the same `metadata` object, on the status-update event and on the stored
-task:
+keys to the same `metadata` object:
 
 | key | meaning |
 | --- | --- |
-| `txHash` | The settlement transaction, from the facilitator's receipt. |
+| `txHash` | The settlement transaction. A chain tx hash on the crypto rails; on fiat it is the **PSP transaction id** (a Stripe PaymentIntent, `pi_...`), so do not assume it is a hash. |
 | `creditsCharged` | The credits the settle **actually redeemed**. |
+
+⚠️ **Where they land depends on the transport, so do not assume `task.metadata`.**
+
+| transport | status-update event | stored task (`task.metadata`) |
+| --- | --- | --- |
+| non-streaming | ✅ both keys | ✅ both keys |
+| streaming | ❌ neither — the event is `yield`ed before settlement | ❌ neither |
+| batch redemption (`useBatch`) | ❌ neither | ❌ neither |
+
+On the streaming transport the settlement block runs *after* the event has been
+yielded, and only `event.metadata` is touched — `task.metadata` has a single
+writer, on the non-streaming path. Under `useBatch` the block is skipped
+entirely, on both transports.
 
 **`creditsCharged` is not always present, and its absence is meaningful.** It
 reports what was redeemed, which is not necessarily what you asked to burn —
@@ -267,13 +279,20 @@ Pay-as-you-go is the case to understand. Such a plan holds no credit balance, so
 `creditsRedeemed` is the string `'0'` **even on a settle that charged the buyer**.
 Publishing that as `creditsCharged: 0` would tell a buyer they were charged
 nothing for a payment that really happened, so nothing is published instead. The
-charge is referenced by `orderTx` (fiat) or `transaction` (crypto), and the full
-settlement receipt is on the task under `x402.payment.receipts`.
+charge is referenced by `orderTx` (fiat) or `transaction` (crypto).
+
+The full settlement receipt is also available, but **only on the in-band x402
+transport** and at `task.status.message.metadata['x402.payment.receipts']` —
+*not* `task.metadata`. On the legacy `payment-signature` header path there are no
+receipts at all, and under `useBatch` a deferred marker is written instead of
+one.
 
 **Read it with `in`, not truthiness** — `0` is a legitimate value:
 
 ```typescript
-if ('creditsCharged' in task.metadata) {
+// `Task.metadata` is optional in @a2a-js/sdk, and `'x' in undefined` throws —
+// so guard it, or the else-branch below is unreachable exactly when you need it.
+if (task.metadata && 'creditsCharged' in task.metadata) {
   console.log(`Redeemed ${task.metadata.creditsCharged} credits`)
 } else {
   // Pay-as-you-go, or no usable figure. `creditsUsed` still tells you what the
--- a/markdown/a2a-integration.md
+++ b/markdown/a2a-integration.md
@@ -240,6 +240,50 @@ eventBus.publish({
 })
 ```
 
+### What the SDK adds to `metadata` after settlement
+
+Your executor sets `creditsUsed`. After the request settles, the SDK adds two more
+keys to the same `metadata` object, on the status-update event and on the stored
+task:
+
+| key | meaning |
+| --- | --- |
+| `txHash` | The settlement transaction, from the facilitator's receipt. |
+| `creditsCharged` | The credits the settle **actually redeemed**. |
+
+**`creditsCharged` is not always present, and its absence is meaningful.** It
+reports what was redeemed, which is not necessarily what you asked to burn —
+under margin-based pricing the two differ. The SDK omits the key entirely rather
+than report a figure that would be false:
+
+| situation | `creditsCharged` |
+| --- | --- |
+| credits plan | the credits redeemed |
+| credits plan, nothing redeemed | `0` — a real figure, not an absence |
+| **pay-as-you-go** | **key absent** |
+| the backend reported no usable figure | **key absent** (and the SDK warns) |
+
+Pay-as-you-go is the case to understand. Such a plan holds no credit balance, so
+`creditsRedeemed` is the string `'0'` **even on a settle that charged the buyer**.
+Publishing that as `creditsCharged: 0` would tell a buyer they were charged
+nothing for a payment that really happened, so nothing is published instead. The
+charge is referenced by `orderTx` (fiat) or `transaction` (crypto), and the full
+settlement receipt is on the task under `x402.payment.receipts`.
+
+**Read it with `in`, not truthiness** — `0` is a legitimate value:
+
+```typescript
+if ('creditsCharged' in task.metadata) {
+  console.log(`Redeemed ${task.metadata.creditsCharged} credits`)
+} else {
+  // Pay-as-you-go, or no usable figure. `creditsUsed` still tells you what the
+  // request asked to burn.
+}
+```
+
+The credits your request asked to burn remain on the event as `creditsUsed`,
+untouched, so nothing is lost by the omission.
+
 ### Message Event (Streaming)
 
 ```typescript
--- a/src/a2a/paymentsRequestHandler.ts
+++ b/src/a2a/paymentsRequestHandler.ts
@@ -57,6 +57,13 @@ const terminalStates: TaskState[] = ['completed', 'failed', 'canceled', 'rejecte
  *   referenced by `orderTx` (fiat) or `transaction` (crypto); the credits the
  *   request asked to burn remain on the event as `creditsUsed`, so nothing is
  *   lost by omitting a figure that has no meaning on this billing model.
+ *
+ *   ⚠️ MCP does the OPPOSITE and that is deliberate — a settled decision (repo
+ *   owner, on the #443 review), not drift. `_meta['nevermined/credits']` reports
+ *   the `'0'` because it emits `billingModel` BESIDE it, so a consumer can read
+ *   it; `creditsCharged` here has no discriminator next to it, so a bare `0`
+ *   would say "you were charged nothing" about a charge that succeeded. Do not
+ *   reconcile the two.
  * - absent — a Nevermined API older than the discriminator (`creditsRedeemed`
  *   has been on the settle response since 2026-03-25, `billingModel` only since
  *   2026-08-06). Apply the `credits` rule: a missing discriminator must never be
--- a/src/a2a/paymentsRequestHandler.ts
+++ b/src/a2a/paymentsRequestHandler.ts
@@ -41,6 +41,42 @@ import type {
 const terminalStates: TaskState[] = ['completed', 'failed', 'canceled', 'rejected']
 
 /**
+ * The credits a settle actually redeemed, for `event.metadata.creditsCharged`.
+ *
+ * Reads {@link SettlePermissionsResult.creditsRedeemed} per billing model rather
+ * than trusting it flat, because it does not mean the same thing under both:
+ *
+ * - `credits` — the plan holds a balance and `creditsRedeemed` is what came out
+ *   of it. Report it. It can legitimately differ from the credits the request
+ *   asked to burn (margin-based pricing), which is the whole reason to report
+ *   the settled figure rather than the requested one.
+ * - `pay-as-you-go` — the plan holds no credit balance at all, so
+ *   `creditsRedeemed` is the string `'0'` *even on a settle that charged the
+ *   buyer*. Returning that `0` would tell a buyer they were charged nothing for
+ *   a payment that really happened, so report nothing instead. The charge is
+ *   referenced by `orderTx` (fiat) or `transaction` (crypto); the credits the
+ *   request asked to burn remain on the event as `creditsUsed`, so nothing is
+ *   lost by omitting a figure that has no meaning on this billing model.
+ * - absent — a Nevermined API older than the discriminator (`creditsRedeemed`
+ *   has been on the settle response since 2026-03-25, `billingModel` only since
+ *   2026-08-06). Apply the `credits` rule: a missing discriminator must never be
+ *   read as pay-as-you-go.
+ *
+ * Mind the type: these are **strings**. `'0'` is truthy while `Number('0') > 0`
+ * is false, so a truthiness check and a numeric one disagree on exactly the
+ * value that matters.
+ *
+ * @param settlement - The settlement receipt as the facilitator returned it
+ * @returns The credits redeemed, or `undefined` when the billing model has no
+ *   credits to report or the backend reported no usable figure
+ */
+function resolveCreditsCharged(settlement: SettlePermissionsResult): number | undefined {
+  if (settlement.billingModel === 'pay-as-you-go') return undefined
+  const redeemed = Number(settlement.creditsRedeemed)
+  return Number.isFinite(redeemed) ? redeemed : undefined
+}
+
+/**
  * Options for configuring the PaymentsRequestHandler
  */
 export interface PaymentsRequestHandlerOptions {
@@ -560,7 +596,13 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
               // There is no `txHash` on the wire — see handleTaskFinalization.
               if (response && event.metadata) {
                 event.metadata.txHash = response.transaction
-                event.metadata.creditsCharged = event.metadata.creditsUsed
+                // Left unset when the billing model has no credits to report —
+                // see resolveCreditsCharged. `creditsUsed` still carries what
+                // the request asked to burn.
+                const creditsCharged = resolveCreditsCharged(response)
+                if (creditsCharged !== undefined) {
+                  event.metadata.creditsCharged = creditsCharged
+                }
               }
 
               // x402 v2 A2A transport: stamp the settlement receipt onto the
@@ -867,16 +909,14 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
           // Update event metadata with redemption results. `txHash` is the A2A
           // metadata key, not a wire field: the settle response reports the
           // transaction id as `transaction`.
+          const creditsCharged = resolveCreditsCharged(response)
           event.metadata = {
             ...event.metadata,
             txHash: response.transaction,
-            // The credits this request asked to burn. The settle response does
-            // report what was actually redeemed, as `creditsRedeemed`, but
-            // reading it here would change behaviour and is not free: it is the
-            // string '0' on `billingModel: 'pay-as-you-go'` plans even when the
-            // buyer WAS charged, so a truthiness read reports 0 on a real
-            // charge. See SettlePermissionsResult before wiring it up.
-            creditsCharged: creditsToBurn,
+            // The credits actually redeemed, omitted entirely when the billing
+            // model has none to report — see resolveCreditsCharged. The credits
+            // this request asked to burn stay available as `creditsUsed`.
+            ...(creditsCharged !== undefined && { creditsCharged }),
           }
         }
       } catch (err) {
--- a/src/a2a/paymentsRequestHandler.ts
+++ b/src/a2a/paymentsRequestHandler.ts
@@ -104,17 +104,40 @@ function resolveCreditsCharged(settlement: SettlePermissionsResult): number | un
     return undefined
   }
 
+  // A FAILED settle has no measured redemption to report, so it gets no figure.
+  // `settlePermissions` does not throw on one — it returns 200 with
+  // `success: false` verbatim — so without this a failure published
+  // `creditsCharged` as though it were measured: `{success: false,
+  // creditsRedeemed: '7'}` reported 7 credits charged for a settle that charged
+  // nothing. The published contract makes that a lie rather than merely noise,
+  // because `markdown/a2a-integration.md` defines `0` as "a real figure, not an
+  // absence" — a number here asserts the settle completed.
+  //
+  // ⚠️ `!== true`, not `=== false`, and the asymmetry with `billingModel` above
+  // is deliberate. An ABSENT `billingModel` is a documented legacy shape (APIs
+  // predating the field) and is read as `credits`; an absent `success` is not a
+  // legacy shape — the field has always been there and is declared
+  // non-optional — so its absence means the response is malformed, and a
+  // malformed response is exactly when not to publish a figure.
+  //
+  // Silent, like the pay-as-you-go branch: a failed settle is a legitimate
+  // outcome the caller already sees on `success`, not a reporting fault.
+  if (settlement.success !== true) {
+    return undefined
+  }
+
   const raw = settlement.creditsRedeemed
   if (typeof raw !== 'string' || !DECIMAL_INTEGER_STRING.test(raw)) {
     // Silence here would be indistinguishable from the pay-as-you-go omission,
     // which is a deliberate design decision rather than a fault. Warn, as
     // `x402/express/middleware.ts` does on this same field.
-    if (raw !== undefined) {
-      console.warn(
-        `[PaymentsRequestHandler] settle reported an unusable creditsRedeemed ` +
+    console.warn(
+      raw === undefined
+        ? `[PaymentsRequestHandler] settle reported no creditsRedeemed; omitting ` +
+          `creditsCharged rather than publishing a figure`
+        : `[PaymentsRequestHandler] settle reported an unusable creditsRedeemed ` +
           `(${JSON.stringify(raw)}); omitting creditsCharged rather than publishing a figure`,
-      )
-    }
+    )
     return undefined
   }
 
--- a/src/a2a/paymentsRequestHandler.ts
+++ b/src/a2a/paymentsRequestHandler.ts
@@ -66,14 +66,73 @@ const terminalStates: TaskState[] = ['completed', 'failed', 'canceled', 'rejecte
  * is false, so a truthiness check and a numeric one disagree on exactly the
  * value that matters.
  *
+ * Three ways this returns `undefined`, and they mean different things — the
+ * first is a design decision, the other two are faults, which is why only the
+ * faults warn:
+ *
+ * 1. the billing model has no credits figure to report (pay-as-you-go, or any
+ *    discriminator that is present and is not `credits`) — silent, by design;
+ * 2. `creditsRedeemed` is absent, or not a plain decimal string — warns;
+ * 3. the value is a decimal string too large for a JS number — warns.
+ *
  * @param settlement - The settlement receipt as the facilitator returned it
- * @returns The credits redeemed, or `undefined` when the billing model has no
- *   credits to report or the backend reported no usable figure
+ * @returns The credits redeemed, or `undefined` per the three cases above
  */
+/**
+ * A decimal string, and nothing else — no sign, no whitespace, no `0x`/`1e3`,
+ * no fractional part. `Number()` accepts every one of those, and worse, maps the
+ * empty-ish family (`''`, `'   '`, `null`, `[]`, `false`) to **0** rather than
+ * `NaN`, so `Number.isFinite` does not catch them. Publishing that 0 would say
+ * "you were charged nothing" on a settle we simply could not read — the same
+ * class of lie this helper exists to prevent on pay-as-you-go.
+ *
+ * Same reasoning, same shape as `DECIMAL_INTEGER_STRING` in `mpp/mpp-api.ts`.
+ */
+const DECIMAL_INTEGER_STRING = /^\d+$/
+
 function resolveCreditsCharged(settlement: SettlePermissionsResult): number | undefined {
-  if (settlement.billingModel === 'pay-as-you-go') return undefined
-  const redeemed = Number(settlement.creditsRedeemed)
-  return Number.isFinite(redeemed) ? redeemed : undefined
+  // ALLOWLIST, not a denylist. `=== 'pay-as-you-go'` let every other value fall
+  // into the credits branch and publish a `0`: `null`, `''`, `'PAY-AS-YOU-GO'`,
+  // a future third billing model, a non-string. The response is an unchecked
+  // `as SettlePermissionsResult` over `response.json()`, so the union type
+  // provides no runtime guarantee and the wrong values are reachable.
+  //
+  // `undefined` still means `credits` — that is the documented ruling above and
+  // it is preserved exactly. What changes is that a discriminator which is
+  // present but not `credits` now omits rather than reporting a zero.
+  if (settlement.billingModel !== undefined && settlement.billingModel !== 'credits') {
+    return undefined
+  }
+
+  const raw = settlement.creditsRedeemed
+  if (typeof raw !== 'string' || !DECIMAL_INTEGER_STRING.test(raw)) {
+    // Silence here would be indistinguishable from the pay-as-you-go omission,
+    // which is a deliberate design decision rather than a fault. Warn, as
+    // `x402/express/middleware.ts` does on this same field.
+    if (raw !== undefined) {
+      console.warn(
+        `[PaymentsRequestHandler] settle reported an unusable creditsRedeemed ` +
+          `(${JSON.stringify(raw)}); omitting creditsCharged rather than publishing a figure`,
+      )
+    }
+    return undefined
+  }
+
+  const redeemed = Number(raw)
+  if (!Number.isSafeInteger(redeemed)) {
+    // The wire carries this as a string precisely because it can exceed what a
+    // JS number represents. Above 2^53 the conversion is silently wrong
+    // (`'9007199254740993'` becomes 9007199254740992), and a wrong number is
+    // worse than no number: the exact value is still on the task, verbatim, in
+    // the `x402.payment.receipts` entry that `recordPaymentSuccess` writes.
+    console.warn(
+      `[PaymentsRequestHandler] creditsRedeemed ${raw} exceeds Number.MAX_SAFE_INTEGER; ` +
+        `omitting creditsCharged rather than publishing a rounded figure`,
+    )
+    return undefined
+  }
+
+  return redeemed
 }
 
 /**
--- a/src/a2a/paymentsRequestHandler.ts
+++ b/src/a2a/paymentsRequestHandler.ts
@@ -245,13 +245,17 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
    * @param bearerToken - The bearer token for authentication
    * @param creditsUsed - The number of credits to burn
    * @param httpContext - Optional HTTP context with endpoint and method information
-   * @returns Promise resolving to the redemption result
+   * @returns The settlement receipt, verbatim as the facilitator returned it. A
+   *   settle the backend accepted but could not complete comes back as
+   *   `success: false` with an `errorReason` rather than throwing — see
+   *   {@link SettlePermissionsResult} for which fields evidence a charge under
+   *   which billing model.
    */
   private async executeRedemption(
     bearerToken: string,
     creditsUsed: bigint | number,
     httpContext?: HttpRequestContext,
-  ): Promise<any> {
+  ): Promise<SettlePermissionsResult> {
     const decodedAccessToken = decodeAccessToken(bearerToken)
     if (!decodedAccessToken) {
       throw PaymentsError.unauthorized('Invalid access token.')
@@ -551,12 +555,12 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
                 httpContext,
               )
 
-              // Update event metadata with response data
+              // `txHash` is the A2A metadata key a buyer is shown for payment
+              // support; the settle response reports that id as `transaction`.
+              // There is no `txHash` on the wire — see handleTaskFinalization.
               if (response && event.metadata) {
-                event.metadata.txHash = response.txHash ?? response.transaction
-                event.metadata.creditsCharged = response.amountOfCredits
-                  ? Number(response.amountOfCredits)
-                  : event.metadata.creditsUsed
+                event.metadata.txHash = response.transaction
+                event.metadata.creditsCharged = event.metadata.creditsUsed
               }
 
               // x402 v2 A2A transport: stamp the settlement receipt onto the
@@ -567,11 +571,7 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
               if (httpContext?.inBand) {
                 const task = resultManager.getCurrentTask()
                 if (task) {
-                  this.recordInBandSettlement(
-                    task,
-                    httpContext,
-                    response as SettlePermissionsResult,
-                  )
+                  this.recordInBandSettlement(task, httpContext, response)
                   await resultManager.processEvent(task)
                 }
               }
@@ -862,16 +862,21 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
             BigInt(creditsToBurn),
             httpContext,
           )
-          settlement = response as SettlePermissionsResult
+          settlement = response
 
-          // Update event metadata with redemption results
+          // Update event metadata with redemption results. `txHash` is the A2A
+          // metadata key, not a wire field: the settle response reports the
+          // transaction id as `transaction`.
           event.metadata = {
             ...event.metadata,
-            txHash: response.txHash ?? response.transaction,
-            // Store the actual credits charged (especially important for margin-based)
-            creditsCharged: response.amountOfCredits
-              ? Number(response.amountOfCredits)
-              : creditsToBurn,
+            txHash: response.transaction,
+            // The credits this request asked to burn. The settle response does
+            // report what was actually redeemed, as `creditsRedeemed`, but
+            // reading it here would change behaviour and is not free: it is the
+            // string '0' on `billingModel: 'pay-as-you-go'` plans even when the
+            // buyer WAS charged, so a truthiness read reports 0 on a real
+            // charge. See SettlePermissionsResult before wiring it up.
+            creditsCharged: creditsToBurn,
           }
         }
       } catch (err) {
--- a/tests/unit/a2a/payments-request-handler.test.ts
+++ b/tests/unit/a2a/payments-request-handler.test.ts
@@ -530,6 +530,45 @@ describe('PaymentsRequestHandler', () => {
      * must not make it refuse the ordinary case — without this, deleting the
      * whole helper body and returning `undefined` would pass every test here.
      */
+    /**
+     * #439 review — `resolveCreditsCharged` read only `billingModel` and
+     * `creditsRedeemed`, never `success`. `settlePermissions` does NOT throw on
+     * a settle the backend accepted but could not complete — it returns 200 with
+     * `success: false` verbatim — so a FAILED settle published `creditsCharged`
+     * as though it were a measured redemption. With `creditsRedeemed: '7'` it
+     * reported 7 credits charged for a settle that charged nothing.
+     *
+     * That is a lie rather than noise, because the published contract defines
+     * `0` as "a real figure, not an absence": any number here asserts the settle
+     * completed.
+     */
+    test.each([
+      ['a failed settle reporting 0', '0'],
+      ['a failed settle reporting a non-zero figure', '7'],
+    ])('%s publishes nothing', async (_label, redeemed) => {
+      const { event, taskRef } = await finalize(
+        settleResponse({ success: false, billingModel: 'credits', creditsRedeemed: redeemed }),
+      )
+      expect(event.metadata).not.toHaveProperty('creditsCharged')
+      expect(taskRef.metadata).not.toHaveProperty('creditsCharged')
+    })
+
+    /**
+     * `!== true`, not `=== false`. An absent `billingModel` is a documented
+     * legacy shape read as `credits`; an absent `success` is not — the field has
+     * always been there and is declared non-optional, so its absence means a
+     * malformed response, which is exactly when not to publish a figure.
+     */
+    test('a settle with no success field at all publishes nothing', async () => {
+      const { event } = await finalize({
+        transaction: '0xabc',
+        network: 'eip155:84532',
+        billingModel: 'credits',
+        creditsRedeemed: '7',
+      } as any)
+      expect(event.metadata).not.toHaveProperty('creditsCharged')
+    })
+
     test('an ordinary decimal string is still published', async () => {
       const { event } = await finalize(
         settleResponse({ billingModel: 'credits', creditsRedeemed: '42' }),
--- a/tests/unit/a2a/payments-request-handler.test.ts
+++ b/tests/unit/a2a/payments-request-handler.test.ts
@@ -81,13 +81,14 @@ describe('PaymentsRequestHandler', () => {
     test('should burn credits when event has creditsUsed', async () => {
       // The settle response as the facilitator really returns it: the transaction
       // id is `transaction`, and there is no `txHash` or `amountOfCredits` on the
-      // wire (#438). `creditsRedeemed` is deliberately 3 rather than 5 so the
-      // `creditsCharged` assertion below distinguishes the credits this request
-      // ASKED to burn from the credits the settle reports redeeming.
+      // wire (#438). `creditsRedeemed` is deliberately 3 against a requested burn
+      // of 5 so the `creditsCharged` assertion below distinguishes the credits
+      // this request ASKED to burn from the credits the settle actually redeemed.
       const settleMock = jest.fn().mockResolvedValue({
         success: true,
         transaction: '0xabc',
         network: 'eip155:84532',
+        billingModel: 'credits',
         creditsRedeemed: '3',
       })
       mockPayments.facilitator.settlePermissions = settleMock
@@ -143,11 +144,11 @@ describe('PaymentsRequestHandler', () => {
         maxAmount: 5n,
       })
       // `txHash` is the A2A metadata key, sourced from the wire's `transaction`;
-      // `creditsCharged` reports the requested burn (5), not `creditsRedeemed` (3).
+      // `creditsCharged` reports what was REDEEMED (3), not what was requested (5).
       expect(event.metadata?.txHash).toBe('0xabc')
-      expect(event.metadata?.creditsCharged).toBe(5)
+      expect(event.metadata?.creditsCharged).toBe(3)
       expect(taskRef.metadata?.txHash).toBe('0xabc')
-      expect(taskRef.metadata?.creditsCharged).toBe(5)
+      expect(taskRef.metadata?.creditsCharged).toBe(3)
       expect(mockResultManager.processEvent).toHaveBeenCalledWith(taskRef)
     })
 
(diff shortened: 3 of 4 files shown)


r-marques:
## Review panel folded — 2 SHOULD-FIX, at `883d956`

Both reproduced before folding, not taken on report.

### 1. The discriminator was a denylist

`if (settlement.billingModel === 'pay-as-you-go')` meant **every other value** fell into the credits branch and published `creditsCharged: 0` — the exact "you were charged nothing" claim this PR exists to prevent. Measured: `null`, `''`, `'PAY-AS-YOU-GO'`, `'payAsYouGo'`, `' pay-as-you-go '`, `'subscription'` and a non-string all published `0`.

That is reachable: the settle response is an unchecked `(await response.json()) as SettlePermissionsResult` (`facilitator-api.ts:667`), so the union type provides no runtime guarantee.

Now an allowlist. **`undefined` still means `credits`** — the documented ruling is preserved verbatim — but a discriminator that is *present* and is not `credits` omits.

### 2. `Number()` maps the empty-ish family to `0`, not `NaN`

So `Number.isFinite` waved them straight through. Reproduced in node:

| input | old `creditsCharged` |
| --- | --- |
| `''`, `'   '`, `null`, `[]`, `false` | **`0`** |
| `true` | **`1`** |
| `'-5'` / `'0x10'` / `'1e3'` | **`-5` / `16` / `1000`** |
| `'9007199254740993'` | **`9007199254740992`** — silently off by one |

Now validated against the decimal-string contract the wire actually promises, mirroring `DECIMAL_INTEGER_STRING` in `mpp/mpp-api.ts`, which documents this identical trap for `BigInt`.

**Above `MAX_SAFE_INTEGER` it omits rather than rounds.** A wrong number is worse than no number, and nothing is lost: the exact value survives verbatim in the `x402.payment.receipts` entry `recordPaymentSuccess` writes onto the task. Flagging that as a judgement call I made rather than one I was given — say if you'd rather publish the decimal string and widen the type.

### Both fault branches now warn

Matching `x402/express/middleware.ts` on this same field. The pay-as-you-go branch stays **silent**: it is a design decision, not a fault, and an operator has to be able to tell them apart. That was the panel's "the omission is silent" finding.

### Mutation-tested, non-overlapping

| mutant | tests killed |
| --- | --- |
| restore the denylist | **7** — exactly the discriminator cases |
| restore `Number()` / `isFinite` | **11** — the 10 unusable values + the `MAX_SAFE_INTEGER` case |
| unmutated control | 662 passed, 48 suites |

Each group also has a **negative control** (`'42'` is still published), so a guard that passed by refusing everything would fail.

### Docs

`creditsCharged` was documented **nowhere** — `markdown/a2a-integration.md` covers `creditsUsed` in six places and never mentioned it. Added: what it means, the four cases, that absence is meaningful, and to read it with `in` rather than truthiness since `0` is legitimate.

### Verification

`tsc` clean · **662** unit tests (was 643) · lint 0 errors, 3 pre-existing warnings. Runner: jest via `pnpm test:unit`.

## Two findings deliberately NOT folded

- **The streaming path's write reaches nobody.** `processStreamingEventsWithFinalization` yields the event and only *then* settles and mutates `event.metadata`, and a2a-js's `ResultManager` copies only `status`/`history` onto the task — so on that path neither the old nor the new value reaches a remote consumer. **Pre-existing, not introduced here**, and the code already acknowledges a stream cannot retract what it yielded. Stating it rather than quietly fixing it, because the fix is a behaviour change on the streaming contract that deserves its own PR. It does mean this PR's headline claim holds for the non-streaming path only.
- **`src/mcp/core/paywall.ts:245`/`:408`** still does `settlement.creditsRedeemed ?? credits.toString()` — substituting the credits the request *asked* to burn and labelling them redeemed, i.e. exactly what this PR removes from A2A. Same repo, same money question, opposite answer depending on protocol. That is a product call, not a cleanup, so I have asked rather than decided it.

aaitor (commented):
Reviewed the diff (`resolveCreditsCharged` allowlist + decimal-string validation in `paymentsRequestHandler.ts`) against the PR's own description of round-1 findings — both (denylist → allowlist, `Number()` coercion trap) are genuinely fixed, with matching negative/positive controls in the test file (empty/whitespace/null/array/bool/hex/exponent/fraction all covered, plus a `MAX_SAFE_INTEGER` overflow case and an ordinary-value negative control). Checked the non-streaming and streaming call sites both route through the same helper and that `creditsUsed` (requested burn) is never conflated with `creditsCharged` (redeemed). CI green (9/9). No blockers found.

eruizgar91:
👀 Reviewing this now.

src/a2a/paymentsRequestHandler.ts
```diff
@@ -40,6 +40,101 @@ import type {
...
+  // ALLOWLIST, not a denylist. `=== 'pay-as-you-go'` let every other value fall
+  // into the credits branch and publish a `0`: `null`, `''`, `'PAY-AS-YOU-GO'`,
+  // a future third billing model, a non-string. The response is an unchecked
+  // `as SettlePermissionsResult` over `response.json()`, so the union type
+  // provides no runtime guarantee and the wrong values are reachable.
+  //
+  // `undefined` still means `credits` — that is the documented ruling above and
+  // it is preserved exactly. What changes is that a discriminator which is
+  // present but not `credits` now omits rather than reporting a zero.
+  if (settlement.billingModel !== undefined && settlement.billingModel !== 'credits') {
```
eruizgar91, line 110: **HIGH** — `resolveCreditsCharged` never reads `settlement.success`, so a settle that *failed* publishes `creditsCharged` as if it were a measured redemption.

`settlePermissions` does **not** throw on a settle the backend accepted but could not complete — it returns `200 + success: false` verbatim, and says so in its own comment: [`facilitator-api.ts#L666-L699`](https://github.com/nevermined-io/payments/blob/22ee030d0c68bb551f4db62102d35d7761ef611c/src/x402/facilitator-api.ts#L666-L699) — *"of the in-tree consumers only the MCP paywall actually branches on `success` — so a caller that forgets to check it serves a request it was never paid for with nothing in the log."* `executeRedemption` returns that result unchanged, and neither call site checks `success` before assigning the metadata.

Traced against the head SHA, `resolveCreditsCharged` reads only `billingModel` and `creditsRedeemed`:

| settle | published |
| --- | --- |
| `{success: false, billingModel: 'credits', creditsRedeemed: '0'}` | `creditsCharged: 0` |
| `{success: false, billingModel: 'credits', creditsRedeemed: '7'}` | `creditsCharged: 7` |

That collides head-on with the contract this PR publishes. `markdown/a2a-integration.md:262` defines `0` as *"a real figure, not an absence"* — i.e. a settle that completed and redeemed nothing. A failed settle is a different claim, and the second row is worse: a redemption figure for a settle that did not settle. `SettlePermissionsResult.billingModel`'s own docblock warns about exactly this — *"Reported whether or not the settle succeeded — so check `success` before treating it as evidence of a charge"* ([`facilitator-api.ts#L253-L264`](https://github.com/nevermined-io/payments/blob/22ee030d0c68bb551f4db62102d35d7761ef611c/src/x402/facilitator-api.ts#L253-L264)).

Both sibling surfaces already gate on it. The MCP paywall: `creditsRedeemed: creditsResult?.success ? (...) : '0'` ([`paywall.ts#L245-L247`](https://github.com/nevermined-io/payments/blob/22ee030d0c68bb551f4db62102d35d7761ef611c/src/mcp/core/paywall.ts#L245-L247)). The Express middleware: `if (!settlement.success) { creditsSettled = 0 }` ([`middleware.ts#L1081-L1083`](https://github.com/nevermined-io/payments/blob/22ee030d0c68bb551f4db62102d35d7761ef611c/src/x402/express/middleware.ts#L1081-L1083)). This helper is the one new consumer that doesn't.

It is also untested: `settleResponse` hardcodes `success: true` (`tests/unit/a2a/payments-request-handler.test.ts:387`), so every one of the 19 tests runs on a successful settle.

Fix, in this file:

```ts
if (settlement.success !== true) {
  return undefined
}
```

`!== true` rather than `=== false`, mirroring `facilitator-api.ts:687` — a 200 whose body carries no `success` at all is not a settlement either. Add it as a fourth documented `undefined` case in the JSDoc above, add the row to the doc table, and add a `success: false` fixture.

src/a2a/paymentsRequestHandler.ts
```diff
@@ -40,6 +40,101 @@ import type {
...
+    return undefined
+  }
+
+  const raw = settlement.creditsRedeemed
+  if (typeof raw !== 'string' || !DECIMAL_INTEGER_STRING.test(raw)) {
+    // Silence here would be indistinguishable from the pay-as-you-go omission,
+    // which is a deliberate design decision rather than a fault. Warn, as
+    // `x402/express/middleware.ts` does on this same field.
+    if (raw !== undefined) {
```
eruizgar91: **LOW** — the absent case does **not** warn, contradicting this comment, the JSDoc above, and the precedent it cites.

`if (raw !== undefined)` makes the most common "no usable figure" — `creditsRedeemed` missing entirely, i.e. a Nevermined API older than the field — completely silent. Three places say otherwise:

- the JSDoc at line 75: *"`creditsRedeemed` is absent, **or** not a plain decimal string — warns"*;
- `markdown/a2a-integration.md:264`: *"the backend reported no usable figure | key absent **(and the SDK warns)**"*;
- the cited precedent actually warns on this exact branch: `'[paymentMiddleware] MPP settlement reported no creditsRedeemed'` ([`middleware.ts#L1096-L1106`](https://github.com/nevermined-io/payments/blob/22ee030d0c68bb551f4db62102d35d7761ef611c/src/x402/express/middleware.ts#L1096-L1106)).

The test that covers it (`payments-request-handler.test.ts:622`, `settleResponse({})`) asserts only the omission, so nothing caught the divergence.

Either warn on the absent case too — which is what all three documents promise, and which keeps the "only faults are silent-free" property the JSDoc claims — or, if silence is intended because an old backend is not a fault, say so here and fix lines 74-76 and the doc row. Right now the code and every description of it disagree.

src/a2a/paymentsRequestHandler.ts
```diff
@@ -862,16 +963,19 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
...
-            txHash: response.txHash ?? response.transaction,
-            // Store the actual credits charged (especially important for margin-based)
-            creditsCharged: response.amountOfCredits
-              ? Number(response.amountOfCredits)
-              : creditsToBurn,
+            txHash: response.transaction,
+            // The credits actually redeemed, omitted entirely when the billing
+            // model has none to report — see resolveCreditsCharged. The credits
+            // this request asked to burn stay available as `creditsUsed`.
+            ...(creditsCharged !== undefined && { creditsCharged }),
```
eruizgar91, line 1008: **LOW** — `creditsCharged` and `txHash` have no home in any exported type, and this PR is what makes that expensive.

`PaymentMetadata` — the interface for this object ([`types.ts#L128-L141`](https://github.com/nevermined-io/payments/blob/22ee030d0c68bb551f4db62102d35d7761ef611c/src/a2a/types.ts#L128-L141)) — declares `creditsUsed`, `planId`, `paymentType`, `costDescription`, `agentId`, `redemptionConfig`, and neither of the two keys the SDK writes here. They have only ever reached consumers through the `& Record<string, any>` escape hatch at `types.ts:95`, and `@a2a-js/sdk`'s `Task.metadata` is an optional index signature, so `tsc` can express nothing about them either way.

That gap predates the PR. What the PR changes is the cost of it: `creditsCharged`'s **absence** is now load-bearing and documented (`markdown/a2a-integration.md:254`, "its absence is meaningful"), and the only machine-readable statement of that contract is prose in a markdown file.

Bounded fix in one adjacent file — add both to `PaymentMetadata` as optional, with the absence rule on `creditsCharged`'s doc comment:

```ts
/** The settle's transaction reference (chain tx hash on crypto rails, PSP id on fiat). */
txHash?: string
/** Credits actually redeemed. ABSENT when the billing model has none to report — see resolveCreditsCharged. */
creditsCharged?: number
```

markdown/a2a-integration.md
```diff
@@ -240,6 +240,50 @@ eventBus.publish({
 })
 ```
 
+### What the SDK adds to `metadata` after settlement
+
+Your executor sets `creditsUsed`. After the request settles, the SDK adds two more
+keys to the same `metadata` object, on the status-update event and on the stored
```
eruizgar91: **MEDIUM** — "and on the stored task" is false on the streaming transport, and false under batch redemption.

`task.metadata` has exactly one writer in this file — `handleTaskFinalization:1034` (`grep -n "task.metadata" src/a2a/paymentsRequestHandler.ts` returns only line 1034-1035 at the head SHA). `processStreamingEventsWithFinalization` mutates only `event.metadata` (lines 657-664); its `resultManager.processEvent(task)` runs solely inside `if (httpContext?.inBand)` (line 672) and passes the task, not the merged metadata. It is also a moot point on the wire: the streaming loop does `yield event` at line 628, *before* the settlement block, so the streamed event does not carry these keys either.

So on the streaming transport a consumer following the snippet below finds neither key on `task.metadata` — ever. Under `useBatch` (line 637 / 958) the whole block is skipped and neither key is written on either transport.

The underlying asymmetry is pre-existing, so mirroring the merge into the streaming path is a behaviour change beyond this PR. The bounded fix is to scope the sentence: say the keys land on the status-update event, and on the stored task on the non-streaming path, and that batch redemption writes neither.

markdown/a2a-integration.md
```diff
@@ -240,6 +240,50 @@ eventBus.publish({
...
+| credits plan, nothing redeemed | `0` — a real figure, not an absence |
+| **pay-as-you-go** | **key absent** |
+| the backend reported no usable figure | **key absent** (and the SDK warns) |
+
+Pay-as-you-go is the case to understand. Such a plan holds no credit balance, so
+`creditsRedeemed` is the string `'0'` **even on a settle that charged the buyer**.
+Publishing that as `creditsCharged: 0` would tell a buyer they were charged
+nothing for a payment that really happened, so nothing is published instead. The
+charge is referenced by `orderTx` (fiat) or `transaction` (crypto), and the full
+settlement receipt is on the task under `x402.payment.receipts`.
```
eruizgar91: **MEDIUM** — this is the doc's designated fallback for the case where `creditsCharged` is deliberately withheld, and it is wrong in two ways.

**It does not exist on the legacy transport.** `recordInBandSettlement` early-returns unless the token arrived in band — `if (!task || !httpContext?.inBand) { return }` (`paymentsRequestHandler.ts:579-581`), whose own docblock says *"Only applied when the token arrived in band (so the legacy `payment-signature` header path is unchanged)"*. `recordPaymentSuccess` is the only writer of the receipts key on success (`x402-a2a.ts:252-259`) and is reachable only through that guard. On the header transport there are **no receipts at all**. This section sits under "EventBus Events", which is transport-agnostic, and the in-band qualifier is ~140 lines earlier with no cross-reference.

**The location is wrong.** Receipts are written to `task.status.message.metadata` — that is what `ensureStatusMetadata` returns (`x402-a2a.ts:100-117`) — not `task.metadata`. Three lines below, the sample reads `task.metadata`, so a reader will look in the wrong place.

Under `useBatch`, `recordPaymentDeferred` writes the deferred marker and no receipts (`x402-a2a.ts:240-246`), so the fallback is absent there too, even in band.

Fix: scope the sentence to the in-band x402 transport and name `task.status.message['x402.payment.receipts']`.

Worth a few words in the same edit: the `txHash` row at line 251 reads "The settlement transaction, from the facilitator's receipt". It is `response.transaction` verbatim, which on fiat rails is a Stripe PaymentIntent id, not a hash — `markdown/x402.md:812-822` shows `"network": "stripe", "transaction": "pi_3U6tgr…"`. Saying "a chain tx hash on crypto rails, the PSP transaction id on fiat rails" costs one line and stops a support ticket.

markdown/a2a-integration.md
```diff
@@ -240,6 +240,50 @@ eventBus.publish({
...
+
+**Read it with `in`, not truthiness** — `0` is a legitimate value:
+
+```typescript
+if ('creditsCharged' in task.metadata) {
```
eruizgar91: **LOW** — this snippet throws when `task.metadata` is undefined, which is the normal state on the path it is most likely to be used on.

`Task.metadata` is optional in `@a2a-js/sdk`, and `'creditsCharged' in undefined` is a `TypeError`, not a falsy branch — so the `else` arm this snippet advertises is unreachable in exactly the case it is written for. On the streaming transport `task.metadata` is never written at all (see the comment on line 246).

```typescript
if (task.metadata && 'creditsCharged' in task.metadata) {
```

The `in`-over-truthiness advice itself is right and worth keeping.

eruizgar91 (commented):
Independent review of the money-path change, by construction rather than by reading the description. **Head SHA `22ee030` verified unchanged before and after the analysis** — no mid-review push.

Findings are inline: **1 HIGH, 2 MEDIUM, 3 LOW**. No verdict is attached to this review.

### The two round-1 fixes: both genuinely fixed, verified by execution

**The denylist is now an allowlist, and the default branch is the safe one.** Enumerated the whole discriminator space against `if (settlement.billingModel !== undefined && settlement.billingModel !== 'credits')` in a runtime harness. Only `'credits'` and `undefined` fall through to publish; `null`, `''`, `'   '`, `'PAY-AS-YOU-GO'`, `'Credits'`, `'payAsYouGo'`, `' pay-as-you-go '`, `'subscription'`, `0`, `{}`, `[]` all omit. `undefined` falling through is the deliberate documented ruling and matches the repo-wide rule in `SettlePermissionsResult` ("never read a missing discriminator as pay-as-you-go") — consistent, not a hole.

**The `Number()` claim holds, and the regex is what carries it.** Confirmed `Number('')`, `Number('   ')`, `Number(null)`, `Number([])`, `Number(false)` are all `0` with `Number.isFinite` **true** — so `isFinite` alone would have published each as `creditsCharged: 0`. `typeof raw !== 'string' || !/^\d+$/.test(raw)` runs first and rejects all of them, plus `'-5'`, `'0x10'`, `'1e3'`, `'1.5'`, `'+1'`, `' 1 '` and `'123\n'` (JS `$` is end-of-input without `m`, so no trailing-newline hole). `Number.isSafeInteger` correctly catches `'9007199254740993'` → `9007199254740992` and the `Infinity` overflow. Both tables are pinned by non-vacuous tests.

### Consumer side of the omission

Swept the whole tree for readers of `creditsCharged` (`src/`, `cli/`, `openclaw/`, `tests/`, `docs/`, `markdown/`, `scripts/`, `.github/`, plus a binary-skipping case-insensitive pass). **Exactly three files match: the handler, its test, and the doc this PR updates.** No `creditsCharged ?? x`, no `Number(metadata.creditsCharged)`, no interface declaring it required — so nothing in-repo reintroduces a misleading zero. The residual exposure is external published-SDK consumers; the changelog is generated from PR titles by `auto-changelog`, so the title is what they will read.

### Checked and deliberately not raised

- **`txHash = response.transaction`** — the dropped `response.txHash` really was dead. `SettlePermissionsResult` has no such field, and on fiat pay-as-you-go `transaction` carries the same `pi_…` as `orderTx` (`markdown/x402.md:812-822`), so no charge reference is lost by the change. Only the doc wording is imprecise (noted inline).
- **Omission is "don't set", not "delete"** — a pre-existing `creditsCharged` on `event.metadata` would survive both call sites. Could not demonstrate a path where anything but the SDK writes that key, so not raised as a finding.
- **Tests drive production code**, not a re-implementation: `finalize` calls the real `handleTaskFinalization`, `finalizeStreaming` the real generator, and `taskRef` is the object line 1034 mutates — the negative assertions are anchored by the positive control at line 545.
- **The `absent billingModel` + pay-as-you-go gap** (an older backend publishes `creditsCharged: 0` on a PAYG plan) is inherent to the documented ruling and is the correct trade against breaking credits plans. Not a defect.
- **CI green at this SHA** (10/10 checks); `pnpm lint` does not cover `tests/`, as the description notes.

eruizgar91 (changes requested):
🔴 **Blocked** — 1 blocker(s) to address before merge. Details are inline on the changed lines.

Inline review: https://github.com/nevermined-io/payments/pull/439#pullrequestreview-5177367958

src/a2a/paymentsRequestHandler.ts
r-marques: You are right, and this is the one I most needed caught. Folded in `1dab1c2`.

`resolveCreditsCharged` now returns `undefined` unless `settlement.success === true`, so a failed settle publishes no figure. Your `creditsRedeemed: '7'` row is what makes it a defect rather than noise — **my own earlier analysis probed only `'0'` and concluded it was truthful.** With `'0'` a failed settle publishes a zero, which reads as harmless; with `'7'` it publishes a measured-looking redemption for a settle that redeemed nothing. Same bug, and I picked the input that hid it.

That it collides with the contract this PR itself publishes — `markdown/a2a-integration.md` defining `0` as *"a real figure, not an absence"* — is the part that settles it. Any number there asserts the settle completed.

**One decision inside the fix, flagged because it goes beyond the literal finding:** the guard is `!== true`, not `=== false`, so an ABSENT `success` also omits. The asymmetry with `billingModel` directly above is deliberate and documented in the comment: an absent `billingModel` is a *documented legacy shape* (APIs predating the field) and is read as `credits`; `success` has always been there and is declared non-optional, so its absence means a malformed response — which is exactly when not to publish a figure. Say if you would rather it were `=== false`.

It is **silent**, like the pay-as-you-go branch, on the grounds that a failed settle is a legitimate outcome the caller already sees on `success`, not a reporting fault.

Three tests, mutation-verified: removing the guard kills all three (failed-with-`'0'`, failed-with-`'7'`, and absent-`success`), against a 670-green control.

src/a2a/paymentsRequestHandler.ts
r-marques: Fixed in `1dab1c2`. Both branches warn now, with distinct messages.

You are right that three separate places promised it and the code did not deliver: the JSDoc at line 75, the markdown row *"(and the SDK warns)"*, and — the one that settles it — the precedent I cited actually warns on **exactly this branch** (`'[paymentMiddleware] MPP settlement reported no creditsRedeemed'`). Citing a precedent and then not following it on the one branch it covers is worse than not citing it.

I chose to warn rather than to relax the docs, for the reason your comment implies: the absent case is the **most common** "no usable figure", so making it the silent one inverts the property the JSDoc claims — that only faults are silent-free. Silence is reserved for pay-as-you-go, which is a design decision rather than a fault.

Also noting your point that `payments-request-handler.test.ts:622` asserted only the omission, so nothing caught the divergence. The test was right about the behaviour and blind to the signal — which is why the docs and the code could disagree for as long as they did.

src/a2a/paymentsRequestHandler.ts
r-marques: Folded in `1dab1c2` — both added to `PaymentMetadata`, with the absence rule on `creditsCharged`'s doc comment.

Your framing is the reason I did it here rather than deferring: the gap predates this PR, but **this PR is what makes it expensive.** Once `creditsCharged`'s absence is load-bearing and documented, the only machine-readable statement of that contract being prose in a markdown file is the actual problem — not the missing declaration.

I also took your `txHash` wording into the type rather than only the docs: *"A chain tx hash on the crypto rails; the PSP transaction id on fiat (a Stripe PaymentIntent `pi_...`), so do not assume it is a hash."* That one is a support ticket waiting to happen, and the type is where someone will read it.

markdown/a2a-integration.md
r-marques: Folded in `1dab1c2`, and I verified each leg in source rather than from the report — the claim is about three transports and I did not want to document a fourth wrong thing:

- `task.metadata` has exactly **one** writer, line 1057, on the non-streaming path.
- the streaming loop does `yield event` at **648**, before the settlement block.
- `useBatch` gates both paths (`if (!redemptionConfig.useBatch)` at 664 and 982).

All three hold. Replaced the sentence with a per-transport table — non-streaming gets both keys on the event and the task; streaming gets neither, on either; batch gets neither.

I took your scoping rather than mirroring the merge into the streaming path, for the reason you gave: that is a behaviour change beyond this PR. Worth stating plainly though, since the table now makes it visible — **on the streaming transport these keys reach nobody.** Whether that should change is a separate question from whether the docs should stop claiming otherwise.

markdown/a2a-integration.md
r-marques: Folded in `1dab1c2`. Both halves were wrong and the second is the one that would have cost someone real time.

**Transport:** the fallback is now scoped to the in-band x402 path, with the note that the legacy `payment-signature` header path has **no receipts at all** and `useBatch` writes a deferred marker instead. `recordInBandSettlement`'s early return (`if (!task || !httpContext?.inBand) return`) makes that unambiguous.

**Location:** corrected to `task.status.message.metadata['x402.payment.receipts']`. I had written `task.metadata` three lines above a sample that also reads `task.metadata`, so a reader would have looked in the wrong place twice and concluded the receipts did not exist.

**And the `txHash` wording** — taken, in both the markdown and the `PaymentMetadata` doc comment: *"A chain tx hash on the crypto rails; on fiat it is the PSP transaction id (a Stripe PaymentIntent, `pi_...`), so do not assume it is a hash."* You are right that it is one line against a support ticket.

markdown/a2a-integration.md
r-marques: Fixed in `1dab1c2` — the sample now guards `task.metadata &&` before the `in`.

Your point that it throws *"in exactly the case it is written for"* is what makes this more than a style nit: the `else` arm exists to explain pay-as-you-go and no-usable-figure, and on the streaming transport — where `task.metadata` is never written at all — the snippet would `TypeError` instead of reaching that arm. A reader following the doc on the path it most applies to would get a crash and no explanation.

The `in`-over-truthiness advice itself stays, as you said — `0` is a legitimate value and truthiness is the trap the whole section is about.

aaitor (commented):
Reviewed anyway despite the bot tooling blocker (see Discord) — same run-dir permission gate as #437, so no deterministic verdict.

Read the full diff. The HIGH from the prior round (`success: false` with `creditsRedeemed: '7'` publishing a measured-looking redemption for a settle that charged nothing) is fixed: `resolveCreditsCharged` now gates on `settlement.success !== true` before reading `creditsRedeemed` at all, ahead of the billing-model/decimal-string checks. The asymmetry with the `billingModel` check (absent → treated as legacy `credits`, vs. absent `success` → treated as malformed) is explained in the code comment and matches the field's non-optional contract. Allowlist (not denylist) on `billingModel === 'credits'`, decimal-string regex guard against `Number()`'s empty-ish-family-to-0 coercion, and a `MAX_SAFE_INTEGER` guard are all mutation-style tested (`test.each` covering null/empty/wrong-case/padded/third-model/non-string for the billing-model guard, and empty/whitespace/null/array/bool/negative/hex/exponent/fraction for the decimal guard). Both call sites (plain response, streaming) are driven from the same test suite. CI green.

No blockers found.

aaitor (commented):
Reviewed independently (fresh context), all 7 commits: the resolveCreditsCharged billing-model rule, the allowlist-not-denylist fix, the decimal-string validation (Number() empty-family and MAX_SAFE_INTEGER traps), and the round-3 fold — resolveCreditsCharged now correctly guards on `success !== true` so a failed settle with a stray creditsRedeemed never publishes a figure. Mutation-driven test suite (19+ cases with negative controls) backs every guard. No unresolved threads, CI green. No blockers found.

aaitor (approved):
✅ **Approved** — no blockers, nothing to address.

Inline review: https://github.com/nevermined-io/payments/pull/439#pullrequestreview-5189869585

eruizgar91 (commented):
**Round 3** — re-review of the round-2 blocking finding on `c954a298`, verified by running the branch rather than reading it.

### The round-2 blocking finding is closed

The guard is `settlement.success !== true` (not `=== false`), it lives in the single shared `resolveCreditsCharged`, and **both** publish paths for `event.metadata.creditsCharged` go through it — `handleTaskFinalization` (`:1001`) and `processStreamingEventsWithFinalization` (`:691`). A repo-wide grep for `creditsCharged` finds no third writer in `src/`.

Evidence, run locally at `c954a298`:

- **Probe, non-zero sentinel `creditsRedeemed: '7'`** (not `'0'`, so a leak is distinguishable from "nothing happened"):
  - `{success: false, billingModel: 'credits', creditsRedeemed: '7'}` → `event.metadata` = `{creditsUsed: 5, txHash: ""}`, `task.metadata` identical. No `creditsCharged`. Same on the streaming path.
  - `success` **absent entirely**, `creditsRedeemed: '7'` → `{creditsUsed: 5, txHash: "0xdead"}`. No `creditsCharged`. Same on the streaming path.
- **Mutation A** — delete the `success` guard: 3 tests go red (`a failed settle reporting 0`, `a failed settle reporting a non-zero figure`, `a settle with no success field at all`). Matches the author's claim exactly.
- **Mutation B** — weaken `!== true` to `=== false`: `a settle with no success field at all publishes nothing` goes red on its own. So the `!== true`/`=== false` distinction is itself pinned, not incidental.
- The tests are not vacuous in the other direction either: `an ordinary decimal string is still published` and `a credits plan that redeemed nothing reports 0` mean a helper that returned `undefined` unconditionally would fail.

The asymmetry (`billingModel` absent ⇒ `credits`; `success` absent ⇒ refuse) is right, and matches `settlePermissions` in `src/x402/facilitator-api.ts:684`, which already reasons `!== true` for the same reason. `SettlePermissionsResult.success` is declared non-optional, so its absence really is malformed rather than legacy.

### The merge from main is clean

`git diff 6756942..c954a298 -- cli/ tests/ .github/ tsconfig.json package.json` returns only this PR's own test file. `cli/jest.config.js`, `cli/tsconfig.json`, the workflows and the tsconfigs are byte-identical to main — no test config or coverage was weakened by the merge.

### Local validation

`tsc` ✅ · `tsc --noEmit -p tests/tsconfig.json` ✅ · `eslint ./src` ✅ (5 pre-existing warnings, 0 errors) · `jest --testPathPattern=tests/unit` → **49 suites, 673 passed, 1 skipped**, reproducing the author's numbers.

### What is filed, and why none of it blocks

4 findings inline: **2 MEDIUM, 2 LOW, 0 CRITICAL/HIGH.**

The two MEDIUMs are the *adjacent* surfaces this PR's own doctrine now makes visible — `recordInBandSettlement`'s `=== false` (unchanged since before this branch; it is at `:484` in the base) and the unconditional `txHash`. Neither is a regression introduced here, and neither reaches `creditsCharged`. They are small enough to close in this PR and the reasoning is already written three lines away, which is why they are filed rather than deferred — but they are not the round-2 finding and should not gate it.

### Considered and discarded

- *`task.metadata` assertions in the new tests are vacuous* — they are not: `handleTaskFinalization` spreads `event.metadata` into `task.metadata` at `:1060`, and the probe confirms both objects.
- *The streaming path only proves the `success` guard via the non-streaming call site* — true, but the guard is a single shared function now, and both call sites are driven for the billing-model rule. Not worth a finding.
- *The doc's `streaming → ❌ neither` row contradicts the streaming test asserting `creditsCharged: 3`* — the doc is describing what the **client** observes (the event is `yield`ed before settlement mutates it), which is correct.
- *`Number.isSafeInteger` / `/^\d+$/` are over-strict* — no: `Number('')`, `Number('   ')`, `Number(null)`, `Number([])`, `Number(false)` are all `0`, and each is pinned by a `test.each` row.

_No verdict is submitted with this review — this is a bot-mode `COMMENT`. The standing `CHANGES_REQUESTED` is untouched here and is resolved separately._

src/a2a/paymentsRequestHandler.ts
```diff
@@ -567,11 +702,7 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
               if (httpContext?.inBand) {
                 const task = resultManager.getCurrentTask()
                 if (task) {
-                  this.recordInBandSettlement(
-                    task,
-                    httpContext,
-                    response as SettlePermissionsResult,
-                  )
+                  this.recordInBandSettlement(task, httpContext, response)
```
eruizgar91, line 717: **MEDIUM** — The second publish path. `resolveCreditsCharged` now refuses on `success !== true`, and the comment above it argues (correctly) that an absent `success` means a malformed response and "a malformed response is exactly when not to publish a figure". But `recordInBandSettlement`, which this line feeds, still routes on `=== false`:

```ts
if (settlement && settlement.success === false) {
  x402A2AUtils.recordPaymentFailure(...)
} else if (settlement) {
  x402A2AUtils.recordPaymentSuccess(task, settlement)
}
```

So the very response the new guard refuses to read lands on the task as a *success*. Probed at `c954a298`, in-band, settle returning `{transaction: '0xdead', network: 'n', billingModel: 'credits', creditsRedeemed: '7'}` with no `success`:

```json
{"x402.payment.status": "payment-completed",
 "x402.payment.receipts": [{"transaction":"0xdead","network":"n","billingModel":"credits","creditsRedeemed":"7"}]}
```

`creditsCharged` is correctly omitted, and the `7` is published anyway — one field over, under a `payment-completed` banner. That is the same shape as the finding this PR fixed, on the surface a buyer is told to read for the authoritative receipt.

It is also the odd one out in the repo: `settlePermissions` already uses `!== true` for exactly this case and warns "did not settle. Do not treat the request as paid" — and then this line stamps it completed.

Not a regression (the `=== false` predates the branch — it is at `:484` in `6756942`), so it is not blocking. But the fix is `settlement.success !== true` in `recordInBandSettlement`, in a file you are already editing, with the rationale already written in the block above `resolveCreditsCharged`. Worth closing here rather than leaving the two guards disagreeing three lines apart.

https://github.com/nevermined-io/payments/blob/c954a298a950566a7793f87d8139db0ad23da72c/src/a2a/paymentsRequestHandler.ts#L610-L620

src/a2a/paymentsRequestHandler.ts
```diff
@@ -862,16 +993,19 @@ export class PaymentsRequestHandler extends DefaultRequestHandler {
...
+          // transaction id as `transaction`.
+          const creditsCharged = resolveCreditsCharged(response)
           event.metadata = {
             ...event.metadata,
-            txHash: response.txHash ?? response.transaction,
-            // Store the actual credits charged (especially important for margin-based)
-            creditsCharged: response.amountOfCredits
-              ? Number(response.amountOfCredits)
-              : creditsToBurn,
+            txHash: response.transaction,
```
eruizgar91: **MEDIUM** — `txHash` is written unconditionally on the same two lines whose sibling key was just made conditional. `SettlePermissionsResult.transaction` is documented as an "empty string if settlement failed", so a failed settle publishes `txHash: ''`.

Probed at `c954a298`, `{success: false, errorReason: 'INSUFFICIENT', transaction: '', creditsRedeemed: '7'}`, legacy (non-in-band) path:

```json
{"creditsUsed": 5, "txHash": ""}
```

The task completes, and that is the buyer's entire record of a settle that failed — on the legacy `payment-signature` path there are no x402 receipts to fall back on. The PR's new doc table presents `txHash` as "The settlement transaction" with no absence semantics, right beside the `creditsCharged` row whose absence is explicitly load-bearing, so a consumer following the doc reads a blank id rather than a failure.

The value is falsy so a truthiness check degrades safely, and `''` is what the old `response.txHash ?? response.transaction` produced too — hence MEDIUM, not blocking. But the fix is the conditional spread already two lines below:

```ts
const settled = response.success === true
event.metadata = {
  ...event.metadata,
  ...(settled && response.transaction ? { txHash: response.transaction } : {}),
  ...(creditsCharged !== undefined && { creditsCharged }),
}
```

Apply it at `:687` too. If you'd rather keep publishing it, say so in the doc table — an empty `txHash` meaning "no settlement" is a contract, and it is currently undocumented.

markdown/a2a-integration.md
```diff
@@ -240,6 +240,69 @@ eventBus.publish({
...
+reports what was redeemed, which is not necessarily what you asked to burn —
+under margin-based pricing the two differ. The SDK omits the key entirely rather
+than report a figure that would be false:
+
+| situation | `creditsCharged` |
+| --- | --- |
+| credits plan | the credits redeemed |
+| credits plan, nothing redeemed | `0` — a real figure, not an absence |
+| **pay-as-you-go** | **key absent** |
+| the backend reported no usable figure | **key absent** (and the SDK warns) |
```
eruizgar91, line 277: **LOW** — This table is the published contract for `creditsCharged`'s absence, and it is missing the row the round-2 review added the guard for: **a settle that failed** (or one with no `success` field). Verified by probe — both omit the key.

The omission is not neutral, because the last row says the absence comes "(and the SDK warns)". A failed settle deliberately does *not* warn (the code comment: "Silent, like the pay-as-you-go branch: a failed settle is a legitimate outcome the caller already sees on `success`"), so a reader who sees no key and no warning concludes from this table that one of the two documented cases is impossible.

One row closes it:

```
| the settle failed, or reported no `success` | **key absent** (silent) |
```

`src/a2a/types.ts:148-157` already documents the failed-settle case correctly — this table is the one that drifts.

src/a2a/paymentsRequestHandler.ts
```diff
@@ -40,6 +40,131 @@ import type {
 
 const terminalStates: TaskState[] = ['completed', 'failed', 'canceled', 'rejected']
 
+/**
+ * The credits a settle actually redeemed, for `event.metadata.creditsCharged`.
+ *
+ * Reads {@link SettlePermissionsResult.creditsRedeemed} per billing model rather
+ * than trusting it flat, because it does not mean the same thing under both:
+ *
+ * - `credits` — the plan holds a balance and `creditsRedeemed` is what came out
```
eruizgar91, line 61: **LOW** — Two `/** */` blocks are stacked here, so the wrong one is attached. This block documents `resolveCreditsCharged` — it opens "The credits a settle actually redeemed" and ends with `@param settlement` / `@returns` — but the declaration that immediately follows it is the `DECIMAL_INTEGER_STRING` docblock, and then the regex. Net effect at `:59` and `:100`:

- `DECIMAL_INTEGER_STRING` gets its own docblock (fine, but this 55-line one is the first TSDoc a tool sees above it),
- `resolveCreditsCharged` ends up with **no** docblock at all — which is the one with `@param`/`@returns` and the whole billing-model ruling.

The repo runs `pnpm doc` (typedoc) and eslint's `tsdoc/syntax`, so this is the documentation that goes stale first. Move the `resolveCreditsCharged` block down to sit directly above `function resolveCreditsCharged(...)`, leaving `DECIMAL_INTEGER_STRING`'s block on the const.

eruizgar91 (approved):
🟡 **Approved with comments** — no blockers; 2 should-fix, 2 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/439#pullrequestreview-5209649521

src/a2a/paymentsRequestHandler.ts
aaitor, line 717: **SHOULD-FIX** — The new `resolveCreditsCharged` guard correctly treats an absent `success` as malformed, but the in-band seam called here still uses `settlement.success === false`; every other value reaches `recordPaymentSuccess`. A response with no `success` is therefore stamped `payment-completed` and its receipt is published even though the sibling guard refuses to call it a measured redemption. Use the same `success !== true` rule in `recordInBandSettlement` and pin the missing-field case through the in-band path.

src/a2a/paymentsRequestHandler.ts
aaitor: **SHOULD-FIX** — This publishes `txHash` unconditionally while the new sibling guard deliberately suppresses `creditsCharged` on a failed settle. `SettlePermissionsResult` uses an empty `transaction` on failure, so the legacy path leaves buyers with `{ txHash: '' }` even though the new docs define this key as the settlement transaction. Omit it unless `response.success === true` and the transaction is usable (and apply the same rule at the streaming writer), or document the blank-value contract explicitly.

markdown/a2a-integration.md
aaitor, line 277: **SHOULD-FIX** — The `success !== true` branch added in this PR creates another documented absence case: a failed settle, or a malformed settle with no `success`, omits `creditsCharged` silently. This table only offers pay-as-you-go or a no-usable-figure case that warns, so it now contradicts the code and the exported type comment. Add the failed/missing-success row and its silent behavior.

src/a2a/paymentsRequestHandler.ts
aaitor, line 61: **SHOULD-FIX** — This block documents `resolveCreditsCharged` and ends with `@param settlement` / `@returns`, but another `/** */` block plus `DECIMAL_INTEGER_STRING` sits between it and the function. Documentation tools attach only the adjacent block to a declaration, leaving the helper without this intended contract. Move this block directly above `resolveCreditsCharged`, after the regex declaration.

aaitor (commented):
Reviewed the current head with the incomplete-refactor, documentation-drift, mutation, and money/error-path lenses. The new `success !== true` guard and its focused tests are sound; four adjacent consistency issues remain inline. Validation: build, test typecheck, lint (0 errors; 5 pre-existing warnings), and the focused 38-test A2A suite all pass.

src/a2a/paymentsRequestHandler.ts
r-marques: Folded in `0d33463`. You are right that two guards disagreeing three lines apart is not a state to leave, and the argument for `!== true` was already written directly above — so keeping `=== false` here would have meant the file arguing against itself.

`recordInBandSettlement` now refuses on `settlement.success !== true`, with the rationale pointing at `resolveCreditsCharged` and at `settlePermissions`, which already reasons this way.

Pinned rather than asserted: reverting it to `=== false` turns **1 test red** — `a settle with no success field is recorded in band as payment-failed, not completed`. The failed-settle row stays green under that mutation, which is the point: `=== false` genuinely does catch that one, so a test that only covered it would have proved nothing about the change. There is a positive control too (`a SUCCESSFUL settle is still recorded in band as payment-completed`), otherwise a guard that stamped everything `payment-failed` would satisfy the new assertions.

Noted that it is not a regression from this branch — I folded it because it is small, in a file already open, and the reasoning was already written, not because the PR caused it.

src/a2a/paymentsRequestHandler.ts
r-marques: Folded in `0d33463` — took the fix rather than the "say so in the doc table" alternative, and the reason is worth stating because it was the one judgement call in this round.

The two options are not symmetric. Omitting is an application of a rule that has **already been decided on this PR** — a failed settle has no measured figure to report, so it gets no figure — to the sibling key on the same two lines. Keeping `''` would mean *creating* a contract that does not currently exist ("empty `txHash` means no settlement"), which is a larger commitment than removing an undocumented blank. And your legacy-path point is what settles it: with no x402 receipts to fall back on, a blank id under a completed task reads as *"settled, reference unavailable"* rather than *"did not settle"*.

Applied at both call sites — `:687` (streaming) and the finalization spread — plus documented in **both** places an absence has to be written down to be a contract: `types.ts` on `txHash` itself, and the `txHash` row of the markdown field table.

Mutation-tested: making either one unconditional again turns **3 tests red**. The positive control (`a SUCCESSFUL settle still publishes txHash on both paths`, asserting `'0xabc'` on event, task and stream) means a `txHash` that was simply never published could not pass.

If you would rather have the `''` back as a documented contract, say so — it is a one-line revert and the doc rows are already written either way.

markdown/a2a-integration.md
r-marques: Folded in `0d33463`, with your row as written:

```
| the settle failed, or reported no `success` | **key absent** (silent) |
```

Your diagnosis of *why* the omission was not neutral is the part I would have missed — the last row's "(and the SDK warns)" makes silence read as impossible, so adding the row alone would have left the reader with two silent cases and one warning and no way to tell which they were in. Added a short note under the table saying the silent rows are deliberately silent and why: a failed settle is an outcome the caller already sees on `success`, not a reporting fault.

While in there, the `txHash` row of the field table above now carries its absence semantics too — that became load-bearing with the other MEDIUM, and an absence is only a contract if it is written down.

src/a2a/paymentsRequestHandler.ts
r-marques: Folded in `0d33463`. Moved the 45-line block down to sit directly above `function resolveCreditsCharged`, leaving `DECIMAL_INTEGER_STRING`'s own block on the const.

Confirmed by boundary rather than by eye — the two blocks were `43-87` and `88-97` with the const at `98`, so the first was attached to the second's declaration and the function had none at all. `npx eslint ./src` after the move: 0 errors, 5 warnings, all pre-existing and the same five you measured.

src/a2a/paymentsRequestHandler.ts
r-marques: Folded in `0d33463`, and worth noting you and Kike landed on this independently — same site, same rule. `recordInBandSettlement` now uses `settlement.success !== true`.

Pinned through the in-band path as you asked, rather than only through the helper: two new tests drive `recordInBandSettlement` directly and assert `x402.payment.status` is `payment-failed` for both a failed settle and one with no `success` field, plus a positive control that a successful settle is still `payment-completed`. Reverting the guard to `=== false` turns 1 of those red — the missing-field case, which is precisely the one `=== false` does not catch.

src/a2a/paymentsRequestHandler.ts
r-marques: Folded in `0d33463` — you and Kike raised this independently, same line.

Took the omit option rather than documenting the blank: omitting applies a rule already decided on this PR to the sibling key on the same two lines, whereas keeping `''` would mean creating a contract that does not exist today. Applied at the streaming writer too, as you asked.

Then documented it anyway, in both places an absence has to be written down to count: `types.ts` on `txHash`, and the `txHash` row of the markdown field table — its presence now means a settlement really happened. Making either writer unconditional again turns 3 tests red; a positive control pins that a successful settle still publishes it.

markdown/a2a-integration.md
r-marques: Folded in `0d33463` — same finding as Kike's, independently.

Added the row, and also the thing that made the gap harmful rather than merely missing: the last row's "(and the SDK warns)" made silence read as impossible, so there is now a note under the table stating that the two silent rows are deliberately silent — a failed settle is an outcome the caller already sees on `success`, not a reporting fault. That removes the contradiction with the exported type comment you flagged.

src/a2a/paymentsRequestHandler.ts
r-marques: Folded in `0d33463` — same finding as Kike's.

Moved the block directly above `function resolveCreditsCharged`, leaving `DECIMAL_INTEGER_STRING`'s own block on the const. Confirmed by boundary rather than by eye: the blocks were `43-87` and `88-97` with the const at `98`, so the helper had no docblock at all. `eslint ./src` after the move: 0 errors, 5 pre-existing warnings.

tests/unit/a2a/payments-request-handler.test.ts
```diff
@@ -347,4 +366,387 @@ describe('PaymentsRequestHandler', () => {
...
+        expect(taskRef.metadata).not.toHaveProperty('txHash')
+        // The sibling key stays absent too — same response, same rule.
+        expect(event.metadata).not.toHaveProperty('creditsCharged')
+      })
+
+      test('a failed settle publishes no txHash on the streaming path either', async () => {
+        const { event } = await finalizeStreaming({
+          success: false,
+          errorReason: 'INSUFFICIENT',
+          transaction: '',
```
eruizgar91: **MEDIUM** — this half of the new `txHash` guard is not load-bearing.

Mutation-verified against this suite: deleting `response.success === true &&` from the streaming writer, leaving `if (response.transaction)`, keeps the run at **45 passed / 45 total**. The non-streaming twin is properly pinned — dropping `settled &&` at `src/a2a/paymentsRequestHandler.ts:1019` turns `a settle with no success field publishes no txHash` red.

The cause is the fixture: this case sets `transaction: ''`, so it only ever exercises the truthiness conjunct. The `success`-absent case — a malformed response, which is exactly when `transaction` cannot be trusted to be empty — is covered on the non-streaming path (`{ transaction: '0xdead' }`) and nowhere on the streaming one.

That matters more than a coverage gap normally would: the report on this round claims "making either `txHash` writer unconditional → 3 red", which is true and which I reproduced — but it is satisfied by the truthiness half alone, so the `success` half can be deleted later with a fully green suite.

Mirror the non-streaming test.each:

```ts
test.each([
  ['a failed settle', { success: false, errorReason: 'INSUFFICIENT', transaction: '' }],
  ['a settle with no success field', { transaction: '0xdead' } as any],
])('%s publishes no txHash on the streaming path either', async (_label, settle) => {
  const { event } = await finalizeStreaming({ network: 'eip155:84532', ...settle } as any)
  expect(event.metadata).not.toHaveProperty('txHash')
})
```

markdown/a2a-integration.md
```diff
@@ -240,6 +240,75 @@ eventBus.publish({
...
 ```
 
+### What the SDK adds to `metadata` after settlement
+
+Your executor sets `creditsUsed`. After the request settles, the SDK adds two more
+keys to the same `metadata` object:
+
+| key | meaning |
+| --- | --- |
+| `txHash` | The settlement transaction. A chain tx hash on the crypto rails; on fiat it is the **PSP transaction id** (a Stripe PaymentIntent, `pi_...`), so do not assume it is a hash. **Absent when the settle did not succeed** — the key is omitted rather than published blank, so its presence means a settlement really happened. |
```
eruizgar91: **MEDIUM** — the new sentence states a contract slightly stronger than the guard, and the gap lands on the fiat rail this same row describes.

The guard is `success === true && response.transaction` (`src/a2a/paymentsRequestHandler.ts:697`, `:1019`). So `txHash` is omitted on **two** conditions, not one — the second being a settle that succeeded but reported no `transaction`.

That is not hypothetical on fiat. Per this SDK's own wire documentation, [`facilitator-api.ts#L274-L282`](https://github.com/nevermined-io/payments/blob/0d334638d102d6c5b802ae72bdf955f59bdae1f0/src/x402/facilitator-api.ts#L272-L284), a fiat `pay-as-you-go` charge is referenced by **`orderTx`** (`pi_…`), while `transaction` is the blockchain hash — and the interface's own example gates on the disjunction `Boolean(settlement.orderTx || settlement.transaction)` precisely because either can be empty. A successful fiat pay-as-you-go settle therefore plausibly carries `transaction: ''` and now omits `txHash`.

A reader of *"Absent when the settle did not succeed"* concludes that charge failed. And because `resolveCreditsCharged` already omits `creditsCharged` on pay-as-you-go, that combination leaves **neither** key in `metadata` — on the legacy `payment-signature` path, where this file notes there are no receipts either, a real charge leaves no trace at all.

Note this is not an argument for reverting to `''` — `''` was equally free of the PSP id. It is that the sentence should keep only its sound direction. Suggested row text:

> **Absent unless the settle both succeeded and reported a `transaction`** — the key is omitted rather than published blank, so its presence means a settlement really happened. Absence does not by itself mean failure: fiat rails report the charge in the receipt's `orderTx`, not here.

Same clause in `src/a2a/types.ts:146-149`, which currently reads "ABSENT when the settle did not succeed".

eruizgar91 (commented):
**Round 4 — verifying the standing approval against code it never read.**

This bot approved at `c954a298`; the head is now `0d334638` and `dismiss_stale_reviews` is off, so that approval carried forward onto unread code. This round read the delta `c954a298..0d334638` (4 files, +125/−16) and re-derived every claim rather than trusting the report. **Two MEDIUM findings inline** — neither blocks, both are one-line fixes that belong here.

### The four round-3 findings: all genuinely closed

Verified by mutation against a symlinked `node_modules`, `jest --config ./tests/jest.config.json`:

| # | fix | mutation | result |
| --- | --- | --- | --- |
| 1 | `recordInBandSettlement` `=== false` → `!== true` | reverted to `=== false` | **1 red** — exactly `a settle with no success field ... payment-failed, not completed` |
| 2 | `txHash` omitted unless the settle succeeded | both writers made unconditional | **3 red** |
| 2 | — | *streaming writer only, `success === true` conjunct dropped* | **0 red** ← finding 1 inline |
| 2 | — | *non-streaming only, `settled` conjunct dropped* | **1 red** |
| 3 | failed-settle row + silent-rows note in the markdown | read against the code | accurate |
| 4 | `resolveCreditsCharged` docblock moved onto its function | `DECIMAL_INTEGER_STRING` hoisted above it | correct, docblock now adjacent |

Baseline reproduced: **49 suites / 680 passed / 1 skipped**, `tsc --noEmit -p tests/tsconfig.json` exit 0, working tree clean after every mutation.

### The hunt for a third surface: there isn't one

The recurring defect on this PR was a second writer the guard missed, so this round enumerated every publisher rather than spot-checking. In `src/a2a/`, payment state reaches a buyer through exactly three writers, and all three now apply the same rule:

- `recordInBandSettlement` (`:619`) — `!== true`. All **five** call sites (`:717`, `:732`, `:752`, `:1058`, `:1066`) funnel through it, and `x402A2AUtils.recordPaymentSuccess` / `recordPaymentFailure` have no other caller in `src/`.
- the streaming `event.metadata` writer (`:697-706`)
- the non-streaming `event.metadata` writer (`:1014-1024`)

`resolveCreditsCharged` is shared by the latter two. No fourth publisher exists.

### Considered and discarded

- **`x402.payment.receipts` still carries the raw `{success:false, creditsRedeemed:'7', transaction:''}` verbatim.** Correct — it is the wire receipt and `success` sits beside it as a discriminator. The defect was ever the derived summary, never the receipt.
- **MCP's streaming `_meta` gates `txHash` on `transaction` truthiness alone** (`src/mcp/core/paywall.ts:452`), not on `success`. Not reachable in practice (`transaction` is `''` on failure), different transport, outside this diff.
- **MCP reports `'0'` where A2A omits.** Explicitly a settled decision, documented on both sides (`paywall.ts:48-69`, `paymentsRequestHandler.ts:72-77`). Not drift.
- **Omission does not *delete* a pre-existing `event.metadata.txHash`**, so the "presence means a settlement happened" invariant assumes no executor writes the key itself. `PaymentMetadata` documents it as SDK-written, so that is off-contract; not raised.
- **Anything else in the delta.** Nothing. The four fixes account for every hunk.

---

## The direct question: keep the omission, or revert to `''`?

**Keep the omission.** Three reasons, in order of weight:

1. **It is not a new contract — this SDK already publishes it on the sibling transport.** `src/mcp/core/paywall.ts:276` and `:452` both write `...(creditsResult?.transaction && { txHash: creditsResult.transaction })`, and `tests/unit/mcp.test.ts:350` pins it as *"should not include txHash when transaction is empty"*. So the shape a consumer meets today, on MCP, is already `undefined` and never `''`. Reverting A2A to `''` would not preserve an existing contract — it would create the divergence, and the MCP file's own comment (`paywall.ts:285-291`) states the rule being broken: *"the two protocols must not tell a buyer different things about the same money."*
2. **`''` is the same defect as `creditsCharged: 0`, which this PR removed.** A blank id beside a completed task reads as "settled, reference unavailable" — an assertion about a settle that never happened. Fixing one key and leaving the other is the inconsistency.
3. **The blast radius is bounded by a type that was already optional.** `txHash?: string` in `PaymentMetadata` predates this PR, so every type-correct consumer already had to handle `undefined`; a consumer dereferencing it unconditionally was already unsound. And at runtime `''` and `undefined` are both falsy, so `if (meta.txHash)` is unchanged — only `'txHash' in meta` / `!== undefined` differ, and those are precisely the readers the omission serves.

Finding 2 below is not an argument against the omission; it is an argument that the *sentence describing it* is stronger than the guard.

---

**Head SHA read and posted against: `0d334638d102d6c5b802ae72bdf955f59bdae1f0`** — re-checked immediately before posting, unchanged from `meta.env`.

Verdict-less comment review (bot mode). Severity ladder: 0 CRITICAL, 0 HIGH, 2 MEDIUM, 0 LOW.

eruizgar91 (approved):
🟡 **Approved with comments** — no blockers; 2 should-fix, 0 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/439#pullrequestreview-5209799792

tests/unit/a2a/payments-request-handler.test.ts
r-marques: Folded in `123fcbf`, and this one is on me twice over: the guard was half-pinned, and my report's "either writer unconditional → 3 red" was true while proving less than it read — satisfied by the truthiness half alone. Count is not enough; which test goes red is the evidence.

Took your `test.each` as written. Re-ran your mutation (streaming writer reduced to `if (response.transaction)`) and recorded the shape this time rather than a number:

```
✓ a failed settle publishes no txHash on the streaming path either
✕ a settle with no success field publishes no txHash on the streaming path either
```

The new row is the only thing that moves, and the old row staying green is the point — it never could see the `success` conjunct. 49 suites / 681 passed unmutated.

markdown/a2a-integration.md
r-marques: Folded in `123fcbf`, with your row text. You are right that the sentence claimed the wrong direction: the guard is `success && transaction`, and a successful fiat pay-as-you-go settle is referenced by `orderTx` with an empty `transaction`, so it omits `txHash` too — and with `creditsCharged` also omitted on pay-as-you-go, that combination left no key at all and a reader taking "absent ⇒ did not succeed" would call a real charge a failure.

Both places now keep only the sound direction — presence means a settlement really happened; absence does not by itself mean failure, fiat reports the charge in the receipt's `orderTx`. Markdown field table and the `types.ts` JSDoc on `txHash`. Not reverted to `''`, for the reason you gave: the blank carried no more of the PSP id than the absence does.

PR #441, opened 2026-09-10
fix(types): declare PlanBalance.balance as the string the wire actually sends (#440)

### LINKED ISSUE #440: PlanBalance.balance is declared `bigint` but the wire sends a string — comparisons throw in code that typechecks clean
Found while reviewing #433 / PR #437. Same shape as #438 — a type asserted over unvalidated JSON that nothing forces anyone to reconcile — but this one is **public API a seller reads**, and it throws rather than yielding `undefined`.

## Three sources disagree about one field

`PlanBalance.balance` is declared **`bigint`**.

| Source | Says |
| --- | --- |
| The SDK type | `bigint` |
| `getPlanBalance`'s own JSDoc example (`plans-api.ts:611`) | `balance: '100'` — a **string** |
| The API (`protocol.controller.ts:422`, the endpoint's own OpenAPI example) | `balance: '950'` — a **string** |

**`PlanBalance` is constructed in `src/` in exactly zero places.** It only ever arrives from `response.json()` — `getPlanBalance` (`plans-api.ts:617`) ends in a bare `return response.json()`, and there is no `BigInt(...)` coercion of `balance` anywhere in `src/`. JSON has no bigint, so the value can never be one; the wire sends a quoted string, so at runtime it is a **string**.

## What it costs a seller

`req.paymentContext` is public API — `middleware.ts:132`, *"Available as `req.paymentContext` in route handlers"* — and `paymentContext.agentRequest` is assigned straight from `verification.agentRequest` (`middleware.ts:974` MPP, `:1360` x402), which is `await response.json()`.

So in code that **typechecks clean**:

```ts
typeof ctx.agentRequest.balance.balance === 'bigint'   // false
ctx.agentRequest.balance.balance > 100n                 // TypeError: Cannot mix BigInt and other types
```

The first silently takes the wrong branch; the second throws at runtime. Both are the natural thing to write against a field declared `bigint`.

## The repo already solves this in the other direction

`jsonReplacer` (`common/helper.ts:36`) narrows bigints to strings on **outbound** request bodies (`base-payments.ts:249`, `:277`). So the boundary problem is understood — there is simply no inbound counterpart.

## The fix

**Make the declared type match what the deserializer actually produces.** This is not a preference: the wire sends a string, so the type must say so, or the SDK must coerce at the boundary. If `bigint` is genuinely the shape we want to hand consumers, then coerce it explicitly where the response is parsed — declaring it and hoping is what produced this.

Whichever way it lands, check the other fields on `PlanBalance` and its siblings at the same time: this is one instance of "a type asserted over `response.json()`", and the same audit applies to anything else reached that way.

⚠️ Note this is a **behaviour-visible correction for consumers who typecheck against it** — but they are already broken at runtime today. Changing the type makes existing breakage visible rather than introducing new breakage.

## Related

- #438 / PR #439 — `executeRedemption` returned `Promise<any>` and cast, hiding two dead field reads. Same class, caught the same way.
- #433 / PR #437 — the test-double half. That PR's `jsonSafe` helper in `middleware-hooks-parity` exists **solely** to compensate for this fiction: the type obliged the fixture to carry a bigint the wire never sends. If this issue is fixed, that helper becomes removable.

r-marques:
Closes #440.

## The defect

`PlanBalance.balance` was declared **`bigint`**. It is never a bigint.

`PlanBalance` is constructed in `src/` in **exactly zero places** — it only ever arrives from `getPlanBalance` (`plans-api.ts:637`), which ends in a bare `return response.json()`. JSON has no bigint, and there is no `BigInt(...)` coercion of `balance` anywhere in `src/`. The wire sends a quoted string, so at runtime it is a **string**.

This is public API: `PlanBalance` reaches sellers on `req.paymentContext.agentRequest.balance` (`middleware.ts` assigns `paymentContext.agentRequest` straight from `await response.json()`).

So in code that **typechecks clean**:

```ts
typeof ctx.agentRequest.balance.balance === 'bigint'   // false — silently wrong branch
ctx.agentRequest.balance.balance > 100n                 // TypeError: Cannot mix BigInt and other types
```

Both are the natural thing to write against a field declared `bigint`.

## The type was the only source that said bigint

Every other source already said string — including the ones we publish:

| Source | Says |
| --- | --- |
| `PlanBalance.balance` (before this PR) | `bigint` |
| `getPlanBalance`'s own JSDoc example | `balance: '100'` |
| The API's OpenAPI example | `balance: '950'` |
| `markdown/payments-and-balance.md:45` | `balance: string  // Available credits as string` |
| …and its examples, e.g. `:222` | `BigInt(balance.balance) > 0n` |
| `cli/test/unit/plans.test.ts:109` | `BigInt(parsed.balance)` |

The hand-written docs therefore need **no update** — they have documented the real shape all along. The declaration was the outlier.

## What changes

Type only. The runtime value is untouched (`'950'` before, `'950'` after), so nothing that works today breaks. Consumers who typecheck against it are already broken at runtime; this makes that visible rather than introducing it.

The docblock records *why* it is a string — that it is what the deserializer produces, not a preference — plus the `BigInt(...)` conversion for arithmetic and a warning not to route it through `Number` (the value can exceed `Number.MAX_SAFE_INTEGER`).

`jsonReplacer` (`common/helper.ts`) already narrows bigints to strings on **outbound** bodies. This is the inbound side of the same boundary.

## Verification

- `pnpm build` (`tsc`) clean.
- `pnpm test:unit` — **635 passed**, 1 skipped, 48 suites. (Runner: jest via `pnpm test:unit`.)
- In-tree consumers checked by hand, since they use the SDK directly: `openclaw/src/index.ts:161` `Number(b.balance)`, `:414` a template literal, `openclaw/src/tools.ts:57` `balance.balance.toString()`, `cli/test/unit/plans.test.ts:109` `BigInt(parsed.balance)` — all valid on a string. Their existing `tsc` errors are about `onchainMirror`, `X402TokenOptions`, `getAgents` and `embed`; **none** mentions `balance`, so this adds no new one.

## Follow-on

PR #437's `jsonSafe` helper in `middleware-hooks-parity` exists **solely** to let a fixture carry the bigint this fiction demanded. It becomes removable once this lands — noted there rather than done here, to keep this PR type-only.

### DIFF
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -232,7 +232,30 @@ export interface PlanBalance {
   planName: string
   planType: string
   holderAddress: Address
-  balance: bigint
+  /**
+   * Credit balance as a DECIMAL STRING, e.g. `'950'` — not a `bigint`.
+   *
+   * ⚠️ It is a string because that is what the deserializer produces, not as a
+   * preference. `PlanBalance` is constructed nowhere in `src/`: it only ever
+   * arrives from `getPlanBalance`, which ends in a bare `return response.json()`,
+   * and JSON has no bigint. It was declared `bigint` (#440) while the API sent a
+   * quoted string, so `balance > 100n` threw `Cannot mix BigInt and other types`
+   * in code that typechecked clean, and `typeof balance === 'bigint'` silently
+   * took the wrong branch. This reaches sellers as public API — `PlanBalance` is
+   * on `req.paymentContext.agentRequest.balance`.
+   *
+   * The value can exceed `Number.MAX_SAFE_INTEGER`, so do NOT route it through
+   * `Number`. For arithmetic, convert explicitly:
+   *
+   * ```ts
+   * BigInt(ctx.agentRequest.balance.balance) > 100n
+   * ```
+   *
+   * The outbound direction already has its counterpart: `jsonReplacer`
+   * (`common/helper.ts`) narrows bigints to strings on request bodies. This is
+   * the inbound side of the same boundary.
+   */
+  balance: string
   creditsContract: Address
   isSubscriber: boolean
   pricePerCredit: number
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -238,7 +238,7 @@ export interface PlanBalance {
    * ⚠️ It is a string because that is what the deserializer produces, not as a
    * preference. `PlanBalance` is constructed nowhere in `src/`: it only ever
    * arrives from `getPlanBalance`, which ends in a bare `return response.json()`,
-   * and JSON has no bigint. It was declared `bigint` (#440) while the API sent a
+   * and JSON has no bigint. It had been declared `bigint` since 2025-05-26 while the API sent a
    * quoted string, so `balance > 100n` threw `Cannot mix BigInt and other types`
    * in code that typechecked clean, and `typeof balance === 'bigint'` silently
    * took the wrong branch. This reaches sellers as public API — `PlanBalance` is
--- a/cli/test/helpers/mock-payments.ts
+++ b/cli/test/helpers/mock-payments.ts
@@ -14,7 +14,8 @@ export interface MockPlanBalance {
   planName: string
   planType: string
   holderAddress: string
-  balance: bigint
+  /** Decimal string — the wire shape. See PlanBalance.balance (#440). */
+  balance: string
   creditsContract: string
   isSubscriber: boolean
   pricePerCredit: number
@@ -72,7 +73,7 @@ export class MockPlansAPI {
       planName: plan.name,
       planType: plan.planType,
       holderAddress: '0x1234567890123456789012345678901234567890',
-      balance: BigInt(1000),
+      balance: '1000',
       creditsContract: '0x0987654321098765432109876543210987654321',
       isSubscriber: true,
       pricePerCredit: 0.01,
--- a/openclaw/tests/plugin.test.ts
+++ b/openclaw/tests/plugin.test.ts
@@ -16,7 +16,7 @@ function createMockPayments() {
         planName: 'Test Plan',
         planType: 'credits',
         holderAddress: '0x1234',
-        balance: 100n,
+        balance: '100',
         creditsContract: '0xabcd',
         isSubscriber: true,
         pricePerCredit: 1,


eruizgar91:
👀 Reviewing this now.

src/common/types.ts
```diff
@@ -232,7 +232,30 @@ export interface PlanBalance {
...
+   *
+   * ```ts
+   * BigInt(ctx.agentRequest.balance.balance) > 100n
+   * ```
+   *
+   * The outbound direction already has its counterpart: `jsonReplacer`
+   * (`common/helper.ts`) narrows bigints to strings on request bodies. This is
+   * the inbound side of the same boundary.
+   */
+  balance: string
```
eruizgar91, line 258: **MEDIUM** — With this landed, the two **in-tree consumer fixtures** become the only remaining sources that still contradict the wire:

- [`cli/test/helpers/mock-payments.ts#L12-L21`](https://github.com/nevermined-io/payments/blob/e4037fe9e21cde67b9fcf14fca5e8732c731d8d9/cli/test/helpers/mock-payments.ts#L12-L21) — `MockPlanBalance.balance: bigint`
- [`cli/test/helpers/mock-payments.ts#L70-L80`](https://github.com/nevermined-io/payments/blob/e4037fe9e21cde67b9fcf14fca5e8732c731d8d9/cli/test/helpers/mock-payments.ts#L70-L80) — returns `balance: BigInt(1000)`
- [`openclaw/tests/plugin.test.ts#L14-L24`](https://github.com/nevermined-io/payments/blob/e4037fe9e21cde67b9fcf14fca5e8732c731d8d9/openclaw/tests/plugin.test.ts#L14-L24) — `getPlanBalance` resolves `balance: 100n`

Neither breaks CI, and that is precisely the problem. `MockPlanBalance` is a standalone interface never assigned to `PlanBalance`, and the openclaw mock is typed `Promise<unknown>` — so `cli-sync-and-test.yml` (which triggers on `src/**` and runs the CLI unit tests against this very build) stays green while the only test covering `plans get-plan-balance`, [`cli/test/unit/plans.test.ts#L91-L111`](https://github.com/nevermined-io/payments/blob/e4037fe9e21cde67b9fcf14fca5e8732c731d8d9/cli/test/unit/plans.test.ts#L91-L111), exercises a shape the API cannot produce. It drives the CLI's bigint replacer (`cli/src/utils/output-formatter.ts:19`) down a branch that is now dead for this field, and never covers the string the SDK actually hands it. A regression back to #440 would not be caught by the one test that exists to catch it.

This is the rule `CLAUDE.md` calls out explicitly:

> **CRITICAL:** When changing public SDK interfaces (function signatures, types, options), you MUST also update `openclaw/` and `cli/` source files, tests, and mocks.

The PR body enumerates its hand-check of in-tree consumers and cites `cli/test/unit/plans.test.ts:109` as evidence the CLI already treated this as a string — but the fixture feeding that assertion is the bigint.

**Fix — three lines, and not one assertion needs touching.** Set `MockPlanBalance.balance: string`, return `balance: '1000'`, and make the openclaw mock `balance: '100'`. All four existing assertions are unaffected, checked individually:

- `expect(logs).toContain('1000')` — the table formatter renders via `String(value)`.
- `expect(BigInt(parsed.balance)).toBe(BigInt(1000))` — `parsed.balance` is *already* a JSON string today, because the output formatter's replacer stringifies the bigint before `JSON.parse` ever sees it. Byte-identical input before and after.
- `openclaw/tests/plugin.test.ts:397` already expects `balance: '100'` — `tools.ts:57` calls `.toString()`, which is a no-op on a string.
- `openclaw/tests/plugin.test.ts:1076` expects `'balance: 100 credits'` — `index.ts:161` calls `Number(...)`, and `Number('100') === Number(100n)`.

eruizgar91 (commented):
**Reviewed at `e4037fe`** — one file, `src/common/types.ts`: `PlanBalance.balance` goes `bigint` → `string`, plus a TSDoc block explaining why. Findings are inline; **1 MEDIUM**, no CRITICAL or HIGH.

The premise holds, and on stronger evidence than the PR body claims.

### Verified

- **The wire really does send a string — confirmed in backend source, not just from an OpenAPI example.** `protocol.service.ts#_getBalanceFromPlan` returns a genuine `bigint` (`isSubscriber: balance > 0n` only typechecks because it is one), and `apps/api/src/main.ts:240` registers `BigIntToStringInterceptor` globally, narrowing every outbound bigint to a decimal string. There is no `BigInt.prototype.toJSON` polyfill anywhere in `nvm-monorepo`, so that interceptor is the only reason the endpoint serializes at all — `bigint` was never reachable on the client, on any release.
- **No `src/` consumer changes behaviour.** Nothing in `src/` constructs a `PlanBalance`. The only reads of the nested value are `api/observability-api/utils.ts` (other fields only) and `a2a/server.ts:258` (`isSubscriber`). No arithmetic, comparison, `typeof` or equality check on `balance` anywhere in `src/`.
- **The docblock's claims are accurate.** `jsonReplacer` (`src/common/helper.ts:36-38`) does narrow bigint → string, so framing this as the inbound half of that boundary is correct.
- **`markdown/` genuinely needs no update**, as the PR argues. `markdown/payments-and-balance.md:44` already declares `balance: string` and `:222` already shows `BigInt(balance.balance)`. A repo-wide sweep found **no** markdown claiming bigint for this field.

### Considered and discarded

- **Stale committed TypeDoc.** `docs/` is tracked and `docs/interfaces/PlanBalance.html` still renders `bigint`. Not this PR's job — `prepare-release.yml:59` runs `pnpm run doc` and commits `docs/` during release prep.
- **`openclaw/src/index.ts:161` does `Number(b.balance)`**, which the new docblock explicitly warns against. Real, but pre-existing, independent of the declared type, and only lossy above `Number.MAX_SAFE_INTEGER` (~9e15 credits). Below the bar; not worth holding this PR for.
- **Breaking for downstream TS consumers.** Intentional and argued — anyone typed against `bigint` is already broken at runtime, so this surfaces a defect rather than introducing one. Worth ensuring the release carrying it is at least a `minor`; versioning is owned by `prepare-release.yml`, so there is nothing to change in this PR.
- **No version or CHANGELOG edit here** — correct for this repo, both are generated by the release workflow.
- **Docblock wording.** "It was declared `bigint` (#440)" can be misread as "#440 introduced the bigint"; the declaration actually predates it (`2837bc5`, 2025-05-26) and #440 is the report this closes. The sentence does not literally assert origin, so it is a wording nit, not an error — tighten it if you are touching the file anyway.

eruizgar91 (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 0 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/441#pullrequestreview-5176426195

src/common/types.ts
r-marques: Folded in `17e076c` — all three fixtures, exactly as prescribed, and you were right that not one assertion needed touching. Verified by running them rather than reasoning: CLI 33 unit + 2 integration suites pass, openclaw 67 pass, SDK 635 pass, `tsc` clean.

And you were right about the rule too — `CLAUDE.md:220` says it verbatim. I checked the consumers' *source* (`openclaw/src/index.ts`, `tools.ts`, and a grep of `cli/`) and never looked at their *fixtures*. That is the gap.

**One correction, and it strengthens your finding rather than softening it.** `cli/test/unit/plans.test.ts` is not "the only test covering `plans get-plan-balance`" — it provides **zero** coverage, on two independent counts:

1. It is in `cli/jest.config`'s `testPathIgnorePatterns`, alongside `config.test.ts` and `agents.test.ts`. It cannot run under any script — not `test:unit`, not `test:unit:all`, not `test:all`.
2. Forced to run anyway, it does not compile: it imports `src/commands/plans/list.js`, `get.js` and `balance.js`, which no longer exist. The CLI moved to generated verb-prefixed files, so those are now `get-plans.ts`, `get-plan.ts` and `get-plan-balance.ts`.

Also worth knowing: `cli-sync-and-test.yml` runs `pnpm test:unit`, which is a **hardcoded four-file list** (`simple`, `output-formatter-basic`, `x402`, `generator-complex-type`) — so even without the ignore list, that test was never in CI's path.

So the regression you describe would not have been caught *in any case*, which makes the fixture fix necessary rather than merely tidy: the fixtures were the only thing left asserting the wrong shape, and nothing was checking them.

I have **not** folded the dead tests — that is three files, one of which (`config.test.ts`) has a genuine assertion failure rather than a path problem, and repointing them means working through the generated-command harness. Filed as nevermined-io/payments#446 with the specifics, per-file. Flagging it here before resolving so you can push back if you would rather it rode along.

The `#440` wording nit is fixed too — now "declared `bigint` since 2025-05-26", so it cannot read as #440 having introduced it.

PR #442, opened 2026-09-10
fix(mcp): make McpConfig.planId optional, matching the runtime and the docs

r-marques:
## Three surfaces disagreed about one field

Whether an MCP server must set a plan at **config** time:

| Source | Says |
| --- | --- |
| The type — `McpConfig.planId: string` | **required** |
| `withPaywall`'s docblock | *"a planId must be resolvable (**per-tool `options.planId`** or a server-level one set via `configure()`/`start()`)"* |
| The runtime — `paywall.ts` | `options?.planId ?? this.config.planId`, refusing only when **both** are absent |

So `payments.mcp.configure({ agentId })` was a **compile error** for a server that sets its plan per tool — a configuration the runtime supports and the docblock documents. The type was the only one of the three that forbade it.

`planId` becomes optional. Set it when one plan covers the whole server; omit it when every paywalled handler passes its own.

## Also: three stale `// required` markers

`CreateRouterOptions.planId` has been `planId?: string` all along, with its own docblock saying *"falls back to `configure()`"* — so the `// required` on the `createRouter` (×2) and `createApp` examples was already wrong, independently of this change. Corrected.

The `start()` example **keeps** `// required`: `McpServerConfig.planId` genuinely is `string`. That is a different type and a defensible boundary for the batteries-included path, so I have not widened the change to it — flagging it here rather than deciding it silently. Say the word if you want them unified.

## Tests pin BOTH guards — they are not the same guard

There are **two** planId guards, in two files, reading two **different** config objects. A test exercising one proves nothing about the other:

- **`withPaywall`** (`mcp/index.ts`) — fails fast at **registration**, off the module-level `extendedConfig` that `configure()` merges into.
- **`PaywallDecorator.createWrappedHandler`** (`mcp/core/paywall.ts`) — fails at **call** time, off the decorator's own `this.config`.

Each has a **negative control** asserting that "optional" did not become "skippable", plus a test that a later partial `configure()` (e.g. one that only renames the server) does not blank a plan already set — `planId: options.planId || this.config.planId`.

### Mutation-tested, both independently

| Mutant | Result |
| --- | --- |
| delete the `withPaywall` registration guard | kills **only** `…and with NEITHER, registration is refused` |
| delete the decorator call-time guard | kills **only** `…and with NEITHER, the call is refused rather than served unpriced` |
| unmutated control | 640 passed, 49 suites |

Neither mutant kills the other's control, so the two are genuinely pinning different code — not overlapping and masking each other.

## A note on what is NOT asserted here

The **type** half cannot be tested in this file. Jest in this repo does not typecheck (`tests/**` is typechecked by nothing — that is exactly #433 / PR #437), so a `// @ts-expect-error` here would prove nothing whether it held or not. `tsc` over `src/` pins the signature; these tests pin the behaviour the signature was contradicting. Called out in the test file's header so the next reader does not mistake green for type coverage.

## Verification

- `pnpm build` (`tsc`) clean.
- `pnpm test:unit` — **640 passed**, 1 skipped, 49 suites (5 new). Runner: jest via `pnpm test:unit`.

Decided in-session with the repo owner; no separate issue was filed.

### DIFF
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -50,7 +50,7 @@
  * // Create OAuth router
  * const router = payments.mcp.createRouter({
  *   baseUrl: 'http://localhost:5001',
- *   planId: 'plan_123', // required
+ *   planId: 'plan_123', // optional — falls back to configure()
  *   agentId: 'agent_123', // optional
  *   serverName: 'my-mcp-server'
  * })
@@ -258,7 +258,7 @@ export function buildMcpIntegration(paymentsService: Payments) {
    * @example
    * ```typescript
    * payments.mcp.configure({
-   *   planId: 'plan_123', // required
+   *   planId: 'plan_123', // optional — omit it if every tool sets its own
    *   serverName: 'my-mcp-server',
    *   agentId: 'agent_123', // optional
    *   baseUrl: 'http://localhost:5001',
@@ -430,7 +430,7 @@ export function buildMcpIntegration(paymentsService: Payments) {
    * ```typescript
    * const router = payments.mcp.createRouter({
    *   baseUrl: 'http://localhost:5001',
-   *   planId: 'plan_123', // required
+   *   planId: 'plan_123', // optional — falls back to configure()
    *   serverName: 'my-mcp-server',
    *   tools: ['hello_world']
    * })
@@ -472,7 +472,7 @@ export function buildMcpIntegration(paymentsService: Payments) {
    * ```typescript
    * const app = payments.mcp.createApp({
    *   baseUrl: 'http://localhost:5001',
-   *   planId: 'plan_123', // required
+   *   planId: 'plan_123', // optional — falls back to configure()
    *   serverName: 'my-mcp-server'
    * })
    *
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -17,7 +17,7 @@
  *
  * // Configure MCP integration
  * payments.mcp.configure({
- *   planId: process.env.NVM_PLAN_ID!, // required
+ *   planId: process.env.NVM_PLAN_ID!, // optional — omit it if every tool sets its own
  *   agentId: process.env.NVM_AGENT_ID, // optional
  *   serverName: 'my-mcp-server'
  * })
@@ -34,7 +34,7 @@
  * // Start a managed server with all OAuth endpoints
  * const { baseUrl, stop } = await payments.mcp.startServer({
  *   port: 5001,
- *   planId: process.env.NVM_PLAN_ID!, // required
+ *   planId: process.env.NVM_PLAN_ID!, // optional — falls back to configure()
  *   agentId: process.env.NVM_AGENT_ID, // optional
  *   serverName: 'my-mcp-server',
  *   tools: ['hello_world']
@@ -513,7 +513,7 @@ export function buildMcpIntegration(paymentsService: Payments) {
    * ```typescript
    * const { baseUrl, stop } = await payments.mcp.startServer({
    *   port: 5001,
-   *   planId: process.env.NVM_PLAN_ID!, // required
+   *   planId: process.env.NVM_PLAN_ID!, // optional — falls back to configure()
    *   serverName: 'my-mcp-server',
    *   tools: ['hello_world', 'weather']
    * })
--- a/src/mcp/types/paywall.types.ts
+++ b/src/mcp/types/paywall.types.ts
@@ -96,9 +96,25 @@ export interface PaywallContext {
  * MCP integration configuration
  */
 export interface McpConfig {
-  /** Plan ID the server charges against (required). The facilitator is
-   * plan-centric: verify/settle resolve everything from the plan + token. */
-  planId: string
+  /**
+   * Plan ID the server charges against. The facilitator is plan-centric:
+   * verify/settle resolve everything from the plan + token.
+   *
+   * Optional because a planId only has to be **resolvable by the time a handler
+   * runs**, and a per-tool `options.planId` satisfies that on its own:
+   * `PaywallDecorator` resolves `options?.planId ?? this.config.planId` and
+   * refuses only when both are absent (`Server misconfiguration: missing
+   * planId`). `withPaywall`'s own docblock has always documented that form.
+   *
+   * It was declared required anyway, which made `configure({ agentId })` a
+   * compile error for a server that sets its plan per tool — a configuration
+   * the runtime supports and the docs describe. The type was the only one of the
+   * three that forbade it.
+   *
+   * Set it here when one plan covers the whole server; omit it when every
+   * paywalled handler passes its own.
+   */
+  planId?: string
   /** Agent ID — optional, informational/observability only. */
   agentId?: string
   serverName?: string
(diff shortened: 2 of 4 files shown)


eruizgar91:
👀 Reviewing this now.

src/mcp/types/paywall.types.ts
```diff
@@ -96,9 +96,25 @@ export interface PaywallContext {
...
+   *
+   * It was declared required anyway, which made `configure({ agentId })` a
+   * compile error for a server that sets its plan per tool — a configuration
+   * the runtime supports and the docs describe. The type was the only one of the
+   * three that forbade it.
+   *
+   * Set it here when one plan covers the whole server; omit it when every
+   * paywalled handler passes its own.
+   */
+  planId?: string
```
eruizgar91, line 117: **MEDIUM** — `markdown/mcp-integration.md` still asserts the contract this line reverses, in two places. The PR's own framing is that three surfaces disagreed about `planId`; making the type optional leaves a fourth — the published guide — stating the opposite:

- [`markdown/mcp-integration.md:54`](https://github.com/nevermined-io/payments/blob/875fe571b87126744ccece8726bd5e3583899632/markdown/mcp-integration.md#L52-L58) — "**`planId` is required; `agentId` is optional.**"
- [`markdown/mcp-integration.md:404`](https://github.com/nevermined-io/payments/blob/875fe571b87126744ccece8726bd5e3583899632/markdown/mcp-integration.md#L403-L405) — "A server-level `planId` (set via `configure`/`start`) is required; set this only to charge a different plan for this handler."

Line 54 was written by `1dee621` (#391) — the very commit that made `McpConfig.planId` required. It is that change's documentation half, so it should travel back with this one. Line 404 is the worse of the two: it now actively tells readers that the per-tool-only configuration this PR exists to unblock is illegal.

CLAUDE.md, "After Modifying Source Files": *"If changing public interfaces (function signatures, options, types, response fields), update the corresponding documentation in `markdown/` to reflect the changes."* The same section notes `generate-docs.sh` *"only validates structure — it does NOT auto-update code examples"*, so the on-merge `update-docs.yml` run will not repair this, and `publish-docs.yml` ships it to the public docs on the next tag.

Precedent: this exact class was raised and fixed on #391 — *"`markdown/mcp-integration.md` was updated but `README.md` was missed"* ([review](https://github.com/nevermined-io/payments/pull/391#pullrequestreview-4517438155)). #391 carried 27 lines of `markdown/mcp-integration.md` alongside the type change.

Two sentences in one adjacent file — small enough to land here rather than as a follow-up.

src/mcp/index.ts
```diff
@@ -50,7 +50,7 @@
  * // Create OAuth router
  * const router = payments.mcp.createRouter({
  *   baseUrl: 'http://localhost:5001',
- *   planId: 'plan_123', // required
+ *   planId: 'plan_123', // optional — falls back to configure()
```
eruizgar91, line 53: **LOW** — Re: `src/mcp/index.ts:20`. You corrected this marker for `configure()` at line 261, but the module-header copy of the same example still reads `// required`:

https://github.com/nevermined-io/payments/blob/875fe571b87126744ccece8726bd5e3583899632/src/mcp/index.ts#L18-L22

`configure()` takes `ExtendedMcpConfig extends McpConfig` — precisely the field this PR makes optional — so after this change the file contradicts itself about the same function, 241 lines apart. One line, reusing the wording you already chose at line 261: `// optional — omit it if every tool sets its own`.

While you are in the file: lines 37 and 516 mark `planId` `// required` on `startServer()`, whose `StartServerOptions.planId` has been `?: string` all along ([`src/mcp/index.ts:208`](https://github.com/nevermined-io/payments/blob/875fe571b87126744ccece8726bd5e3583899632/src/mcp/index.ts#L206-L210)). Those are pre-existing and not caused by this PR, but they are the same stale-marker class you set out to clean up and are one line each. Line 666 is correct as-is — `McpServerConfig.planId` genuinely is `string`, exactly as your description states.

eruizgar91 (commented):
### Code review

`McpConfig.planId` goes required -> optional, four docblock markers are corrected, and a new unit-test file pins both planId guards. Two findings, both inline; neither blocks the type change itself.

**The relaxation is sound.** `git diff` over `src/mcp/core/paywall.ts` is empty and the eager `withPaywall` throw in `src/mcp/index.ts` is untouched, so the fail-fast behaviour `1dee621` (#391) deliberately restored still fires: registering a handler with neither a per-tool nor a server-level plan still throws `Server misconfiguration: missing planId`. #391's own commit message already described that guard as "per-tool option **or** server config" — the type was the outlier, not the defence. `CreateRouterOptions.planId`, `StartServerOptions.planId` and `ExtendedMcpConfig` were already optional, so `McpConfig` was the last holdout and widening it opens no new gap.

**Checked and discarded:**

- *Reintroducing the defect #391 fixed* — no. The mechanism #391 restored was the runtime guard, not the TS type, and the guard is byte-for-byte unchanged.
- *The new tests being vacuous* — no. Neutering `if (!configuredPlanId)` in `src/mcp/core/paywall.ts` makes the negative control ("…with NEITHER, the call is refused rather than served unpriced") fail, so it is pinning real behaviour. The file's header is also right that the type half cannot be asserted here.
- *CLAUDE.md's in-tree consumer rule* ("you MUST also update `openclaw/` and `cli/`") — does not fire. Neither tree references `McpConfig` or any MCP surface, and required -> optional cannot break a call site. `CLI Sync Check` and `OpenClaw Plugin Check` are both green.
- *`McpServerConfig.planId` staying `string`* — deliberate and defensible, as the description argues: the `registerTool`/`start` path has no per-tool override. Not widened, correctly so.

eruizgar91 (approved):

Inline review: https://github.com/nevermined-io/payments/pull/442#pullrequestreview-5176449874

src/mcp/types/paywall.types.ts
r-marques: Folded in `3c57d7b`. You are right that it left a fourth surface — and the provenance you spotted is the clinching detail: `markdown/mcp-integration.md:54` was written by `1dee621` (#391), the commit that made `McpConfig.planId` required. It is that change's documentation, so it reverses with it.

Both places now state the real contract — a plan must be resolvable by the time a handler is **registered**, and a per-tool `planId` satisfies that on its own — rather than asserting a server-level one is required. The per-handler table row says the same and names the error you get with neither.

src/mcp/index.ts
r-marques: Folded in `3c57d7b`, and I audited every marker in the file rather than patching the two you named — which confirmed you were right about more than the header:

| line | documents | option type | verdict |
| --- | --- | --- | --- |
| `:20` | `configure()` | `ExtendedMcpConfig extends McpConfig` | now optional — **fixed** |
| `:37` | `startServer()` | `StartServerOptions.planId?: string` | **always** was optional — fixed |
| `:516` | `startServer()` | same | **always** was optional — fixed |
| `:666` | `start()` | `McpServerConfig.planId: string` | genuinely required — **left alone** |

So the two `startServer` markers were wrong before this PR and independent of it, exactly as you said.

`:666` is the one that stays: `start()` takes `McpServerConfig`, a different type whose `planId` really is non-optional. I deliberately did not widen the change to it and called that out in the PR body — say the word if you would rather the two were unified, but that is a behaviour change to `start()` rather than a doc fix.

PR #443, opened 2026-09-10
fix(mcp): stop publishing the requested burn as creditsRedeemed

r-marques:
Follow-up to #439, on the protocol that PR was scoped out of. Raised by the review panel on #439 and confirmed against the code.

## The defect

`src/mcp/core/paywall.ts` published, at **both** the non-streaming (`:245`) and streaming (`:408`) sites:

```ts
creditsRedeemed: settlement?.success
  ? (settlement.creditsRedeemed ?? credits.toString())   // ⚠️
  : '0',
```

When a **successful** settle reported no figure, `?? credits.toString()` substituted the credits the request **asked** to burn and published them under the name of the figure the settle **reported**. That is a number the facilitator never sent, labelled as one it did.

It is exactly the substitution #439 removed from the A2A handler. Same repo, same money question, opposite answer depending on which protocol the buyer happened to use.

## The fix

A successful settle that reports no figure now **omits the key**.

Omission is the honest answer, and nothing is lost — `remainingBalance`, `orderTx`, `billingModel` and the complete spec receipt under `x402/payment-response` are all still emitted, so a consumer that wants to reason about the charge has everything it needs.

The `credits` parameter of `wrapAsyncIterable` existed **only** to feed that fallback, so it is removed rather than left dangling as an unused argument.

## ⚠️ An existing test was asserting the defect

`tests/unit/mcp.test.ts` used a bare `{ success: true }` settle double — no `creditsRedeemed` — with a tool asking to burn `3n`, and asserted:

```ts
expect(out._meta['nevermined/credits'].creditsRedeemed).toBe('3')
```

`'3'` could only ever come from `credits.toString()`. The test read as a safety property while pinning the bug, and it is why this went unnoticed. It now asserts the key is **absent**, with the reason recorded inline so it is not "fixed" back.

**Added a positive control**, because "the key is absent" is satisfied just as well by never emitting it at all: a settle reporting `'7'` against a tool asking for `3n` still publishes `'7'`. The two numbers differ deliberately so it cannot pass by reading the wrong one.

I also wrote and then **deleted** a third test for the failed-settle `'0'` branch — it was vacuous. A failed settle throws `SettlementFailedError` before `_meta` is built, so the assertion never ran; I confirmed that by mutating the expected value to an impossible one and watching it still pass. Better no test than one that proves nothing.

## Mutation-tested

| mutant | result |
| --- | --- |
| restore `?? credits.toString()` | kills `should add metadata to result after successful redemption` |
| unmutated control | 636 passed, 48 suites |

## Verification

`pnpm build` (`tsc`) clean · `pnpm test:unit` **636 passed**, 1 skipped, 48 suites · `pnpm lint` 0 errors, 3 pre-existing warnings. Runner: jest via `pnpm test:unit`.

Decided in-session with the repo owner, who chose a separate PR over folding it into #439 so that PR stays shippable and this buyer-visible change gets its own review.

### DIFF
--- a/src/mcp/core/paywall.ts
+++ b/src/mcp/core/paywall.ts
@@ -38,6 +38,32 @@ import { CreditsContextProvider } from './credits-context.js'
 let authHeaderDeprecationWarned = false
 
 /**
+ * The `creditsRedeemed` value for `_meta['nevermined/credits']`, or `undefined`
+ * to OMIT the key.
+ *
+ * ⚠️ Exists so the two `_meta` builders cannot drift. They are separate code —
+ * the non-streaming tool result and `wrapAsyncIterable`'s final chunk — and a
+ * test on one proves nothing about the other, which is why each has its own
+ * regression test below this file.
+ *
+ * The rule, which is deliberately NOT the A2A handler's:
+ * - a successful settle publishes whatever the facilitator reported, including
+ *   the `'0'` a pay-as-you-go plan returns — because this key also carries
+ *   `billingModel`, so a consumer can tell that `'0'` from a credits settle that
+ *   burned nothing. A2A's `creditsCharged` omits instead, having no
+ *   discriminator beside it. See the PR discussion on #443.
+ * - a successful settle that reported no figure omits, rather than substituting
+ *   the credits the request ASKED to burn (the defect this helper replaced).
+ * - a failed settle, and a free / no-credit call, report `'0'`.
+ */
+function resolveCreditsRedeemed(
+  settlement: { success?: boolean; creditsRedeemed?: string } | undefined,
+): string | undefined {
+  if (!settlement?.success) return '0'
+  return settlement.creditsRedeemed
+}
+
+/**
  * Main class for creating paywall-protected MCP handlers
  */
 export class PaywallDecorator {
@@ -252,11 +278,10 @@ export class PaywallDecorator {
             // than inventing one. Omission is the honest answer and it is
             // distinguishable: `remainingBalance`, `orderTx` and the full spec
             // receipt under X402_PAYMENT_RESPONSE_META_KEY are all still here.
-            ...(creditsResult?.success
-              ? creditsResult.creditsRedeemed !== undefined && {
-                  creditsRedeemed: creditsResult.creditsRedeemed,
-                }
-              : { creditsRedeemed: '0' }),
+            ...(() => {
+              const r = resolveCreditsRedeemed(creditsResult)
+              return r !== undefined ? { creditsRedeemed: r } : {}
+            })(),
             remainingBalance: creditsResult?.remainingBalance,
             ...(creditsResult?.orderTx && { orderTx: creditsResult.orderTx }),
             planId: authResult.planId,
@@ -418,11 +443,10 @@ function wrapAsyncIterable<T>(
           ...(settlement?.billingModel && { billingModel: settlement.billingModel }),
           // See the non-streaming site: no `?? credits.toString()` substitution.
           // A successful settle that reported no figure omits the key.
-          ...(settlement?.success
-            ? settlement.creditsRedeemed !== undefined && {
-                creditsRedeemed: settlement.creditsRedeemed,
-              }
-            : { creditsRedeemed: '0' }),
+          ...(() => {
+            const r = resolveCreditsRedeemed(settlement)
+            return r !== undefined ? { creditsRedeemed: r } : {}
+          })(),
           remainingBalance: settlement?.remainingBalance,
           ...(settlement?.orderTx && { orderTx: settlement.orderTx }),
           planId,
--- a/src/mcp/core/paywall.ts
+++ b/src/mcp/core/paywall.ts
@@ -46,6 +46,18 @@ let authHeaderDeprecationWarned = false
  * test on one proves nothing about the other, which is why each has its own
  * regression test below this file.
  *
+ * ⚠️ The two protocols deliberately DIVERGE here, and it is a settled decision
+ * (repo owner, on the #443 review) rather than an oversight — do not "fix" one
+ * to match the other:
+ *
+ * | surface | pay-as-you-go | why |
+ * | --- | --- | --- |
+ * | MCP `creditsRedeemed` (here) | reports `'0'` | it is the WIRE field name, and
+ *   `billingModel` is emitted beside it, so a consumer can read the `'0'` |
+ * | A2A `creditsCharged` (#439) | OMITS | it is a DERIVED summary with no
+ *   discriminator beside it, so a bare `0` would read as "you were charged
+ *   nothing" on a charge that succeeded |
+ *
  * The rule, which is deliberately NOT the A2A handler's:
  * - a successful settle publishes whatever the facilitator reported, including
  *   the `'0'` a pay-as-you-go plan returns — because this key also carries
--- a/src/mcp/core/paywall.ts
+++ b/src/mcp/core/paywall.ts
@@ -197,7 +197,6 @@ export class PaywallDecorator {
             onFinally,
             effectivePlanId,
             authResult.subscriberAddress,
-            credits,
           )
         }
 
@@ -242,9 +241,22 @@ export class PaywallDecorator {
             // buyer. Omitted (not '') when the settle carried no discriminator,
             // so a consumer can tell "absent" from "present and empty".
             ...(creditsResult?.billingModel && { billingModel: creditsResult.billingModel }),
-            creditsRedeemed: creditsResult?.success
-              ? (creditsResult.creditsRedeemed ?? credits.toString())
-              : '0',
+            // ⚠️ NO `?? credits.toString()` FALLBACK. That substituted the credits
+            // this request ASKED to burn and published them under the name
+            // `creditsRedeemed` — a figure the settle never reported, labelled as
+            // one it did. The A2A handler was corrected the same way in #438/#439;
+            // the two protocols must not tell a buyer different things about the
+            // same money.
+            //
+            // A successful settle that reports no figure now OMITS the key rather
+            // than inventing one. Omission is the honest answer and it is
+            // distinguishable: `remainingBalance`, `orderTx` and the full spec
+            // receipt under X402_PAYMENT_RESPONSE_META_KEY are all still here.
+            ...(creditsResult?.success
+              ? creditsResult.creditsRedeemed !== undefined && {
+                  creditsRedeemed: creditsResult.creditsRedeemed,
+                }
+              : { creditsRedeemed: '0' }),
             remainingBalance: creditsResult?.remainingBalance,
             ...(creditsResult?.orderTx && { orderTx: creditsResult.orderTx }),
             planId: authResult.planId,
@@ -376,7 +388,6 @@ function wrapAsyncIterable<T>(
   onFinally: () => Promise<any>,
   planId: string,
   subscriberAddress: Address,
-  credits: bigint,
 ) {
   async function* generator() {
     let creditsResult: any = null
@@ -405,9 +416,13 @@ function wrapAsyncIterable<T>(
           // See the non-streaming site: without `billingModel`, `creditsRedeemed`
           // cannot be read — it is '0' on a pay-as-you-go settle that charged.
           ...(settlement?.billingModel && { billingModel: settlement.billingModel }),
-          creditsRedeemed: settlement?.success
-            ? (settlement.creditsRedeemed ?? credits.toString())
-            : '0',
+          // See the non-streaming site: no `?? credits.toString()` substitution.
+          // A successful settle that reported no figure omits the key.
+          ...(settlement?.success
+            ? settlement.creditsRedeemed !== undefined && {
+                creditsRedeemed: settlement.creditsRedeemed,
+              }
+            : { creditsRedeemed: '0' }),
           remainingBalance: settlement?.remainingBalance,
           ...(settlement?.orderTx && { orderTx: settlement.orderTx }),
           planId,
--- a/tests/unit/mcp.test.ts
+++ b/tests/unit/mcp.test.ts
@@ -179,7 +179,15 @@ describe('MCP Integration', () => {
       // Note the figure the settle reports (7) deliberately differs from the
       // credits the tool asks to burn (3n), so this cannot pass by reading the
       // wrong one.
-      const mockInstance = new PaymentsMock({ success: true, creditsRedeemed: '7' })
+      // Complete shape: #437 (open) types this double as SettlePermissionsResult,
+      // where `transaction` and `network` are REQUIRED — so an incomplete fixture
+      // stops compiling once its `typecheck:tests` gate lands. Inert today.
+      const mockInstance = new PaymentsMock({
+        success: true,
+        transaction: '0xabc',
+        network: 'eip155:84532',
+        creditsRedeemed: '7',
+      })
       const pm = mockInstance as any as Payments
       const mcp = buildMcpIntegration(pm)
       mcp.configure({ agentId: 'did:nv:agent', serverName: 'test-mcp' })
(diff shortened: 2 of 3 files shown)


aaitor (commented):
Reviewed the diff against #439's fix (same substitution pattern, removed from MCP paywall instead of A2A). Confirmed the `?? credits.toString()` fallback is gone from both the streaming (`wrapAsyncIterable`) and non-streaming sites, and that the `credits`/`credits: bigint` parameters dropped from both signatures are not referenced anywhere else (checked the single call site of `wrapAsyncIterable`, which now passes 4 args matching the trimmed signature). Test changes include both the pinned-defect rewrite (no longer asserting the substituted value) and a positive control (settle reports 7, tool asks to burn 3, asserts 7) so a blanket 'never emit' change can't pass by accident. CI green (7/7). No blockers found.

eruizgar91:
👀 Reviewing this now.

src/mcp/core/paywall.ts
```diff
@@ -405,9 +416,13 @@ function wrapAsyncIterable<T>(
           // See the non-streaming site: without `billingModel`, `creditsRedeemed`
           // cannot be read — it is '0' on a pay-as-you-go settle that charged.
           ...(settlement?.billingModel && { billingModel: settlement.billingModel }),
-          creditsRedeemed: settlement?.success
-            ? (settlement.creditsRedeemed ?? credits.toString())
-            : '0',
+          // See the non-streaming site: no `?? credits.toString()` substitution.
+          // A successful settle that reported no figure omits the key.
+          ...(settlement?.success
+            ? settlement.creditsRedeemed !== undefined && {
```
eruizgar91: **MEDIUM** — The streaming half of this fix has **zero** regression coverage.

I restored `?? credits.toString()` at this site only (re-adding the `credits: bigint` parameter and the argument at its call site) and ran the full unit suite: **49 suites, 641 tests, 0 failures**. The mutant survives completely. The same mutant applied to the non-streaming site at `paywall.ts:256` is killed by exactly 1 test — so the guard exists on one `_meta` builder and not on the other, and the PR body's mutation table lists only the one that is covered.

The obligation is already written down in the file being edited, on an existing streaming test:

> `tests/unit/mcp.test.ts:681-682` — "Same guard as the non-streaming site — the two _meta builders are separate code, so a test on one proves nothing about the other."

Cheapest fix, no new fixture required: `should redeem after async iterable completes` already drives a bare `new PaymentsMock()` (settle `{ success: true }`, no `creditsRedeemed`) with `credits: 5n`, and stops at `expect(lastChunk._meta).toBeDefined()`. Add one line there:

```ts
expect(lastChunk._meta['nevermined/credits']).not.toHaveProperty('creditsRedeemed')
```

Before this PR that assertion would have read `'5'`, so it is a real mutant-killer.

https://github.com/nevermined-io/payments/blob/ef73c526e20e9f3cd95dedcde35de78bf74b190a/tests/unit/mcp.test.ts#L622-L626

Alternative that fixes the cause rather than the symptom: extract the shared ternary into one helper used by both builders — the shape #439 chose with `resolveCreditsCharged`, explicitly so "the two copies cannot drift". Right now they are verbatim duplicates and only one is tested.

src/mcp/core/paywall.ts
```diff
@@ -242,9 +241,22 @@ export class PaywallDecorator {
...
+            // one it did. The A2A handler was corrected the same way in #438/#439;
+            // the two protocols must not tell a buyer different things about the
+            // same money.
+            //
+            // A successful settle that reports no figure now OMITS the key rather
+            // than inventing one. Omission is the honest answer and it is
+            // distinguishable: `remainingBalance`, `orderTx` and the full spec
+            // receipt under X402_PAYMENT_RESPONSE_META_KEY are all still here.
+            ...(creditsResult?.success
+              ? creditsResult.creditsRedeemed !== undefined && {
```
eruizgar91: **MEDIUM** — This changes a documented response field's contract, and the docs are not updated.

Root [CLAUDE.md](https://github.com/nevermined-io/payments/blob/ef73c526e20e9f3cd95dedcde35de78bf74b190a/CLAUDE.md), "After Modifying Source Files", item 4:

> "If changing public interfaces (function signatures, options, types, **response fields**), update the corresponding documentation in `markdown/` to reflect the changes. Validate with `./scripts/generate-docs.sh`."

`markdown/mcp-integration.md` is the page that specifies this exact object, and it still presents `creditsRedeemed` as always present on a settled call — it covers only the free-call `'0'` and the pay-as-you-go `'0'` cases:

https://github.com/nevermined-io/payments/blob/ef73c526e20e9f3cd95dedcde35de78bf74b190a/markdown/mcp-integration.md#L471-L475

So an integrator reading it has no way to learn the one thing that changed: the key can now be absent on a **success**. Add a sentence to the "Response — settlement receipt" section mirroring the comment above — a successful settle that reports no figure omits the key rather than substituting the requested amount — and say that absence must be read with `in`/`hasOwnProperty`, not truthiness, since `'0'` is a legitimate value. #439 treated the docs as part of shipping these same semantics on A2A.

eruizgar91 (commented):
### Code review

Money-path change to the MCP `_meta['nevermined/credits']` receipt. I re-derived every claim from the PR head (`ef73c526`) in a fresh worktree rather than from the PR description, and re-checked that the head had not moved immediately before posting.

**4 findings, all MEDIUM, posted inline.** None is a defect in the fix itself: removing the substitution is right, and the rewritten test is genuinely non-vacuous.

**Verified by running it** (worktree at `ef73c526`, full `tests/unit` suite):

- **The rewritten test is not vacuous.** Restoring `?? credits.toString()` at the non-streaming site kills exactly 1 test — the rewritten `should add metadata to result after successful redemption`. Deleting the `creditsRedeemed` emission outright kills 7, so the added positive control does its job too.
- **The streaming site is unguarded.** The identical mutant at `wrapAsyncIterable` leaves 49 suites / 641 tests green. That is finding 1, and it is the only reason this review is not a clean pass.
- **No other test in the tree still pins the old substitution.** Every remaining `creditsRedeemed` assertion drives an explicit settle double that reports the figure it asserts. The `String(params.maxAmount)` doubles under `tests/integration/**` model a facilitator that redeems what was asked — that is the mock's own behaviour, not the paywall's fallback.
- **Nothing downstream re-introduces a number.** No `src/`, `cli/` or `openclaw/` code reads `_meta['nevermined/credits']` — the object is untyped and producer-only. The documented "was the buyer charged" recipes (`Number(settlement.creditsRedeemed ?? '0') > 0`) read the **spec** receipt under `x402/payment-response`, which this PR passes through verbatim, and their `?? '0'` already tolerates absence.
- **Behaviour is unchanged** for free calls (`'0'`), failed streaming settles (`'0'` + `errorReason`), and `''`. Absence stays distinguishable from a free call, because the spec receipt is present in one and not the other. Dropping `wrapAsyncIterable`'s `credits` parameter dangles nothing — history shows it was single-purpose from its introduction in `0a32b33`.

**Considered and deliberately discarded:**

- **`!== undefined` vs `!= null`.** `SettlePermissionsResult` is an unvalidated cast of backend JSON, where the only wire-representable "no figure" values are an absent key and `null` — and `null` would now be published verbatim under a field typed `string`. `src/x402/langsmith/spans.ts:290` answers the same question about the same field with `!= null`. I found no evidence the backend emits `null`, so this stays a note rather than a finding, but `!= null` would cost one character and close it.
- `paymentMiddleware`'s fallback to `creditsToCharge` when a settle omits `creditsRedeemed` (`src/x402/express/middleware.ts:1095-1106`). Same shape, different rail: the hook parameter is typed `number` and cannot express absence, and the fallback is deliberate, documented and warned. Not this PR's defect and out of scope.
- The `#438/#439` reference is written in the past tense while #439 is still open; the double blank line added in the test file is unchecked by anything (`lint` and `format` are scoped to `./src`). Both nits.

<sub>Reviewed at head `ef73c526e20e9f3cd95dedcde35de78bf74b190a`; head had not moved since the review was claimed.</sub>

src/mcp/core/paywall.ts
```diff
@@ -242,9 +241,22 @@ export class PaywallDecorator {
...
             // so a consumer can tell "absent" from "present and empty".
             ...(creditsResult?.billingModel && { billingModel: creditsResult.billingModel }),
-            creditsRedeemed: creditsResult?.success
-              ? (creditsResult.creditsRedeemed ?? credits.toString())
-              : '0',
+            // ⚠️ NO `?? credits.toString()` FALLBACK. That substituted the credits
+            // this request ASKED to burn and published them under the name
+            // `creditsRedeemed` — a figure the settle never reported, labelled as
+            // one it did. The A2A handler was corrected the same way in #438/#439;
+            // the two protocols must not tell a buyer different things about the
```
eruizgar91, line 286: **MEDIUM** — This comment claims a parity the change does not deliver, and the half left behind is the one that costs money.

#439's `resolveCreditsCharged` is an **allowlist**: `if (settlement.billingModel !== undefined && settlement.billingModel !== 'credits') return undefined`. So on a **pay-as-you-go** settle, A2A **omits** `creditsCharged` entirely — on the stated grounds that reporting `0` would tell a buyer they were charged nothing for a payment that really happened.

This PR's condition is `creditsRedeemed !== undefined`. On pay-as-you-go the settle reports the *defined* string `'0'`:

https://github.com/nevermined-io/payments/blob/ef73c526e20e9f3cd95dedcde35de78bf74b190a/src/x402/facilitator-api.ts#L264-L270

so the condition holds and MCP still publishes `creditsRedeemed: '0'`. The two protocols therefore still tell a buyer different things about the same pay-as-you-go charge — precisely what this comment says must not happen.

I am **not** asking you to change that behaviour here. MCP deliberately ships `billingModel` alongside as the discriminator and documents the `'0'` explicitly, so switching to omission would change a published contract, and it is the product call already raised on #439 ("That is a product call, not a cleanup, so I have asked rather than decided it"). The bounded in-PR fix is to stop the comment overclaiming: say the `?? credits.toString()` substitution is gone, and that the pay-as-you-go divergence is deliberately out of scope, naming the issue that owns it (#438). Left as written, the next reader concludes the two protocols already agree and stops looking.

tests/unit/mcp.test.ts
```diff
@@ -156,11 +156,52 @@ describe('MCP Integration', () => {
...
 
+    it('publishes creditsRedeemed when the settle actually reports one', async () => {
+      // The positive control for the test above. Without it, deleting the
+      // creditsRedeemed emission entirely would pass — "the key is absent" is
+      // satisfied just as well by never emitting it at all.
+      //
+      // Note the figure the settle reports (7) deliberately differs from the
+      // credits the tool asks to burn (3n), so this cannot pass by reading the
+      // wrong one.
+      const mockInstance = new PaymentsMock({ success: true, creditsRedeemed: '7' })
```
eruizgar91: **MEDIUM** — This new fixture will break the `Typecheck tests` CI gate that open PR #437 adds.

#437 (`fix/433-typed-settle-doubles`, OPEN) types the double — `constructor(settleResult?: SettlePermissionsResult)` — and adds `pnpm typecheck:tests` (`tsc --noEmit -p tests/tsconfig.json`) as a step in `.github/workflows/testing.yml`. `SettlePermissionsResult` declares both `transaction: string` and `network: string` as **required**:

https://github.com/nevermined-io/payments/blob/ef73c526e20e9f3cd95dedcde35de78bf74b190a/src/x402/facilitator-api.ts#L242-L255

`{ success: true, creditsRedeemed: '7' }` satisfies neither, so it stops compiling once both land. #437 went through every other fixture in this file to complete them, so this would be the only one left behind.

Make it whole now — inert today, correct after #437, and it does not weaken the test (the point is that `7` differs from the `3n` requested):

```ts
const mockInstance = new PaymentsMock({
  success: true,
  transaction: '0xdefaultsettletx',
  network: 'eip155:84532',
  creditsRedeemed: '7',
})
```

Separately, and worth flagging since both PRs are yours: #437 adds a comment on `PaymentsMock`'s default fixture that this PR invalidates — "`creditsRedeemed` is deliberately absent, so tests inheriting this default still exercise the paywall's **fallback to the requested credit amount**". It lives in a hunk this PR does not touch, so it merges cleanly and survives as standing guidance to re-add the substitution you are deleting here. Whichever of the two lands second should delete that sentence.

eruizgar91 (approved):
🟡 **Approved with comments** — no blockers; 4 should-fix, 0 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/443#pullrequestreview-5177397433

src/mcp/core/paywall.ts
r-marques: Folded in `d85c235`, and this is the finding I should have caught myself — my own mutation only probed the site I happened to be editing.

The gap is closed and measured. Same mutant, restoring the substitution at `wrapAsyncIterable` **only**:

| | before | after |
| --- | --- | --- |
| streaming mutant | 49 suites / 641 tests, **0 failures** | **kills** `streaming: a settle reporting no figure omits creditsRedeemed, never the requested burn` |

Two things rather than one:

1. **A shared `resolveCreditsRedeemed`**, so the two `_meta` builders no longer re-derive the rule independently. The duplication was the root cause, not the missing test.
2. **A streaming test plus a positive control** (a settle reporting `'9'` against a tool asking for `5n`), so deleting the streaming emission outright cannot satisfy "the key is absent".

The sharpest part of your comment is that the obligation was **already written in the file**, two tests above mine: *"the two _meta builders are separate code, so a test on one proves nothing about the other."* I added a test next to a comment telling me to add exactly the test I did not add. That is worse than not knowing the rule.

src/mcp/core/paywall.ts
r-marques: Folded in `d85c235`. You are right that this is a documented response field and root `CLAUDE.md` item 4 applies.

`markdown/mcp-integration.md` now carries the four cases explicitly — figure reported, **no figure reported (key absent)**, pay-as-you-go `'0'`, and failed / free-call `'0'` — plus the `in`-over-truthiness rule, since `'0'` is both a legitimate value and truthy.

I also added the point that makes the omission safe to document rather than alarming: the facilitator's response is passed through **verbatim** under `x402/payment-response`, so the raw figure is available whatever this summary does. That is the same reasoning that makes omission correct for a summary key and wrong for the receipt.

tests/unit/mcp.test.ts
r-marques: Folded in `d85c235` — and thank you for catching a cross-PR interaction I would not have seen from inside this branch.

`{ success: true, creditsRedeemed: '7' }` now carries `transaction` and `network`, which `SettlePermissionsResult` declares required. Inert today, correct once #437's `typecheck:tests` gate lands.

I completed the **two new streaming fixtures** in this same fold for the same reason, rather than only the one you named — otherwise I would have fixed the fixture you found and added two more of exactly the same shape in the same commit.

Worth noting #437 has since grown a related fix from your other review: `tests/utils.ts` carried a stale `@ts-nocheck` that exempted it from the very gate #437 adds. So that gate is both closer to landing and wider than it was when you wrote this.

src/mcp/core/paywall.ts
r-marques: You are right that the two protocols still differ on pay-as-you-go, and I am **not** folding this one — it needs a product call, so I am asking rather than deciding.

**Your analysis is correct.** #439's `resolveCreditsCharged` is an allowlist, so A2A omits `creditsCharged` on pay-as-you-go. This PR's condition is `creditsRedeemed !== undefined`, and a PAYG settle reports the *defined* string `'0'` — so MCP still publishes it. My commit message claimed a parity the change does not deliver, and that was overstated.

**What stopped me implementing it:** the MCP behaviour is deliberate and tested, not an oversight. `surfaces billingModel + orderTx on a pay-as-you-go receipt, where creditsRedeemed is "0"` asserts it on **both** transports, and its comment argues the design — *"Without `billingModel` a consumer of this key cannot tell this from a credits settle that burned nothing"*. Making MCP omit breaks three existing tests, including the free-call case.

So there are two defensible designs, chosen for the same #2999 reason:

| | shape | argument |
| --- | --- | --- |
| **A2A** `creditsCharged` | omit on PAYG | a derived summary with **no discriminator beside it** — a bare `0` is unreadable |
| **MCP** `creditsRedeemed` | publish `'0'` + `billingModel` | the wire field name, **with** the key to read it, and the raw receipt alongside |

Picking between those is a preference, not a deduction, so it is the repo owner's call. I have put it to him; this PR leaves the behaviour as it is and says so in the commit message rather than quietly shipping half a parity claim.

If the answer is "make them match", it is a small change here plus updating those three tests — and I will do it in this PR rather than a follow-up.

r-marques: ⚠️ I resolved this thread by accident a moment ago and have **reopened** it — apologies.

My resolve loop filtered by line number to skip this one, and the push had shifted the anchors to `null`, so the filter matched nothing and closed all four. The other three are genuinely addressed; this one is not, and it should stay open until the repo owner rules on it.

Reopened rather than left closed with an explanation, because a resolved thread is an assertion to the next reader that the point was handled — and this one is waiting on a decision, not on me.

r-marques: **Ruled: keep them different.** Put to the repo owner, who chose to leave MCP reporting `'0'` and A2A omitting, on the grounds you can see in your own finding — they are different fields solving different problems:

| surface | pay-as-you-go | why |
| --- | --- | --- |
| MCP `creditsRedeemed` | reports `'0'` | the **wire field name**, with `billingModel` emitted beside it, so the `'0'` is readable |
| A2A `creditsCharged` | **omits** | a **derived summary** with no discriminator next to it, so a bare `0` reads as "you were charged nothing" on a charge that succeeded |

So no behaviour change, but your finding was still right and has produced something: the divergence was **undocumented**, which is why it read as drift. It is now recorded **on both sides** — at `resolveCreditsRedeemed` here (`2f1064e`) and at `resolveCreditsCharged` in #439 (`a65cc53`) — each naming the other, each saying *do not reconcile the two*.

That is the lesson from a sibling review this week: two statements that merely agree today hand whoever reconciles them later a coin flip. Two that **disagree** are worse, unless each says why and points at the other.

Resolving now that the decision is recorded. And apologies again for the accidental resolve earlier — the anchors went `null` after a push and my filter closed all four.

Issue #446, opened 2026-09-11
cli: three unit test files are silently ignored, and plans.test.ts cannot compile


r-marques:
Found while folding the review on #441.

`cli/jest.config.*` carries:

```js
testPathIgnorePatterns: [
  '/node_modules/',
  '/test/unit/config.test.ts',
  '/test/unit/plans.test.ts',
  '/test/unit/agents.test.ts',
],
```

So those three files cannot run under **any** script — not `test:unit`, not `test:unit:all`, not `test:all`. `cli-sync-and-test.yml` runs `pnpm test:unit`, which is itself a hardcoded four-file list (`simple`, `output-formatter-basic`, `x402`, `generator-complex-type`) that never named them either. They are dead weight that reads like coverage.

## Why it matters

The review on #441 cited `cli/test/unit/plans.test.ts:91-111` as *"the only test covering `plans get-plan-balance`"* — the test that would catch a regression of #440. It provides **zero** coverage: it is ignored by config, and it does not even compile.

## What is actually wrong with each

**`plans.test.ts` — compile failure against pre-rename paths.** It imports modules that no longer exist:

```
TS2307: Cannot find module '../../src/commands/plans/list.js'
TS2307: Cannot find module '../../src/commands/plans/get.js'
TS2307: Cannot find module '../../src/commands/plans/balance.js'
```

The commands were not deleted — the CLI moved to generated, verb-prefixed files, so they live at `src/commands/plans/get-plans.ts`, `get-plan.ts` and `get-plan-balance.ts`. **Repoint the imports** (and the command classes they drive) rather than deleting the file.

**`config.test.ts` — a real assertion failure**, not a compile error. Forced to run, `should return null for non-existent key` fails. That is a genuine behaviour question to look at, not a path fix.

**`agents.test.ts`** — not investigated; check it the same way.

## The work

1. Repoint `plans.test.ts` at the current generated command modules.
2. Diagnose `config.test.ts`'s failing assertion — the test or the code is wrong; determine which.
3. Triage `agents.test.ts`.
4. Remove each from `testPathIgnorePatterns` as it goes green, and add it to `test:unit` so CI actually runs it.

Not folded into #441 because that PR changes one SDK type plus three fixture literals, and this needs its own review: it touches the CLI's test harness, and step 2 may be a code fix rather than a test fix.

## Not blocked on this

#441's own fix stands on its own — the fixtures were contradicting the wire whether or not anything reads them, and they are corrected there.

[...]

PR #448, opened 2026-09-13
test(cli): un-ignore all three CLI test files, and fix the live mock they exposed

r-marques:
Partial fix for #446 — **the issue stays open** for the third file.

## Two of the three only needed their imports repointed

`plans.test.ts` and `agents.test.ts` imported pre-rename command modules (`plans/list.js`, `plans/get.js`, `plans/balance.js`, `agents/get.js`). The commands were never deleted — the CLI moved to generated verb-prefixed files. Both pass after repointing.

## ⚠️ What being ignored was hiding — a defect on a merged PR

Running `plans.test.ts` for the first time in however long surfaced two genuine drifts in `test/__mocks__/@nevermined-io/payments.ts`, **the mock jest actually resolves**:

**1. `balance: bigint` / `BigInt(1000)`** — the exact defect #440 describes, still present after #441 was supposed to have fixed it.

#441 fixed `test/helpers/mock-payments.ts`, which is imported by **nothing**. Two copies of the same mock existed; the review cited the dead one, I fixed the dead one, and neither of us checked which was wired in. Nothing caught it because the only test reading the live mock is the one jest ignores — so the ignore list was concealing the evidence that the fix had missed.

The orphan is **deleted** here rather than also corrected. A duplicate mock with no importer is what caused a fix to land in the wrong file, and it would again.

**2. `getPlans` returned `{ data: plans }`.** `PlansAPI.getPlans` returns `{ total, page, offset, plans }`. The test asserted the mock's shape, so the two were wrong *together* and agreed with each other — which is why nothing looked suspicious.

Two stale assertions fixed alongside: the table formatter has no Yes/No mapping for booleans (it renders `true` via `String()`), and the JSON output is the paginated envelope, not a bare array.

## Wired into CI, or it rots again

Removing them from `testPathIgnorePatterns` is **not sufficient**: `cli-sync-and-test.yml` runs `pnpm test:unit`, which is a hardcoded four-file list that never named them. Both are added to that script too.

| | before | after |
| --- | --- | --- |
| `test:unit` | 4 suites / 33 tests | **6 suites / 45 tests** |
| `test:integration` | 28 pass | 28 pass |

## Why `config.test.ts` is still ignored

It does **not** fail on a stale assertion — that was my initial reading and it was wrong. It fails with a Jest **transform** error via `src/commands/config/init.ts`:

```
/node_modules/.pnpm/inquirer@9.3.8/node_modules/inquirer/lib/index.js:6
import { default as List } from './prompts/list.js';
```

`inquirer@9.3.8` ships ESM the transform does not handle. That is a jest-config problem of its own, not a test fix, and solving it risks the CLI's whole test setup — so it stays ignored with the diagnosis recorded in a comment beside the remaining entry, and #446 stays open for it.

## Verification

- CLI `pnpm test:unit` — 6 suites / **45 tests** pass · `pnpm test:integration` — 2 suites / **28** pass
- SDK unaffected: `pnpm typecheck:tests` exit 0, `pnpm test:unit` **643** pass
- Run against the **local SDK symlink** the CI workflow uses (`ln -s "$PWD" cli/node_modules/@nevermined-io/payments`). Without it a plain worktree resolves the *published* 1.3.2 and every API added since reads as a CLI defect — the trap recorded on #446.


r-marques:
## Now closes #446 completely — the third file is fixed too

When I opened this, `config.test.ts` stayed ignored: it failed with a Jest **transform** error (`inquirer@9` ESM), which I called a jest-config problem of its own.

**#451 landed that exact fix for the SDK's `tests/jest.config.json` a few hours later**, and the pattern transfers — including the `(\.pnpm/)?` hop, which is what pnpm's store layout requires and is the part easy to omit. Applied to `cli/jest.config.js` with inquirer's chain named, plus an `^.+\.m?js$` transform entry.

`config.test.ts`: transform error, 0 tests → **7 passing**.

So the PR title is updated: all three previously-ignored files now run.

| | before this PR | after |
| --- | --- | --- |
| `test:unit` (what CI runs) | 4 suites / 33 tests | **7 suites / 52 tests** |
| `test:integration` | 2 / 28 | unchanged, 28 pass |

## Two pre-existing failures I did NOT fold

Running the whole `test/unit` directory surfaces two more:

- `config-manager.integration.test.ts` — 1 assertion fails
- `commands-structure.test.ts` — suite fails to run

**Neither is caused by this change** — I re-ran both against the *previous* jest config and they fail identically. And neither has ever been in `testPathIgnorePatterns` **or** in `test:unit`'s hardcoded list, so CI has never run either of them.

They are outside #446, which names three specific files. Left for their own triage rather than folded in silently — but worth knowing the CLI has two more tests that exist and have never run.

cli/test/unit/agents.test.ts
```diff
@@ -3,7 +3,7 @@
  */
 
 import { expect, test, describe, beforeEach, afterEach, jest } from '@jest/globals'
-import AgentsGet from '../../src/commands/agents/get.js'
+import AgentsGet from '../../src/commands/agents/get-agent.js'
```
aaitor, line 6: **Should-fix:** The command-rename sweep is incomplete. `cli/test/unit/commands-structure.test.ts:10-14` still imports the removed `plans/list`, `plans/get`, `plans/balance`, `agents/get`, and `x402/get-token` modules. Running the existing `pnpm test:unit:all` script fails with five TS2307 errors before this suite can run. Update this sibling caller to the generated command paths (and its now-stale arg/description assertions) so the repository-wide test command works.

CLAUDE.md
```diff
@@ -253,7 +253,11 @@ Key files to check:
 - `openclaw/src/index.ts` — auto-pay flow, `getX402AccessToken()` calls
 - `openclaw/tests/plugin.test.ts` — mock assertions for `getX402AccessToken`
 - `cli/src/commands/x402token/get-x402-access-token.ts` — CLI command flags and call
-- `cli/test/helpers/mock-payments.ts` — mock `getX402AccessToken` signature
+- `cli/test/__mocks__/@nevermined-io/payments.ts` — the CLI's payments mock. This is the
+  one `jest.mock('@nevermined-io/payments')` resolves, so it is the only one that can
+  affect a test. A second copy at `cli/test/helpers/mock-payments.ts` was imported by
```
aaitor, line 258: **Should-fix:** Deleting the helper leaves several repository docs pointing at a file that no longer exists: `cli/CLI_PHASE2_TESTING.md:98,388`, `cli/CLI_PHASE2_PROGRESS.md:97`, `cli/CLI_GUIDE.md:396`, and `cli/CLI_STATUS.md:79`. Please update or remove those references in the same change; otherwise the new guidance here conflicts with the CLI testing documentation.

cli/test/__mocks__/@nevermined-io/payments.ts
```diff
@@ -94,7 +103,7 @@ export const Payments = {
           planName: plan.name,
           planType: plan.planType,
           holderAddress: '0x1234567890123456789012345678901234567890',
-          balance: BigInt(1000),
+          balance: '1000',
```
aaitor, line 106: **Should-fix:** The newly corrected wire type is not pinned by a test. `plans.test.ts:115` coerces `parsed.balance` through `BigInt(...)`, while `OutputFormatter` serializes an actual bigint to a string with its JSON replacer, so changing this mock back to `BigInt(1000)` still produces the same parsed value and leaves the assertions green. Add an assertion on the risky dimension (for example `typeof parsed.balance === 'string'`) so the drift this PR fixes cannot recur unnoticed.

aaitor (commented):
3 should-fix findings inline.

aaitor (approved):
🟡 **Approved with comments** — no blockers; 3 should-fix, 0 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/448#pullrequestreview-5210968421

cli/test/unit/agents.test.ts
r-marques: Folded in `b9aa981`. Repointed the five imports to the generated verb-prefixed files and corrected the assertions that had rotted with them — the real arg names are `plan` and `agent` (not `planId`/`agentId`), the balance flag is `account-address`, and `x402/get-token` is now `x402token/get-x402-access-token`. Descriptions on generated commands are the generator's prose, so the test now pins presence rather than exact text — an exact-string assertion there would break on the next regen for no information.

The five TS2307s are gone. `pnpm test:unit:all` has **one** remaining failure, `config-manager.integration.test.ts › should return null for non-existent key` (received `undefined`) — which is the exact assertion #446 flags as *"a genuine behaviour question to look at, not a path fix"*. Whether `get()` returns `null` or `undefined` for a missing key is a behaviour decision, not something to settle inside a test-un-ignore PR, and this PR does not touch that file or `cli/src/config`. It stays with #446, which stays open.

CLAUDE.md
r-marques: Folded in `b9aa981` — all four (`CLI_GUIDE.md`, `CLI_PHASE2_TESTING.md` ×3, `CLI_STATUS.md`, `CLI_PHASE2_PROGRESS.md`). They now name the mock jest actually resolves, `test/__mocks__/@nevermined-io/payments.ts`, and the `CLI_PHASE2_TESTING.md` entry carries the one-line story of why the other copy is gone, since that page is where the next author will look. Swept the repo for `helpers/mock-payments` afterwards: no reference left.

cli/test/__mocks__/@nevermined-io/payments.ts
r-marques: Folded in `b9aa981` — and the fix is not the one suggested, for the reason you gave yourself. `typeof parsed.balance === 'string'` is **also vacuous** against this regression: the replacer serialises a bigint to a string, so `parsed.balance` is a string either way. I added it first and measured — a mock regressed to `BigInt(1000) as any` still passed 9/9. Any assertion downstream of `OutputFormatter` cannot see this drift.

So the guards sit where the formatter cannot launder them:

1. **The mock's declared type, under ts-jest diagnostics.** `MockPlanBalance.balance: string` is already there, and a bare `BigInt(1000)` in the mock **fails to compile** — `TS2322: Type 'bigint' is not assignable to type 'string'`, 0 tests run. That guard existed before this fold; it just was not written down.
2. **A test that reads the mock's return directly** (`Payments.getInstance().plans.getPlanBalance(...)`) and asserts `typeof balance === 'string'` — which also catches a cast that gets past the type. Under the `as any` mutant that is the one test that goes red.

Both are in a comment above the new test, so the next person to "fix" the assertion sees why the obvious fix is vacuous. Declared commands clean: `test:unit` 7 suites / 53, `test:integration` 2 / 28.

PR #451, opened 2026-09-14
fix(tests): transform ESM dependencies, and take jose 6

eruizgar91:
## Problem

`jose` 6 dropped its CommonJS build. Jest never transformed it, so every suite that reaches `src/api/base-payments.ts` (which imports `decodeJwt`) failed on the dependency itself — this is what fails `Unit & Integration` on #316:

```
node_modules/.pnpm/jose@6.2.12/node_modules/jose/dist/webapi/index.js:1
SyntaxError: Cannot use import statement outside a module
```

Two separate things in `tests/jest.config.json` kept it out, and **both** had to change — each alone still fails:

1. **`transform` only mapped `^.+\.tsx?$`.** An explicit `transform` replaces Jest's defaults outright, so `.js` files had no transformer at all and were served raw, ESM and all. No dependency could ever be transformed, whatever the ignore list said.
2. **`transformIgnorePatterns` was `node_modules/(?!(@a2a-js/sdk)/)`**, which does not match pnpm's layout — the real path is `node_modules/.pnpm/<name>@<version>/node_modules/<name>/…`, so the first `node_modules/` is followed by `.pnpm/` and the exemption never applied. It has not been exempting `@a2a-js/sdk` either; that went unnoticed because that package still ships CJS.

## Change

- `tests/jest.config.json`: add a `.js`/`.mjs` transform, and an ignore pattern that matches both the flat and the pnpm layout, covering `@a2a-js/sdk` and `jose`.
- `package.json` / lockfile: `jose` `^5.2.4` → `6.2.12`.

`tests/jest.e2e.config.cjs` spreads this config, so it inherits both.

## Verification

With `jose@6.2.12` installed:

| | result |
|---|---|
| `pnpm build` | clean |
| `pnpm test:unit` | 49 suites, 643 passed / 644 (1 skipped) |
| `pnpm test:integration` | 11 suites, 55 passed |
| `pnpm typecheck:tests` | clean |
| ignore-pattern fix alone, without the `.js` transform | still `SyntaxError: Cannot use import statement outside a module` |
| `.js` transform alone, with the old ignore pattern | same failure |

Supersedes #316 — that PR bumps the same dependency and cannot pass on the current config.


r-marques:
:eyes: Reviewing

package.json
```diff
@@ -132,7 +132,7 @@
     "@traceloop/node-server-sdk": "^0.26.0",
     "axios": "^1.13.1",
     "express": "4.21.2",
-    "jose": "^5.2.4",
+    "jose": "^6.2.12",
```
r-marques, line 138: **🟡 SHOULD FIX — jose 6 dropped Node 18, but the SDK still documents Node ≥ 18 and the CLI declares it**

jose's [v6.0.0 release notes](https://github.com/panva/jose/releases/tag/v6.0.0) list *"drop support for Node.js 18.x and earlier"*, *"all builds and bundles now use ES2022 as target"* and *"CJS-style require is now only possible when require(esm) support is present"* (README footnote: `^20.19.0 || ^22.12.0 || >= 23.0.0`). `jose@6.2.12` ships no `engines`, so nothing warns at install time.

What the repo still says:

```
markdown/installation.md:24   - **Node.js**: Version 18.x or higher
cli/package.json:66           "node": ">=18.0.0"
package.json                  (no engines field)
```

CI runs `node-version: '24.x'` everywhere in `testing.yml`, so no job exercises 18, and Node 18 has been EOL since 2025-04-30.

This is a support-statement inconsistency, not a runtime break: the SDK's only jose call is `decodeJwt` (`src/api/base-payments.ts:1,164`), and its module-load path does not touch WebCrypto (verified by deleting `globalThis.crypto` and decoding a token — works). But it is a public statement the bump contradicts, and cheap to fix here: bump `markdown/installation.md:24` to Node 20 (20.19+ if you want to mention `require(esm)`), align `cli/package.json` `engines.node`, and consider adding `"engines": { "node": ">=20" }` to the root so `pnpm`/`npm` warn on an unsupported combination.

tests/jest.config.json
```diff
@@ -12,8 +21,12 @@
   "moduleNameMapper": {
     "^(\\.{1,2}/.*)\\.js$": "$1"
   },
-  "transformIgnorePatterns": ["node_modules/(?!(@a2a-js/sdk)/)"],
-  "extensionsToTreatAsEsm": [".ts"],
+  "transformIgnorePatterns": [
+    "node_modules/(?!(\\.pnpm/)?(@a2a-js[+/]sdk|jose)[@/])"
```
r-marques, line 25: **💡 NIT — The `@a2a-js/sdk` half of the exemption is inert in this Jest mode — and so is `useESM: true`; the real mechanism is ESM→CJS transpilation of jose**

The regex itself is right — measured on real store paths, old vs new (Jest realpaths symlinks, so the `.pnpm/<name>@<ver>/node_modules/<name>/…` form is what it sees):

```
old:IGNORED  new:TRANSFORM  .pnpm/jose@6.2.12/node_modules/jose/dist/webapi/index.js
old:IGNORED  new:TRANSFORM  .pnpm/@a2a-js+sdk@0.3.13_…/node_modules/@a2a-js/sdk/dist/client/index.js
old:IGNORED  new:IGNORED    .pnpm/@a2a-js+sdk@0.3.13_…/node_modules/@grpc/grpc-js/…   (2nd node_modules/ segment catches transitive deps)
old:IGNORED  new:IGNORED    .pnpm/jose-x@1.0.0/node_modules/jose-x/index.js           ([@/] terminator rejects lookalikes)
old:IGNORED  new:IGNORED    .pnpm/express@4.21.2/node_modules/express/index.js

store .js/.mjs files: 159,948 — exempted from ignore under NEW: 61, all from jose@6.2.12 (46) and @a2a-js+sdk@0.3.13 (15)
```

So the old pattern did ignore a2a (its first `node_modules/` is followed by `.pnpm/`), exactly as the PR body says. Two things the config *reads* as doing, though, that it does not:

1. **`@a2a-js/sdk` never reaches the transformer.** The `test*` scripts run Jest without `--experimental-vm-modules`, so `jest-resolve` uses the `require` export condition, and `@a2a-js/sdk` ships `"require": "./dist/*/index.cjs"` for every subpath the SDK imports (`require.resolve('@a2a-js/sdk/server')` → `dist/server/index.cjs`, on `main` too). `.cjs` matches neither `^.+\.tsx?$` nor `^.+\.m?js$`, so a2a was served raw CJS before this PR and still is — which is why the old no-op exemption never mattered. Only `jose` (exports `default` only → `dist/webapi/index.js`) is actually transformed.
2. **`useESM: true` is inert for the same reason.** Without vm-modules `supportsStaticESM` is false and ts-jest forces `module: CommonJS`. Ran Jest's own `ScriptTransformer` with this resolved config on `jose/dist/webapi/util/decode_jwt.js`: output is `"use strict"; … exports.decodeJwt = decodeJwt; const base64url_js_1 = require("./base64url.js"); …` — 4 `require()`, 0 `import`. The mechanism this PR relies on is *transpiling ESM-only jose to CJS*, not ESM support.

Both harmless; keep the a2a entry (it becomes live if the suite ever moves to real ESM). Suggest saying "transpiles ESM-only deps to CJS" in the commit message / PR body rather than "covering @a2a-js/sdk", so the next reader does not take the config at face value.

One more thing worth knowing about this line: the failure directions are asymmetric. jose *not* exempted fails closed (60/72 suites red in ~4s with the `Cannot use import statement` line). A lookahead that accidentally matches *nothing* fails **open** — every one of those ~160k files becomes transformable and the suite stays green, only ~30× slower (panel measured 268s at `maxWorkers: 1` vs ~8s). Wall time is the only tell; JSON cannot carry a comment saying so, so this note is it.

<sub>🤖 pr-review-toolkit · code-reviewer + pr-test-analyzer + silent-failure-hunter · high</sub>

r-marques (commented):

The diagnosis in the PR body holds on both counts and the fix is correct. The old `transformIgnorePatterns` never matched pnpm's `node_modules/.pnpm/<name>@<ver>/node_modules/<name>/` layout (its only `node_modules/` is followed by `.pnpm/`), so it exempted nothing — and the `@a2a-js/sdk` half never mattered because Jest, run without `--experimental-vm-modules`, resolves that package's `require` condition to its `.cjs` build. jose 6 is the first ESM-only dependency `src/` reaches (`src/api/base-payments.ts:1`), and it now works because ts-jest **transpiles it to CJS** — `useESM: true` is inert in this mode. Measured against the real store: of 159,948 `.js`/`.mjs` files, the new pattern exempts exactly 61, all under `jose@6.2.12` and `@a2a-js+sdk@0.3.13`, so nothing else is transformed and the suite's wall time is unchanged.

One should-fix: jose 6 formally dropped Node 18 while `markdown/installation.md` and `cli/package.json` still promise it — a support-statement drift, not a runtime break (`decodeJwt` loads fine without WebCrypto). Nits: prettier drift that triples the diff, the config reading as more than it does, and a cheap unit-level pin for the one behaviour the bump actually changes (jose 6's stricter decoder).

**What I reviewed**
- Diff range `origin/main...HEAD` — 3 files, +36/−17, head `d956807`. PR head contains `origin/main` (`3e58b24`); no rebase needed.
- Panel: code-reviewer, pr-test-analyzer, silent-failure-hunter (all read-only; reports retrieved by file).
- Applied the old and new regex to real installed paths with a node script, including two-`node_modules/`-segment transitive paths and lookalike names (`jose-x`, `@a2a-js/sdk-x`); counted exemptions across the whole `.pnpm` store.
- Ran Jest's own `ScriptTransformer` under the resolved config on `jose/dist/webapi/util/decode_jwt.js` — output is CommonJS (4 `require()`, 0 `import`).
- `pnpm install --frozen-lockfile` → 0; lockfile delta is the three `jose@…` entries only. `tsc --noEmit` → 0, `pnpm typecheck:tests` → 0, `pnpm build` → 0.
- `pnpm test:unit` (jest 29.7.0 + ts-jest 29.4.9) on a fresh worktree per side, cold cache: `origin/main` 49 suites / 643 passed / 1 skipped, 8.6s; head 49 / 643 / 1, 8.2s. Zero failures either way.
- CJS consumer of the SDK: `require('@nevermined-io/payments')` fails with `ERR_PACKAGE_PATH_NOT_EXPORTED` on the SDK's own `import`-only exports map — on `main` as well — so jose 6's CJS drop changes nothing for consumers.
- All 8 PR checks green on `d956807`. #316 (Dependabot, same bump) is still **open**; Dependabot closes it itself once `main` carries `jose ^6.2.12` — no `ignore` rule needed in `.github/dependabot.yml`.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 1 should-fix, 3 nits.

### 🧭 Inline comments (6)

- 🟡 **SHOULD FIX** — jose 6 dropped Node 18, but the SDK still documents Node ≥ 18 and the CLI declares it — `package.json:135`
- 💡 **NIT** — The `@a2a-js/sdk` half of the exemption is inert in this Jest mode — and so is `useESM: true`; the real mechanism is ESM→CJS transpilation of jose — `tests/jest.config.json:25`
- 💡 **NIT** — Array reformatting drifts from the repo's prettier style and inflates the diff — `tests/jest.config.json:4`
- 💡 **NIT** — jose 6's decoder is stricter than jose 5's, and nothing at unit level pins what `decodeJwt` returns or rejects — `package.json:135`
- ✅ **NICE** — The `.js` → ts-jest route produces loadable CJS for all of jose, and the `.js`-stripping `moduleNameMapper` does not break jose's internal imports — `tests/jest.config.json:41`
- ✅ **NICE** — Lockfile change is exactly the jose dedup; frozen install is clean; typecheck, test typecheck and build all pass — `pnpm-lock.yaml:3743`

tests/jest.config.json
```diff
@@ -1,9 +1,18 @@
 {
   "preset": "ts-jest",
   "testEnvironment": "node",
-  "roots": ["<rootDir>"],
-  "testMatch": ["**/unit/**/*.test.ts", "**/integration/**/*.test.ts", "**/e2e/**/*.test.ts"],
-  "collectCoverageFrom": ["src/**/*.ts", "!src/**/*.d.ts"],
+  "roots": [
```
r-marques, line 4: **💡 NIT — Array reformatting drifts from the repo's prettier style and inflates the diff**

`npx prettier --config prettier.config.js --check tests/jest.config.json` passes on `origin/main`'s version and fails on this one — 23 lines differ, and the `roots` / `testMatch` / `collectCoverageFrom` / `extensionsToTreatAsEsm` hunks carry no change. `npx prettier --config prettier.config.js --write tests/jest.config.json` collapses the file to exactly the two meaningful hunks (`transformIgnorePatterns` and the new `^.+\.m?js$` entry).

No gate catches it (`pnpm format` is `prettier --check ./src`), so CI is green — but the next `--write` will revert these hunks, and the diff is three times harder to read than it needs to be.

package.json
r-marques, line 138: **💡 NIT — jose 6's decoder is stricter than jose 5's, and nothing at unit level pins what `decodeJwt` returns or rejects**

jose 5's Node build decoded the payload with `Buffer.from(s, 'base64url')` (lenient — skips invalid characters, accepts the standard alphabet) and a non-fatal `TextDecoder`; jose 6's single webapi build throws on non-alphabet characters and uses `TextDecoder('utf-8', { fatal: true })`. The panel measured the two side by side:

```
same  canonical base64url (what the platform mints)   jose5: OK   jose6: OK
DIFF  standard base64 alphabet (+ /)                   jose5: OK   jose6: JWTInvalid: Failed to base64url decode the payload
DIFF  stray character in payload                       jose5: OK   jose6: JWTInvalid
DIFF  invalid UTF-8 in payload                         jose5: OK   jose6: JWTInvalid: Failed to parse the decoded payload as JSON
```

Production impact is nil — core-kit's `EthSignJWT` emits canonical base64url (`+`→`-`, `/`→`_`, padding stripped), and the direction is safer. But those are exactly the two behaviours the bump touches, and neither is pinned: `decodeJwt` runs **unmocked** in the `BasePaymentsAPI` constructor (`src/api/base-payments.ts:164`) across 23 unit suites, yet no unit test asserts `getAccountAddress()`, and the only constructor-throw test passes `nvmApiKey: ''`, which returns at line 161 before `decodeJwt` is reached (the e2e suite does read the address — it passed on this head — but it needs staging credentials).

Two cheap assertions in `tests/unit/payments.test.ts` would close that, no new fixtures needed (read `sub` off `TEST_API_KEY` rather than hard-coding it, since line 11 falls back to `process.env.TEST_PROXY_BEARER_TOKEN`):

```ts
test('decodes the account address from the API key payload', () => {
  const sub = JSON.parse(Buffer.from(TEST_API_KEY.split(':')[1].split('.')[1], 'base64url').toString()).sub
  expect(Payments.getInstance({ nvmApiKey: TEST_API_KEY }).getAccountAddress()).toBe(sub)
})

test('rejects a malformed API key', () => {
  expect(() => Payments.getInstance({ nvmApiKey: 'sandbox:not.a-jwt' })).toThrow('Invalid NVM API Key')
})
```

Related but pre-existing, so optional: the `catch` at `base-payments.ts:168-170` discards the jose error, so a key mangled in transit now surfaces as a bare `Invalid NVM API Key` with no hint that it is an encoding problem. Appending `error.message` is safe — jose names the failing step, never the payload.

<sub>🤖 pr-review-toolkit · pr-test-analyzer + silent-failure-hunter · high</sub>

tests/jest.config.json
```diff
@@ -24,6 +37,17 @@
           "target": "ES2020"
         }
       }
+    ],
+    "^.+\\.m?js$": [
```
r-marques, line 41: **✅ NICE — The `.js` → ts-jest route produces loadable CJS for all of jose, and the `.js`-stripping `moduleNameMapper` does not break jose's internal imports**

Verified rather than assumed: with this resolved config, Jest's `ScriptTransformer` turns every one of jose's 46 `dist/webapi/*.js` files into CommonJS, and the `^(\.{1,2}/.*)\.js$ → $1` mapper rewrites jose's internal `./x.js` requires to `./x`, which Jest re-resolves to the `.js` sibling first (`moduleFileExtensions` starts with `js`; no extensionless/`.ts`/`.json` collisions in that tree). `tests/jest.e2e.config.cjs` spreads this file and adds only `globalSetup`, so e2e inherits both halves — and the E2E job on this head ran. Not adding `.js` to `extensionsToTreatAsEsm` is correct: jest-config rejects it ("always inferred from the nearest package.json"). Using ts-jest rather than the already-installed `babel-jest`/`@babel/preset-env` is also the right call — there is no babel config in the repo and `babel-jest` is at 30.x against jest 29.x.

Suite results, same runner (`pnpm test:unit`, jest 29.7.0 + ts-jest 29.4.9), cold cache after `jest --clearCache`:

| | suites | tests | jest time | wall |
|---|---|---|---|---|
| `origin/main` (jose 5.10.0, old config) | 49 passed | 643 passed / 1 skipped / 644 | 8.6s | 9.4s |
| PR head `d956807` | 49 passed | 643 passed / 1 skipped / 644 | 8.2s | 8.9s |

Zero failures either way; the ~46 extra jose transforms are a one-time cache cost inside the noise.

<sub>🤖 pr-review-toolkit · code-reviewer + own verification · high</sub>

pnpm-lock.yaml
```diff
@@ -3740,11 +3740,8 @@ packages:
   jju@1.4.0:
     resolution: {integrity: sha512-8wb9Yw966OSxApiCt0K3yNJL8pnNeIv+OEq2YMidz4FKP6nonSRoOXc80iXY4JaN2FC11B9qsNmDsm+ZOfMROA==}
 
-  jose@5.10.0:
-    resolution: {integrity: sha512-s+3Al/p9g32Iq+oqXxkW//7jk2Vig6FF1CFqzVXoTUXt2qz89YWbL+OwS17NFYEvxC35n0FKeGO2LGYSxeM2Gg==}
-
-  jose@6.2.2:
-    resolution: {integrity: sha512-d7kPDd34KO/YnzaDOlikGpOurfF0ByC2sEV4cANCtdqLlTfBlw2p14O/5d/zv40gJPbIQxfES3nSx1/oYNyuZQ==}
+  jose@6.2.12:
```
r-marques, line 3743: **✅ NICE — Lockfile change is exactly the jose dedup; frozen install is clean; typecheck, test typecheck and build all pass**

`git diff origin/main...HEAD -- pnpm-lock.yaml | grep -E '^[-+]  [a-z@]'` touches only `jose@5.10.0` (−), `jose@6.2.2` (−) and `jose@6.2.12` (+) — one jose now serves both the root and `@modelcontextprotocol/sdk@1.29.0` (range `^6.1.3`). `pnpm install --frozen-lockfile` exits 0. On the head: `tsc --noEmit -p tsconfig.json` 0, `pnpm typecheck:tests` 0, `pnpm build` 0. The whole tree has exactly one jose import (`src/api/base-payments.ts:1`, `decodeJwt`), whose jose 6 signature is unchanged; nothing references the removed `KeyLike` / `KeyObject`-returning importers; `cli/` and `openclaw/` pin no jose of their own.

On the CJS-consumer question the bump raises: **nothing changes.** `require('@nevermined-io/payments')` already fails on the SDK's *own* exports map (`"import"` + `"types"` conditions only, no `"require"`/`"default"`) — reproduced against the built `dist/` on Node 22.22 with `ERR_PACKAGE_PATH_NOT_EXPORTED: No "exports" main defined`, identically on `main`. `import('@nevermined-io/payments')` from CJS works (85 exports). And `require('jose')` 6.2.12 itself actually succeeds on Node ≥ 22.12 / 20.19 via `require(esm)`. So jose 6 is not what stands between a CJS consumer and this SDK; the SDK's ESM-only exports map is, and always was.

<sub>🤖 pr-review-toolkit · own verification · high</sub>

r-marques (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 3 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/451#pullrequestreview-5198922867

eruizgar91:
Fixed in `7ca679d`.

jose 6's release notes do say *"drop support for Node.js 18.x and earlier"*, and `jose@6.2.12` ships no `engines`, so nothing warns at install. The three places now say 20.19:

- `markdown/installation.md` — "Version 20.19 or higher", naming jose as the reason.
- `cli/package.json` — `engines.node` `>=18.0.0` → `>=20.19.0`.
- root `package.json` — gains `"engines": { "node": ">=20.19.0" }`, which it did not have, so npm and pnpm warn on an unsupported combination instead of it surfacing later inside a dependency.

Left as you framed it — a support statement, not a runtime break: the SDK's only jose call is `decodeJwt` and its load path does not touch WebCrypto. `./scripts/generate-docs.sh` clean, build clean, 643 unit tests green.

One deliberate omission: I did not add the `20.19+` / `require(esm)` footnote to the docs. The SDK is ESM-only, so the CJS-require caveat cannot apply to a consumer of this package, and it would read as a caveat about something they cannot do.

cli/package.json
```diff
@@ -63,7 +63,7 @@
     "typescript": "^5.4.3"
   },
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=20.19.0"
```
r-marques: **🟡 SHOULD FIX — Three published pages still promise Node 18 — the CLI's own Getting Started now contradicts this line, and the two OpenClaw pages ship inside its npm tarball**

Round 1 named three places and the commit updated exactly those three — but the same grep, run against the whole tree, finds three more published statements of the old floor that now disagree with this line (my miss in round 1 as much as anyone's; they need to go with this PR, not after it):

```
cli/docs/getting-started.md:15        - **Node.js** >= 18.0.0
openclaw/docs/getting-started.md:14   - Node.js >= 18.0.0
openclaw/docs/guide.md:43             - Node.js >= 18
```

All three are published, two ways:
- `release.yml:527` copies `cli/docs/*` and `:533` copies `openclaw/docs/*` to the docs site on every tag push — they are live today at `api-reference/cli/getting-started` and `api-reference/openclaw-plugin/{getting-started,guide}` (verified on the docs repo's `origin/main`). So after this release the CLI's own Getting Started page will say `>= 18.0.0` while the CLI's `engines` on this very line warns on 18.
- `openclaw/package.json` `files[]` includes `"docs/"`, so the two OpenClaw pages also ship **inside the `@nevermined-io/openclaw-plugin` tarball**, and the plugin declares `peerDependencies["@nevermined-io/payments"] = ">= 1.1.0"` — a Node-18 install of the plugin gets the SDK's `EBADENGINE` while the plugin's bundled docs promise 18.

Fix: set all three to whatever number the root/CLI `engines` end up saying (see the comment on `markdown/installation.md:24`). One internal doc carries the same sample — `cli/CLI_PHASE2_TESTING.md:319` (`node-version: '18'` in a copy-paste CI snippet) — not published, worth bumping while here.

Out of this repo's reach, for a one-line docs-repo PR alongside: `integrations/mcp.mdx:316` (`Node.js >= 18`) and `skills/nevermined-payments/SKILL.md:365` (`Node.js 18+`) are hand-maintained on the docs site and will disagree with the generated installation page once this ships. `openclaw/package.json` itself has no `engines` and never did; it inherits the check through the SDK peer, so adding one there is symmetry, not protection — optional.

<sub>🤖 pr-review-toolkit · code-reviewer + comment-analyzer (verified) · high</sub>

markdown/installation.md
```diff
@@ -21,7 +21,7 @@ The Payments Library provides:
 
 Before installing the Nevermined Payments Library, ensure you have:
 
-- **Node.js**: Version 18.x or higher
+- **Node.js**: Version 20.19 or higher (jose 6, a runtime dependency, dropped Node 18)
```
r-marques: **🟡 SHOULD FIX — 20.19 is the CJS `require(esm)` number the reply says does not apply; the reason written here ("jose 6 dropped Node 18") produces `>=20`, and nothing shipped produces `.19`**

The number and its stated reason disagree, and the number's only source is the footnote the reply correctly rules out.

**Where 20.19 comes from.** `jose@6.2.12` ships **no `engines`** and its README names no Node floor; the v6.0.0 note says only *"drop support for Node.js 18.x and earlier"* (→ `>=20`, no patch). `20.19` appears exactly once in jose's docs — footnote `[^cjs]`: CJS `require('jose')` works where `require(esm)` is default-on (`^20.19.0 || ^22.12.0 || >= 23.0.0`). The round-2 commit body reads that as *"documents ^20.19.0 || …"*, i.e. as a runtime floor; it is not one. And it cannot apply here for a stronger reason than "the SDK is ESM-only": the root `exports` map carries only `types` + `import` on every subpath, so `require('@nevermined-io/payments')` throws `ERR_PACKAGE_PATH_NOT_EXPORTED` on **every** Node version (reproduced on 22.22.2, where `require(esm)` is on) — `require(esm)` never gets a chance to matter. (Even under that reading, `>=20.19.0` is not that range: it admits 21.x and 22.0–22.11.)

**What the tree actually declares** — every direct runtime dep's `engines.node` on this head:

```
jose 6.2.12                                    (none)
@a2a-js/sdk 0.3.13                             >=18
@opentelemetry/exporter-trace-otlp-http 0.215  ^18.19.0 || >=20.6.0   ← strictest; top-level import of observability-api.ts, re-exported from index.ts
@traceloop/node-server-sdk 0.26.0              >=18.0.0
@opentelemetry/api 1.9.1                       >=8.0.0
express 4.21.2                                 >= 0.10.0
axios / uuid / zod / zod-to-json-schema / js-file-download / @helicone/helpers   (none)
peers: @langchain/core >=20 · @langchain/langgraph >=18 · @modelcontextprotocol/sdk >=18 · langsmith (none)
```

Nothing declares above 20.19, nothing declares 20.19, and `src/` uses nothing newer than `AbortSignal.timeout` (17.3+). **Measured** rather than inferred: on Node **20.18.3** (the last minor this floor excludes) the built `dist/index.js` imports (85 exports) and `Payments.getInstance(...).getAccountAddress()` decodes a key through jose 6 — so the SDK works on a version `engines` now rejects. The consumer-visible effect: `pnpm install` on 20.18.3 with this manifest prints `WARN Unsupported engine: wanted {"node":">=20.19.0"}` and proceeds (exit 0, pnpm 10.33, no `engine-strict`); npm behaves the same; Yarn classic hard-fails by default. That is a spurious warning across 20.6–20.18 — 23 releases, all EOL, so a small population, but the doc is what a reader on one of them checks, and it tells them two things.

**Fix — make the reason and the number agree** (three places move together: this line, `package.json:9`, `cli/package.json:66`). Recommended: `>=20.6.0` — the strictest floor anything on the import path declares — with the bullet reading `**Node.js**: Version 20.6 or higher` and **no dependency name**: "jose 6" is an internal a reader cannot act on, and it rots the day `decodeJwt` is replaced by ten lines of `Buffer.from(…, 'base64url')` (the derivation belongs in the commit body, where it already is). `>=20` is also defensible if you would rather match jose's own statement and let the exporter's `engines` self-report 20.0–20.5. If the team would rather keep `>=20.19.0` as a *support policy* ("final 20.x line and later" — reasonable given Node 20 is EOL), that is a product call I am not making here; but then the fix is to the reason, not the number — the bullet must say it is policy, because "jose 6 dropped Node 18" cannot justify `.19` either way. Either way, drop the `documents ^20.19.0 || …` sentence from the commit body or relabel it as the CJS footnote it is.

<sub>🤖 pr-review-toolkit · code-reviewer + comment-analyzer (verified + measured) · high</sub>

r-marques (commented):
## 🤖 Automated PR review — round 2 — 🟡 Mergeable with nits

Round 1's should-fix is half-closed. The three places it named now say `20.19` consistently (`markdown/installation.md:24`, `cli/package.json:66`, a new root `engines` at `package.json:8-10`), and the shape is right — the root block is the only install-time Node-18 signal in the whole graph (jose declares no `engines`; the OTLP exporter admits `^18.19.0`). Two things are off, both about *which* number and *where else it lives*: (1) round 1's grep — mine — missed three more published pages that still say Node 18, and the CLI's own Getting Started (`cli/docs/getting-started.md:15`) now contradicts the CLI `engines` set in the same commit, while the two OpenClaw pages ship inside its npm tarball (`files[]` includes `docs/`); (2) `20.19` is the CJS `require(esm)` footnote's number — the one the reply correctly says cannot apply — while the reason written on the page ("jose 6 dropped Node 18") produces `>=20`, and the strictest floor anything on the import path declares is `@opentelemetry/exporter-trace-otlp-http`'s `>=20.6.0`. Measured, not inferred: the built SDK imports and decodes a key on Node **20.18.3**, a version the new `engines` rejects; the effect on 20.6–20.18 is a spurious `WARN Unsupported engine` (pnpm 10.33 proceeds, exit 0 — it does not hard-fail). Safe in direction, wrong in derivation, and in three published places.

**Floor answer, for the record.** Nothing shipped declares above 20.19, nothing declares 20.19; the honest dependency-derived floor is `>=20.6.0` (or `>=20` to match jose's own statement and let the exporter self-report). Keeping `>=20.19.0` is fine *as a stated support policy* (Node 20 is EOL, last release 20.20.2) — that is the owner's call, not this review's — but then the reason on the page has to say so. Per-dep `engines` table and the `exports`-map evidence are in the inline comment on `installation.md:24`.

**Round-1 disposition**
- should-fix (Node floor vs jose 6): **addressed in the three named places**; three more published copies remain — inline comment on `cli/package.json:66`.
- nit (`useESM` / a2a half inert; mechanism is ESM→CJS transpile): not addressed — config and commit body unchanged; still a wording nit only.
- nit (prettier drift in `tests/jest.config.json`): not addressed — file byte-identical to round 1's head.
- nit (jose 6 stricter decoder; `decodeJwt` unmocked in 23 suites with no malformed-key assertion): not addressed — no test file changed.

**What I reviewed**
- Reviewed `7ca679d3` = round-1 head `d956807` + one commit (`docs: state the Node floor jose 6 actually requires`); `git diff --stat d9568075..7ca679d3` = 3 files, +5/−2. `pnpm-lock.yaml` and `tests/jest.config.json` are byte-identical to round 1, so the unit suite was **deliberately not re-run**. All 9 checks green on `7ca679d3`; auto-merge off.
- **Head moved mid-review to `5227016f`** — `Merge branch 'main'` bringing in #450 (`src/api/observability-api/types.ts`, already on `origin/main` as `3646a1b`). The five PR files are byte-identical across the merge, so every anchor below holds; I moved the review checkout to `5227016f` and re-ran `pnpm install --frozen-lockfile` / `pnpm build` / `pnpm typecheck:tests` there (all 0). Its E2E check is red on a staging `502 Bad Gateway` from `ContractsAPI.getDeploymentInfo` (PAYG suite; job is `continue-on-error: true`; identical PR content passed E2E on `7ca679d3`) — a backend blip, not this PR.
- Panel: code-reviewer + comment-analyzer (read-only; reports retrieved by file; every outside-the-diff claim re-verified by hand before it appears here).
- Read `engines.node` off every direct runtime + peer dep's installed `package.json`; read `jose@6.2.12`'s `package.json`, README (`[^cjs]` footnote) and the v6.0.0 release notes; grepped `src/`, `cli/src/`, `openclaw/src/` for `require('jose'|'@nevermined-io/payments')` (none); confirmed all three in-tree packages are `"type": "module"`.
- Installed Node 20.18.3 and ran the built `dist/index.js` there (import OK, `getAccountAddress()` decodes); ran `pnpm install --frozen-lockfile` on 20.18.3 against this head's manifest (WARN + exit 0) and against round 1's (no WARN) as a control.
- Confirmed the three stale pages are published: `release.yml:527/533` copy `cli/docs/*` and `openclaw/docs/*` to the docs site (live on the docs repo's `origin/main`); `openclaw/package.json` `files[]` includes `docs/`. Two hand-maintained docs-site pages (`integrations/mcp.mdx:316`, `skills/nevermined-payments/SKILL.md:365`) also say 18 — docs-repo follow-up, out of this PR's reach.
- CI matrix: every workflow `node-version: '24.x'`; no `.npmrc`, no `engine-strict` anywhere; `pnpm install --frozen-lockfile` 0, `pnpm build` 0, `pnpm typecheck:tests` 0. `prettier --check` clean on `cli/package.json` and `package.json`; `markdown/installation.md` fails only on pre-existing front-matter quoting (identical on `origin/main`, line 24 untouched). The repo's `pnpm format` gate is red on 4 `src/` files this PR does not touch — pre-existing, not counted.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 2 should-fix, 1 nit.

### 🧭 Inline comments (4)

- 🟡 **SHOULD FIX** — Three published pages still promise Node 18 — the CLI's own Getting Started now contradicts this line, and the two OpenClaw pages ship inside its npm tarball — `cli/package.json:66`
- 🟡 **SHOULD FIX** — 20.19 is the CJS `require(esm)` number the reply says does not apply; the reason written here ("jose 6 dropped Node 18") produces `>=20`, and nothing shipped produces `.19` — `markdown/installation.md:24`
- 💡 **NIT** — The PR title is the changelog line, and it says nothing about the new Node floor — `package.json:8`
- ✅ **NICE** — Adding `engines` to the root at all is the only install-time Node-18 signal in the whole dependency graph — and it warns rather than blocks, as the commit says — `package.json:9`

package.json
```diff
@@ -5,6 +5,9 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "type": "module",
+  "engines": {
```
r-marques, line 8: **💡 NIT — The PR title is the changelog line, and it says nothing about the new Node floor**

`CHANGELOG.md` is generated by `auto-changelog` from PR titles (`prepare-release.yml:61`; the v1.13.0 entries are the merged PR titles verbatim). The one change consumers actually feel from this PR is this block — `engines` appears on the SDK and rises on the CLI, i.e. **Node 18 is dropped** — and the release notes will say "fix(tests): transform ESM dependencies, and take jose 6". Retitle so the changelog names it, e.g. `…take jose 6; declare Node >=20.x (drops Node 18)`. Whether dropping a Node major counts as semver-major for this SDK is a release-policy question for the owner — flagging, not deciding.

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

package.json
```diff
@@ -5,6 +5,9 @@
...
+    "node": ">=20.19.0"
```
r-marques: **✅ NICE — Adding `engines` to the root at all is the only install-time Node-18 signal in the whole dependency graph — and it warns rather than blocks, as the commit says**

jose declares no `engines` and the OTLP exporter admits `^18.19.0`, so before this block nothing in the tree would have flagged a Node-18 install of the SDK; now the package consumers actually install carries the statement. Measured on Node 20.18.3 with pnpm 10.33 and no `engine-strict`: `WARN Unsupported engine …` then a normal install (exit 0) — the commit body's "warn on an unsupported combination" is exactly what happens (I checked because pnpm is often said to hard-fail on a root-project `engines` mismatch; it does not here). Prose and both manifests moved in one commit rather than manifest-then-doc-later, which is the discipline the other comment asks to extend to the three sibling pages.

Gates on this head, all measured: `pnpm install --frozen-lockfile` 0 (Node 22.22 and 20.18), `pnpm build` 0, `pnpm typecheck:tests` 0, `prettier --check` on `cli/package.json` and `package.json` clean (`markdown/installation.md` fails only on pre-existing front-matter quote style — identical on `origin/main`, line 24 untouched by prettier). `pnpm-lock.yaml` and `tests/jest.config.json` are byte-identical to round 1's head `d956807`, so the unit suite was deliberately not re-run. All 9 checks green on `7ca679d3`. The head then moved to `5227016f` — a merge of `main` (#450, `src/api/observability-api/types.ts` only; the five PR files are byte-identical) — where I re-ran install/build/typecheck (all 0). Its one red, E2E, is a staging `502 Bad Gateway` from `ContractsAPI.getDeploymentInfo` in the PAYG suite (`continue-on-error: true`; identical PR content passed E2E 40 minutes earlier) — not this PR.

<sub>🤖 pr-review-toolkit · own verification · high</sub>

r-marques (approved):

Inline review: https://github.com/nevermined-io/payments/pull/451#pullrequestreview-5199393344

eruizgar91:
Both fixed in `d6c22f8`.

**The number was wrong, and it was my own reply that ruled it out.** `jose@6.2.12` ships no `engines` and names no Node floor; v6.0.0 says only *"drop support for Node.js 18.x and earlier"* → `>=20`. 20.19 is from the `[^cjs]` footnote — the versions where `require(esm)` is default-on — which, as I argued one comment earlier, cannot apply to a consumer of an ESM-only package. Quoting it as a runtime floor asserted something jose does not say. Now `>=20.0.0` in both `engines` blocks and "Version 20 or higher" in the docs.

**Three more published pages.** Caught and updated with this PR rather than after it:

```
cli/docs/getting-started.md:15        - **Node.js** >= 18.0.0  → >= 20.0.0
openclaw/docs/getting-started.md:14   - Node.js >= 18.0.0      → >= 20.0.0
openclaw/docs/guide.md:43             - Node.js >= 18          → >= 20
```

`release.yml` copies both directories to the docs site on every tag push and they ship in the npm tarballs, so those were three published statements contradicting the `engines` field this PR adds.

`./scripts/generate-docs.sh` clean, build clean, 643 unit tests green. A repo-wide grep for the old floor now returns nothing outside this PR's own prose.

package.json
r-marques, line 8: **💡 NIT — PR title and body still say nothing about the Node floor — unchanged since round 2**

Carried forward from round 2, untouched: title and body are identical. `CHANGELOG.md` is generated from PR titles (`.github/workflows/prepare-release.yml`), so the one change consumers actually feel from this PR — `engines` appears on the SDK for the first time (this block), and `>=18.0.0` → `>=20.0.0` on `cli/package.json:66`, i.e. **Node 18 is dropped** — will not appear in the release notes. The body's *Change* section also still lists only `tests/jest.config.json` and the jose bump; the `engines` block and the four doc edits are absent.

Suggest `fix(tests): transform ESM dependencies, take jose 6, declare Node >=20 (drops Node 18)` plus one bullet in the body. Whether dropping a Node major is semver-major for this SDK remains the owner's call — flagging, not deciding.

<sub>🤖 pr-review-toolkit · code-reviewer (verified) · high</sub>

markdown/installation.md
```diff
@@ -21,7 +21,7 @@ The Payments Library provides:
 
 Before installing the Nevermined Payments Library, ensure you have:
 
-- **Node.js**: Version 18.x or higher
+- **Node.js**: Version 20 or higher (jose 6, a runtime dependency, dropped Node 18)
```
r-marques, line 24: **💡 NIT — The number is right; the parenthetical names jose as the reason, and jose is neither the only nor the strictest floor in the tree**

Round 2's should-fix is closed: `20` matches "jose 6 dropped Node 18", and the derivation in the commit body is correct (`jose@6.2.12` ships no `engines`; `20.19` was the `[^cjs]` footnote, README line 148, unreachable through an `exports` map that carries only `types`+`import`). Round 2 also asked to drop the dependency name; it stayed. Two facts, both measured on this head, are why that matters:

- **The strictest declared floor in the production closure is `>=20.6.0`, and it is not jose's.** `pnpm ls --prod --depth Infinity` → 146 packages; the only ones whose `engines` reject any 20.x are 33 `@opentelemetry/*` packages, every one `^18.19.0 || >=20.6.0`. The direct one is `@opentelemetry/exporter-trace-otlp-http@0.215.0`, imported at module load by `src/api/observability-api/observability-api.ts:13` and re-exported from `src/index.ts:9`; the other 32 arrive through it and `@traceloop/node-server-sdk@0.26.0`. Everything else in the closure admits 20.0.0.
- **`>=20.0.0` is nevertheless accurate for what the SDK uses — I am not asking for 20.6.** On Node **20.0.0** and **20.5.1** (the first and last releases OTel rejects) the built `dist/index.js` imports (85 exports), `Payments.getInstance(...).getAccountAddress()` decodes a key through jose 6, and an `OTLPTraceExporter` + `BasicTracerProvider` completes a real HTTP export round-trip to a local listener. OTel's `20.6.0` is declared, not exercised, by anything this SDK does.

What a reader on 20.0–20.5 actually sees is package-manager-dependent: `npm install` (npm 9.8 on Node 20.5.1) prints **112 `EBADENGINE`** warnings, every one naming an `@opentelemetry/*` package and `^18.19.0 || >=20.6.0` — none naming jose; `pnpm install` from a lockfile (pnpm 10.33) prints **nothing** for dependencies, because the headless path checks only the root project's `engines` (that check does fire — re-running with `>=20.19.0` produced the WARN). Either way the parenthetical points them at the wrong package, and it rots the day the single `decodeJwt` call (`src/api/base-payments.ts:1`) is replaced.

Suggested: `**Node.js**: Version 20 or higher` — nothing after it. The derivation lives in the commit body, which is where it belongs. Non-blocking.

<sub>🤖 pr-review-toolkit · comment-analyzer + own measurement · high</sub>

package.json
```diff
@@ -5,6 +5,9 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "type": "module",
+  "engines": {
+    "node": ">=20.0.0"
```
r-marques, line 9: **✅ NICE — `>=20.0.0` is now stated once, consistently, everywhere the repo publishes it — and the number is measured, not inferred**

Verified on `d6c22f8`: the three places round 2 named agree (`package.json:9`, `cli/package.json:66`, `markdown/installation.md:24` — all `20`), the three published pages round 2 found agree (`cli/docs/getting-started.md:15`, `openclaw/docs/getting-started.md:14`, `openclaw/docs/guide.md:43`, each keeping its page's own format), and a whole-tree grep (excluding lockfiles and CHANGELOG) finds no other Node-version statement: no `.nvmrc`/`.node-version`/`.tool-versions`, every CI workflow runs `24.x`, no `engine-strict` anywhere, `@types/node` is `^20.11.x` in root and CLI. The only `18` left is `cli/CLI_PHASE2_TESTING.md:319` (`node-version: '18'` in a sample workflow) — outside `cli/package.json` `files[]` and not copied by `release.yml`, so unpublished; bump when convenient.

The built SDK runs on Node 20.0.0 and 20.5.1 (details on `installation.md:24`), so the floor admits nothing that breaks. Gates on this head: `pnpm install --frozen-lockfile` 0, `pnpm build` 0, `pnpm typecheck:tests` 0, `pnpm test:unit` **49 suites / 643 passed / 1 skipped** (jest, run locally), prettier clean on both manifests; `pnpm-lock.yaml` and `tests/jest.config.json` are byte-identical to round-1 head `d956807`. All 9 CI checks green on `d6c22f8`, E2E included this time (round 2's red was a staging 502).

<sub>🤖 pr-review-toolkit · own verification · high</sub>

r-marques (commented):
## 🤖 Automated PR review — round 3 — ✅ Ready to merge

Both round-2 should-fixes are closed by the one new commit (`d6c22f8`, six one-line edits). The Node floor now reads `>=20.0.0` in all three places (`package.json:9`, `cli/package.json:66`, `markdown/installation.md:24`) and the reason on the page produces exactly that number; the three published pages that still promised Node 18 (`cli/docs/getting-started.md:15`, `openclaw/docs/getting-started.md:14`, `openclaw/docs/guide.md:43`) now say 20, and a whole-tree grep finds nothing else. The number is measured rather than inferred: the built SDK imports, decodes a key through jose 6, and completes a real OTLP export round-trip on Node **20.0.0** and **20.5.1**. Two nits remain, neither of which moves the verdict.

**Floor answer, with evidence.** The strictest `engines` in the 146-package production closure is `^18.19.0 || >=20.6.0`, declared by 33 `@opentelemetry/*` packages (direct: `@opentelemetry/exporter-trace-otlp-http@0.215.0`, on the import path via `observability-api.ts:13` → `index.ts:9`). It is declared, not exercised — the exporter runs on 20.0.0 — so `>=20.0.0` is the accurate statement for this SDK and I am **not** asking for 20.6. Note round 2's "let the exporter self-report on 20.0–20.5" holds under npm (112 `EBADENGINE` on 20.5.1) but **not** under pnpm from a lockfile (headless install checks only the root project's `engines`; control confirmed the root check itself fires). The detail is on the inline comment at `installation.md:24`.

**Round-2 disposition**
- should-fix (three published pages still Node 18): **fixed** — all three say 20; the only `18` left in the tree is `cli/CLI_PHASE2_TESTING.md:319`, a sample workflow outside `cli/package.json` `files[]` and not copied by `release.yml`, so unpublished (optional).
- should-fix (`20.19` vs the stated reason): **fixed** — `>=20.0.0` everywhere, prose reason matches, commit body's derivation correct. Residual nit: the parenthetical still names jose as the reason, and jose is not the strictest declared floor (inline).
- nit (PR title omits the Node floor): **untouched** — title and body identical to round 2; carried forward (inline on `package.json:8`).
- Round-1 nits (`useESM` inert / a2a half inert; prettier drift in `tests/jest.config.json`; jose 6 stricter decoder unpinned): not addressed; those files are byte-identical to round 1 and the items were wording/optional — not re-posted.

**What I reviewed**
- Diff range `origin/main...HEAD` — 8 files, +44/−22, head `d6c22f88`. Delta since round 2: `git diff 5227016f..d6c22f8` = 6 files, +6/−6, all doc/manifest text; `pnpm-lock.yaml` and `tests/jest.config.json` unchanged since round-1 head `d956807`.
- Panel: code-reviewer + comment-analyzer (read-only, reports retrieved by file, timestamps and content checked against this head). Every claim below re-verified by hand.
- Whole-tree grep for any `18` within 30 chars of `node`/`Node` (lockfiles and CHANGELOG excluded): only the new "dropped Node 18" phrasing and the unpublished sample above. No `.nvmrc`/`.node-version`/`.tool-versions`; every CI workflow `node-version: '24.x'`; no `engine-strict`; `@types/node` `^20.11.x` in root and CLI.
- Read `engines.node` off every package in `pnpm ls --prod --depth Infinity` (146) with the installed `semver`, probing 20.0.0 / 20.5.1 / 20.6.0 / 20.18.3 / 20.19.0 — only `@opentelemetry/*` rejects any 20.x. (`^20.19.0 || ^22.13.0 || >=24` in the lockfile is `eslint-visitor-keys@5.0.1`, dev-only.)
- Downloaded Node 20.0.0 and 20.5.1 into scratch and ran the built `dist/index.js` on each: 85 exports, `getAccountAddress()` decodes, `OTLPTraceExporter` → local HTTP sink receives 1 request. Ran `pnpm install --frozen-lockfile` on 20.5.1 against this manifest (exit 0, no WARN) and against round 2's `>=20.19.0` as a control (WARN fires); `npm install --legacy-peer-deps` on 20.5.1 → 112 `EBADENGINE`, all `@opentelemetry/*`.
- Commit body says the three pages "ship inside the npm tarballs": only the two OpenClaw pages do (`openclaw/package.json` `files[]` has `docs/`); the CLI page reaches users via the docs site (`release.yml:527`). Message-only, nothing to change.
- Gates on `d6c22f8`: `pnpm install --frozen-lockfile` 0 · `pnpm build` 0 · `pnpm typecheck:tests` 0 · `pnpm test:unit` 49 suites / 643 passed / 1 skipped (jest, local) · `prettier --check` clean on `package.json` and `cli/package.json` (the `.md` files fail only on pre-existing front-matter quoting, identical on `origin/main`, untouched lines). All 9 CI checks green, E2E included; auto-merge off; PR head on GitHub = `d6c22f8`.

**Verdict:** ✅ Ready to merge — 0 blockers, 0 should-fix, 2 nits.

### 🧭 Inline comments (3)

- 💡 **NIT** — PR title and body still say nothing about the Node floor — unchanged since round 2 — `package.json:8`
- 💡 **NIT** — The number is right; the parenthetical names jose as the reason, and jose is neither the only nor the strictest floor in the tree — `markdown/installation.md:24`
- ✅ **NICE** — `>=20.0.0` is now stated once, consistently, everywhere the repo publishes it — and the number is measured, not inferred — `package.json:9`

r-marques (approved):
🟡 **Approved with comments** — no blockers; 0 should-fix, 2 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/451#pullrequestreview-5199749791

PR #452, opened 2026-09-14
fix(cli): resolve modules as Node16, and take TypeScript 6

eruizgar91:
## Problem

`cli/tsconfig.json` paired `module: ES2022` with `moduleResolution: node` — node10 resolution — which TypeScript 6 rejects outright:

```
tsconfig.json(6,25): error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

That is what fails `CLI Sync Check` on #326 (`typescript` 5.9.3 → 6.0.3 in `/cli`).

The CLI is `"type": "module"` and already writes explicit `.js` extensions on relative imports, and both siblings — the root SDK and `openclaw/` — already build with `Node16`/`Node16`. The CLI was the outlier, not a special case, so it is aligned rather than silenced with `ignoreDeprecations`.

## Why 6 and not 7

TypeScript's `latest` on npm is **7.0.2** today, so it is worth saying why this stops at 6: **ts-jest does not support 7**. The current 29.4.12 declares `typescript: ">=4.3 <7"`, and under 7.0.2 every CLI suite dies before running:

```
TypeError: Cannot read properties of undefined (reading 'fileExists')
  at ConfigSet._resolveTsConfig (node_modules/.../ts-jest/dist/legacy/config/config-set.js:516:97)
```

6.0.3 matches what `openclaw/` already uses. The same ceiling applies to the root SDK, which is still on `^5.3.3`.

## Verification

With the local SDK symlinked into `cli/node_modules` the way both CI workflows do it:

| | result |
|---|---|
| `pnpm build` (tsc 6.0.3) | clean |
| `pnpm sync-check` | `✅ CLI is in sync with SDK API — All 44 commands match` |
| `pnpm test` (unit) | 4 suites, 33 passed |
| `pnpm test` (integration) | 2 suites, 28 passed |

Supersedes #326, which bumps the same dependency and cannot build on the current tsconfig.


r-marques:
:eyes: Reviewing

cli/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
     "target": "ES2022",
-    "module": "ES2022",
+    "module": "Node16",
```
r-marques, line 4: **🟡 SHOULD FIX — Node16 makes ts-jest print TS151002 14x per run, and its advice breaks the suite — silence it in jest.config.js with a comment saying why**

Measured on this head: `pnpm test` now prints `ts-jest[config] (WARN) TS151002: Using hybrid module kind (Node16/18/Next) is only supported in "isolatedModules: true". Please set "isolatedModules: true" in your tsconfig.json` **14 times** (0 on `main`). It fires purely because this line now says `Node16` — ts-jest's non-ESM path rewrites the effective options to `module: CommonJS` + `moduleResolution: Node10` anyway (`ts-jest@29.4.9` `dist/legacy/compiler/ts-compiler.js:110-121`, `fixupCompilerOptionsForModuleKind`) and raises the message once per compiled file at `:157-167`.

The trap is that the message's advice is the sibling setting (root and `openclaw/tsconfig.json` both have `isolatedModules: true`), so the next person will follow it — and **it breaks all 4 unit suites here**: with `isolatedModules: true` ts-jest switches to its hybrid transpiler, which emits ESM for this `"type": "module"` package while `cli/jest.config.js` runs CJS (`useESM: false`) → `SyntaxError: Cannot use import statement outside a module`, `Tests: 0 total`. Measured, then reverted.

What does work (measured: 4+2 suites, 61/61 pass, zero warnings, and ts-jest still type-checks `cli/test/**` — a `const x: number = 'nope'` mutation in `test/unit/simple.test.ts` fails with TS2322 both before and after):

```js
// cli/jest.config.js
transform: {
  '^.+\\.tsx?$': ['ts-jest', {
    useESM: false,
    // TS151002 fires because tsconfig says module: Node16, but ts-jest's CJS path
    // rewrites the effective options to CommonJS/node10 (fixupCompilerOptionsForModuleKind).
    // Do NOT follow its advice: isolatedModules: true makes ts-jest emit ESM for this
    // "type": "module" package and every suite fails with "Cannot use import statement".
    diagnostics: { ignoreCodes: [151002] },
  }],
},
```

The comment is the valuable half — it is the only thing that explains why the CLI cannot take the `isolatedModules: true` that its two siblings have.

<sub>🤖 pr-review-toolkit · parent + pr-test-analyzer · high</sub>

cli/package.json
```diff
@@ -60,7 +60,7 @@
     "ts-morph": "^22.0.0",
     "ts-node": "^10.9.2",
     "tsx": "^4.7.1",
-    "typescript": "^5.4.3"
+    "typescript": "^6.0.3"
```
r-marques, line 63: **💡 NIT — For the eventual TS 7 bump: lifting ts-jest's `<7` peer is necessary but likely not sufficient for the CLI**

Confirmed the ceiling this line stops at: in a scratch copy on `typescript@7.0.2`, `tsc --noEmit` passes (0 errors) but every CLI jest suite dies in `ConfigSet._resolveTsConfig` with `TypeError: Cannot read properties of undefined (reading 'fileExists')` — exactly the stack in the PR body. `ts-jest@29.4.9` (locked) and `29.4.12` (npm latest) both declare `typescript: '>=4.3 <7'`.

One thing worth knowing when that bump comes, since it is invisible from the tsconfig: ts-jest's non-ESM path (`useESM: false` in `cli/jest.config.js`) forces `moduleResolution: Node10` regardless of what this tsconfig says (`ts-jest@29.4.9` `dist/legacy/compiler/ts-compiler.js:110-121`), and node10 is the resolution TS 7 removes (the TS5107 text this PR fixes: *"will stop functioning in TypeScript 7.0"*). So a ts-jest that merely accepts `typescript@7` may still not run this suite; the CLI's jest half will likely need `useESM: true` + `extensionsToTreatAsEsm` (the shape `openclaw/jest.config.js` already uses) at the same time. Nothing to change here — a line in the PR body or a code comment would save the next person a detour.

<sub>🤖 pr-review-toolkit · parent + code-reviewer · medium</sub>

cli/tsconfig.json
```diff
@@ -1,9 +1,9 @@
...
     "lib": ["ES2022"],
-    "moduleResolution": "node",
+    "moduleResolution": "Node16",
```
r-marques, line 6: **✅ NICE — The right fix, not a silence: required pair, sibling-consistent, and a byte-identical emit**

Verified rather than taken on trust:
- `moduleResolution: node` is node10, which TS 6.0.3 rejects with TS5107 unless `ignoreDeprecations: "6.0"` — that is the exact failure in #326's `CLI Sync Check` log. `Node16`/`Node16` is the only pair TS accepts (TS5110 otherwise) and neither value is in TS 6.0's deprecation table, so the two keys this PR touches do not come back as a 7.0 problem.
- Root `tsconfig.json:3,5` and `openclaw/tsconfig.json:3,5` are both `Node16`/`Node16`, as the PR body says; `tests/tsconfig.json` (`Bundler`) is the only other config and is a test-tree config, not a build.
- `tsc --noEmit -p cli/tsconfig.json` on 6.0.3: exit 0, 0 errors, 0 deprecation notices. Every relative import in `cli/src/**` already carries `.js`; no `require`, no JSON imports, `import.meta.url` only in the tsx-run generator scripts; every dependency resolves under Node16 `exports`/`types` (`chalk@5` string `exports`, `cli-table3` `export =` default-import, `@types/inquirer` default export, SDK `exports["."]` with `types` + `import`).
- The compiled `cli/dist/**` (`.js` **and** `.d.ts`) is byte-for-byte identical to a `main` build on TS 5.9.3 / `ES2022`+`node` — this is a resolution-mode change with zero runtime effect, and `node bin/run.js --version` / `plans --help` run from that build.
- Lockfile: only `typescript` moved (5.9.3 → 6.0.3); every other hunk is a peer re-key. `pnpm install --frozen-lockfile` in `cli/` exits 0. `cli/` is standalone (no root `workspaces`, no `pnpm-workspace.yaml`), so the root lockfile is correctly untouched. `typescript@6.0.3` `engines.node >=14.17` vs the CLI's `>=18` is fine.

<sub>🤖 pr-review-toolkit · parent + code-reviewer · high</sub>

r-marques (commented):

Moves `cli/tsconfig.json` off node10 resolution (`ES2022`+`node` → `Node16`+`Node16`) and takes `typescript` `^6.0.3`, which is what #326 (Dependabot's 5.9.3 → 6.0.3 in `/cli`) could not do on its own: its `CLI Sync Check` fails with exactly `TS5107: Option 'moduleResolution=node10' is deprecated`. The core change is right and complete — the pair is the one TS accepts, it matches the root SDK and `openclaw/`, and the compiled `cli/dist/**` (`.js` and `.d.ts`) is **byte-identical** to a `main` build, so this is a type-level resolution switch with no runtime effect. The one thing to fold before merge is small and new to this PR: ts-jest now prints `TS151002` 14 times per run, and following its advice (`isolatedModules: true`, which both siblings have) breaks every CLI suite here — see the inline comment on `cli/tsconfig.json:4` for the measured one-line fix.

**What I reviewed**
- Diff range `origin/main...HEAD` — 3 files, +55/−55, head `02a080a1`. `cli/pnpm-lock.yaml`: only `typescript` moved (5.9.3 → 6.0.3), every other hunk is a peer re-key.
- Panel: `pr-review-toolkit:code-reviewer`, `pr-review-toolkit:pr-test-analyzer`; all commands below run by me (Bash) in an isolated worktree with the local SDK build symlinked into `cli/node_modules` the way `testing.yml`'s `cli_sync_check` does.
- **head (TS 6.0.3):** `tsc --noEmit -p cli/tsconfig.json` exit 0, 0 errors, 0 deprecation notices · `pnpm build` clean · `pnpm sync-check` "All 44 commands match" · `pnpm test` = unit 4 suites / 33 tests + integration 2 suites / 28 tests, all passing · `pnpm install --frozen-lockfile` exit 0. The integration tests spawn `node bin/run.js` against `dist/`, so the built ESM artifact is exercised, not only ts-jest's transpile.
- **main (TS 5.9.3) baseline, same procedure:** same 33 + 28 passing, `tsc --noEmit` clean — and **0** `TS151002` lines vs 14 on head. CLI `pnpm lint` is not in CI and is identical on both (100 problems, pre-existing, untouched files).
- **TS 7.0.2 measured in a scratch copy:** `tsc --noEmit` passes, but all 4 unit suites die in ts-jest `ConfigSet._resolveTsConfig` with `TypeError: Cannot read properties of undefined (reading 'fileExists')` — the PR body's stack, reproduced. `ts-jest@29.4.9` (locked) and `29.4.12` (npm latest) both declare `typescript: '>=4.3 <7'`, so "stops at 6 on purpose" holds.
- **TS 6.0 defaults:** every default that changed in 6.0 (`strict`, `module`, `target`, `types` → `[]`, `esModuleInterop`, `rootDir`) is set explicitly in `cli/tsconfig.json`, so none of them changes behaviour here; the only 6.0 deprecation the CLI hit was `node10`, which this PR removes rather than silences. `types: ["node","jest"]` stays load-bearing — ts-jest reuses this tsconfig for `cli/test/**`, which uses bare `describe`/`jest` globals.
- **CI coverage is real for this change:** `testing.yml` runs `on: [push]`, and its `cli_sync_check` job installs `cli/` with `--frozen-lockfile`, builds it with `tsc -b` (the only Node16 gate), runs `sync-check` and `pnpm test` — green on `02a080a1`. `cli-sync-and-test.yml` is path-filtered to root `src/**` and does not fire for this PR, which is fine.
- Not in scope / pre-existing, no comment left: 12 of the 18 `cli/test/**` spec files are run by no workflow (explicit file lists in `test:unit`/`test:integration` + `testPathIgnorePatterns`); `cli-basic.test.ts`'s in-test `pnpm build` swallows failures. Neither interacts with this change.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 1 should-fix, 1 nit.

### 🧭 Inline comments (3)

- 🟡 **SHOULD FIX** — Node16 makes ts-jest print TS151002 14x per run, and its advice breaks the suite — silence it in jest.config.js with a comment saying why — `cli/tsconfig.json:4`
- 💡 **NIT** — For the eventual TS 7 bump: lifting ts-jest's `<7` peer is necessary but likely not sufficient for the CLI — `cli/package.json:63`
- ✅ **NICE** — The right fix, not a silence: required pair, sibling-consistent, and a byte-identical emit — `cli/tsconfig.json:6`

r-marques (approved):

Inline review: https://github.com/nevermined-io/payments/pull/452#pullrequestreview-5198915589

eruizgar91:
Fixed in `3cafc99`.

Reproduced both halves before touching anything: 14 `TS151002` lines on this head, 0 on `main`; and with `isolatedModules: true` all four unit suites fail with `SyntaxError: Cannot use import statement outside a module`, `Tests: 0 total` — ts-jest emits ESM for this `"type": "module"` package while `cli/jest.config.js` runs CJS.

So `diagnostics: { ignoreCodes: [151002] }` in `cli/jest.config.js`, with the reason next to it: ts-jest's CJS path rewrites the effective options to CommonJS/node10 anyway, so the warning describes nothing that happens, and following its advice is what breaks the suite. That comment is also the only place recording why this package cannot take the `isolatedModules: true` its two siblings set.

After: zero TS151002 lines, 4 unit suites (33 tests) and 2 integration suites (28 tests) green.

cli/jest.config.js
```diff
@@ -18,6 +18,18 @@ export default {
...
+        // fires once per compiled file because cli/tsconfig.json says module: Node16.
+        // ts-jest's CJS path rewrites the effective options to CommonJS/node10 anyway
+        // (fixupCompilerOptionsForModuleKind), so the warning describes nothing that
+        // happens here.
+        //
+        // Do NOT follow its advice. With isolatedModules: true, ts-jest switches to its
+        // hybrid transpiler and emits ESM for this "type": "module" package while this
+        // config runs CJS, and every suite dies on "Cannot use import statement outside
+        // a module" — which is why the CLI cannot take the isolatedModules its two
+        // siblings (root SDK, openclaw) both set.
```
r-marques: **💡 NIT — Two clauses would make this comment fully precise: name the mechanism, and say when the line can go (it is not "once we go useESM")**

Non-blocking — the cause is right and this is the comment round 1 asked for. Two precisions, both checked against the `ts-jest@29.4.9` the lockfile pins (`cli/node_modules/ts-jest/dist/`):

1. **"switches to its hybrid transpiler"** (l.27) — that phrase is mine from round 1, and it is loose: ts-jest uses *hybrid* only for the Node16/18/Next **module kinds** (`utils/messages.js:12`, CHANGELOG 29.3.0 "support hybrid `module` values"), never for a transpiler. What actually happens with `isolatedModules: true` is that the language service is never built (`legacy/compiler/ts-compiler.js:74`) and `_transpileOutput` hands `tsTranspileModule` the **un-rewritten** `_initialCompilerOptions` (`ts-compiler.js:245-251`) — so the CommonJS/node10 fixup that l.23-25 describe no longer applies, and `getImpliedNodeFormatForFile` reads `"type": "module"` off `cli/package.json` (`transpilers/typescript/transpile-module.js:130-133`) → ESM out.

2. **When can this line be dropped?** (l.30-31) The natural reading is "once the CLI runs jest as ESM like its siblings" — and on its own that is wrong. The 151002 raise sits inside the language-service branch with **no** `isEsmMode` condition (`ts-compiler.js:157-167`), so `useESM: true` alone still prints it 14×; only `isolatedModules: true` removes it, and that also stops ts-jest type-checking `cli/test/**` (the transpile path reports syntactic/option diagnostics only). The root SDK pays for exactly that with `pnpm typecheck:tests` (root `CLAUDE.md`: "ts-jest transpiles without checking"); the CLI has no such script and `cli/tsconfig.json` excludes `test`, so this `diagnostics` block is today the **only** type gate the CLI specs have — the TS2322 mutation is what proves it is live.

Suggested replacement for l.27-31:

```js
// Do NOT follow its advice. With isolatedModules: true, ts-jest skips the language
// service and transpiles per file (tsTranspileModule) with the UN-rewritten options,
// so module: Node16 + "type": "module" makes it emit ESM while this config runs CJS,
// and every suite dies on "Cannot use import statement outside a module". Both
// siblings (root SDK, openclaw) survive isolatedModules only because they run jest
// as ESM (useESM: true). Drop this line only when the CLI does the same AND sets
// isolatedModules: true — and note that trades away the type-checking of
// cli/test/** this path gives today (the CLI has no typecheck:tests).
```

<sub>🤖 pr-review-toolkit · code-reviewer + parent · high</sub>

cli/jest.config.js
```diff
@@ -18,6 +18,18 @@ export default {
...
+        diagnostics: { ignoreCodes: [151002] },
```
r-marques, line 38: **✅ NICE — Round-1 should-fix fully addressed: the narrowest correct silence, in the place it takes effect, with type-checking of cli/test/** demonstrably retained**

Verified on `3cafc993` rather than taken on trust:
- **Placement is right.** It sits in the `['ts-jest', {…}]` transform tuple, which `legacy/ts-jest-transformer.js:86-93` migrates into the ts-jest `ConfigSet`; `config-set.js:286-300` **pushes** `151002` onto the default ignore list (`6059, 18002, 18003`) and leaves `throws: true` and `exclude: []` — byte-identical to the boolean default except for this one ts-jest-private code. `raiseDiagnostics` (`config-set.js:541-560`) filters on exact code equality, so no TypeScript diagnostic can be caught by it.
- **Measured:** `pnpm test` (local SDK symlinked into `cli/node_modules` the way `cli_sync_check` does) — **0** `TS151002` lines (14 on `02a080a1`), unit 4 suites / 33 tests + integration 2 suites / 28 tests, exit 0. CI's own `CLI Sync Check` log on this head: 0 (the `02a080a1` run had 14).
- **Type-checking retained:** `const x: number = 'nope'` in `cli/test/unit/simple.test.ts` → `pnpm test:unit` exits 1 with `error TS2322`, `Test Suites: 1 failed, 3 passed`. Reverted.
- The three causal claims in l.21-25 are true against the source: raised once per compiled file inside `getCompiledOutput` (`ts-compiler.js:157-167`); cause is `module: Node16` in `cli/tsconfig.json:4`; the CJS path rewrites to CommonJS + Node10 (`fixupCompilerOptionsForModuleKind`, `ts-compiler.js:110-121`, taken because `useESM: false`).
- The tsconfig-side alternative is not viable under TS 6.0.3: `moduleResolution: Node16` requires `module: Node16` (TS5109), and the only way to make ts-jest's fixup a no-op would be `node10`, which 6.0 deprecates as an error (TS5107 — the very failure this PR fixes).

r-marques (commented):
## 🤖 Automated PR review — round 2 — 🟡 Mergeable with nits

Round 2 on `3cafc993` = the round-1 head `02a080a1` + one commit touching **only** `cli/jest.config.js` (+12): `diagnostics: { ignoreCodes: [151002] }` inside the ts-jest transform tuple, with the reason inline. That is exactly the round-1 should-fix, done the narrowest way: `config-set.js:286-300` pushes the one ts-jest-private code onto the default ignore list and leaves `throws`/`exclude` untouched, so it removes a log line and nothing else. Measured: `pnpm test` goes from **14 → 0** `TS151002` with the same 33 + 28 passing, and a TS2322 mutation in `cli/test/unit/simple.test.ts` still fails the run — type-checking of `cli/test/**` is intact. The one inline note is optional wording on the comment (a mechanism label I supplied in round 1 is loose, and the retirement condition is not "once we go `useESM`").

**Round-1 disposition**
- Should-fix (`cli/tsconfig.json:4`, TS151002 × 14) — **addressed** in `cli/jest.config.js:32`. Verified, see the inline ✅.
- Nit (`cli/package.json:63`, TS 7 / `useESM`) — unchanged, non-blocking, nothing to do here. One refinement from this round: `useESM: true` alone will **not** retire the new line either — the 151002 raise (`ts-compiler.js:157`) has no `isEsmMode` condition; only `isolatedModules: true` does, and that drops ts-jest's type-check of the specs. Folded into the inline 💡 as the honest "drop this when…" clause.

**What I reviewed**
- Delta `02a080a1..3cafc993` — 1 file, +12/−0; `cli/tsconfig.json`, `cli/package.json`, `cli/pnpm-lock.yaml` byte-identical to round 1 (md5). Full PR vs `origin/main`: 4 files, +67/−55.
- Panel: `pr-review-toolkit:code-reviewer`, `pr-review-toolkit:pr-test-analyzer` (delta-scoped); every command below run by me in an isolated worktree with the local SDK build symlinked into `cli/node_modules/@nevermined-io/payments` the way `testing.yml`'s `cli_sync_check` does. Tree clean before and after.
- **head:** `tsc --noEmit -p cli/tsconfig.json` exit 0, 0 lines · `pnpm build` 0 · `pnpm sync-check` "All 44 commands match" · `pnpm test` exit 0, **0** `TS151002` (grep of the full log; 14 on `02a080a1`, 0 on `main`), unit 4 suites / 33 tests, integration 2 suites / 28 tests.
- **Mutation:** `const x: number = 'nope'` in `cli/test/unit/simple.test.ts` → `pnpm test:unit` exit 1, `error TS2322: Type 'string' is not assignable to type 'number'`, `Test Suites: 1 failed, 3 passed`. Reverted, `git status` clean.
- **Comment accuracy** checked sentence-by-sentence against `ts-jest@29.4.9` (`dist/legacy/compiler/ts-compiler.js:74,110-121,141-167,231-251`, `dist/transpilers/typescript/transpile-module.js:43-50,120-135`, `dist/legacy/config/config-set.js:286-320,541-566`, `utils/messages.js:12`) and against `tsconfig.json:19` / `openclaw/tsconfig.json:17` (both `isolatedModules: true`). Cause and remedy are right; the two precisions are inline.
- **CI on `3cafc993`:** `CLI Sync Check` green (44/44, 33 + 28), and its job log has **0** `TS151002` — the `02a080a1` job log had 14. `Lint & Build`, `Unit & Integration`, `OpenClaw Plugin Check`, `E2E` all green; auto-merge off.
- Pre-existing, no comment left (from `pr-test-analyzer`): with `useESM: false` ts-jest type-checks the specs under CommonJS/Node10, same branch `main`'s `ES2022`/`node` took, so the checker jest uses did not move — `tsc -b` (Node16) runs first in `cli_sync_check` for `src/`. And this `diagnostics` block is the CLI specs' only type gate (`cli/tsconfig.json` excludes `test`, no `typecheck:tests`); a `cli/tsconfig.tests.json` + `typecheck:tests` mirroring the root's would make that gate independent of one jest option — optional follow-up, not for this PR.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 0 should-fix, 1 nit (optional wording).

### 🧭 Inline comments (2)

- 💡 **NIT** — Two clauses would make this comment fully precise: name the mechanism, and say when the line can go (it is not "once we go useESM") — `cli/jest.config.js:31`
- ✅ **NICE** — Round-1 should-fix fully addressed: the narrowest correct silence, in the place it takes effect, with type-checking of cli/test/** demonstrably retained — `cli/jest.config.js:32`

r-marques (approved):
🟡 **Approved with comments** — no blockers; 0 should-fix, 1 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/452#pullrequestreview-5199359368

eruizgar91:
Wording nit folded in, head `0cd6913`.

You were right that "hybrid transpiler" was loose — `utils/messages.js:12` ties *hybrid* to the Node16/18/Next **module kinds**, and `assertCompilerOptionsWithJestTransformMode` is called from the language-service path at `ts-compiler.js:221`, with no `isEsmMode` guard. The comment now says what actually happens: with `isolatedModules: true` ts-jest never builds the language service and hands `tsTranspileModule` the un-rewritten options, so `module: Node16` plus `"type": "module"` yields ESM while this config runs CJS.

And the retirement condition is no longer "once we go useESM": it now says drop the line only when the CLI runs as ESM **and** sets `isolatedModules`, with the trade named — the transpile path reports syntactic and option diagnostics only, so this language-service path is currently the only type gate over `cli/test/**`, which has no `typecheck:tests` of its own.

Still 0 TS151002 lines, 33 + 28 passing.

cli/jest.config.js
```diff
@@ -18,6 +18,24 @@ export default {
...
+        // happens here.
+        //
+        // Do NOT follow its advice. With isolatedModules: true ts-jest never builds the
+        // language service; it transpiles per file through tsTranspileModule with the
+        // UN-rewritten options, so module: Node16 plus "type": "module" in
+        // cli/package.json makes it emit ESM while this config runs CJS, and every suite
+        // dies on "Cannot use import statement outside a module". Both siblings (root
+        // SDK, openclaw) get away with isolatedModules because they run jest as ESM.
+        //
+        // Drop this line only when the CLI runs as ESM too AND sets isolatedModules —
```
r-marques, line 34: **🟡 SHOULD FIX — The root SDK does not run jest as ESM — this sentence (which round 2 supplied) is wrong for it, and L34's "only" forecloses a route that keeps the type gate**

Non-blocking for the fix, which is right and stays (see the ✅ on L38). This is about the comment being the record of *why*, and one sentence of it is false — **and the sentence is mine**: round 2 suggested *"Both siblings (root SDK, openclaw) survive isolatedModules only because they run jest as ESM (useESM: true)"*, and you adopted it in good faith. Checked properly this round: it is true for openclaw and **wrong for the root SDK**, and L34's "only" inherits the error.

**openclaw really runs ESM.** `openclaw/package.json:23` sets `NODE_OPTIONS='--experimental-vm-modules'`; `openclaw/jest.config.js:3,5,13` is the `default-esm` preset + `extensionsToTreatAsEsm` + `useESM: true`. Node16 + `"type": "module"` + `isolatedModules` there emits ESM into an ESM runtime — the shape L27–32 describe.

**The root SDK does not.** Its `test` scripts are a bare `jest --verbose --config ./tests/jest.config.json` (`package.json`), and the only `--experimental-vm-modules` in the repo outside `node_modules` is openclaw's. Without that flag `vm.SyntheticModule` is undefined, so `jest-resolve/build/shouldLoadAsEsm.js:32,42-44` returns `false` for every file, jest-runtime takes `requireModule` with `supportsStaticESM: false` (`jest-runtime/build/index.js:916`), and ts-jest's `isEsmMode = useESM && supportsStaticESM` (`ts-jest/dist/legacy/compiler/ts-compiler.js:141`) is **false** — the root's `useESM: true` and `extensionsToTreatAsEsm` in `tests/jest.config.json` are inert. **Measured at this head** (`d78bd73`, post-#451 `main`): one root unit test with `--no-cache`, then the fresh entry in jest's transform cache — `"use strict"; Object.defineProperty(exports, "__esModule", …)`, one `require(`, zero `import` statements. CommonJS.

What actually lets the root survive `isolatedModules: true` is its **module kind**, not ESM: `tests/jest.config.json:22-25` overrides `module: "ESNext"` (merged before parse, `config-set.js:528-531`); `ESNext` is not a Node16-family kind (`transpilers/typescript/transpile-module.js:43-47`), so `_transpileOutput` takes the plain `ts.transpileModule` branch with `this._compilerOptions` — the **CJS-rewritten** options (`ts-compiler.js:237-243`) — and emits CJS into a CJS runtime. The CLI cannot copy that unchanged only because `cli/tsconfig.json:4` says `Node16`, which routes to `tsTranspileModule` with the un-rewritten options (`:245-251`) — exactly what L27–30 already say correctly.

That also makes L34's "**only** when the CLI runs as ESM too AND sets isolatedModules" too strong. The 151002 raise (`ts-compiler.js:157`) is gated on `isModernNodeModuleKind(this._initialCompilerOptions.module)`, not on ESM, so overriding the module kind in this transform removes it while **keeping** the language service — the type gate L35–37 defend. **Measured on this head:** replacing L38 with `tsconfig: { module: 'CommonJS', moduleResolution: 'node10' }` (the pair ts-jest's fixup already produces, `:110-121`) → 0 × TS151002, 4 suites / 33 tests green, and a `const x: number = 'nope'` in `test/unit/simple.test.ts` still fails the run with TS2322. Reverted. I am **not** asking to switch to that — rounds 1–2 accepted `ignoreCodes`, and the override would duplicate ts-jest's own fixup — only that the comment stop claiming there is exactly one exit.

Suggested replacement for L31–34 (L35–37 stay as they are):

```js
        // dies on "Cannot use import statement outside a module". The siblings avoid
        // this for different reasons: openclaw really runs jest as ESM
        // (--experimental-vm-modules), so ESM output is what it wants; the root SDK
        // does NOT run as ESM (no vm-modules flag, so jest loads CJS and its useESM is
        // inert) — its tests/jest.config.json overrides module to ESNext, a non-Node16
        // kind, so ts-jest takes the plain transpileModule branch with the CJS-rewritten
        // options.
        //
        // Drop this line when the CLI no longer runs ts-jest's language service under a
        // Node16 tsconfig — e.g. ESM + isolatedModules (openclaw's shape), or a
        // module/moduleResolution override in this transform's tsconfig (which keeps the
        // language service). Going isolatedModules trades away the type check:
```

<sub>🤖 pr-review-toolkit · code-reviewer + comment-analyzer + parent (measured) · high</sub>

cli/jest.config.js
```diff
@@ -18,6 +18,24 @@ export default {
...
+        // and note what that trades away: the transpile path reports syntactic and
+        // option diagnostics only, so this language-service path is currently the only
+        // type gate over cli/test/**, which has no typecheck:tests of its own.
```
r-marques, line 37: **💡 NIT — "over cli/test/**" reads as directory-wide; the gate reaches only the files a run transforms (6 of 18 in CI)**

Optional wording. "Only type gate" is right — `cli/tsconfig.json` includes `src/**/*` and excludes `test`, there is no `typecheck:tests` script, and the root's `typecheck:tests` covers `tests/**` + `../src/**`, never `cli/`. But "over `cli/test/**`" reads as directory-wide, and the language service only checks the files a jest run transforms (`ts-compiler.js:156,420-432`): CI's `test:unit` + `test:integration` name **6 of the 18** `*.test.ts` under `cli/test/`, and `testPathIgnorePatterns` (L8–10) keeps `config`/`plans`/`agents` out of even `test:all`. Something like "the only type check the CLI test files get, and only for the files each run compiles" says what the gate actually reaches.

<sub>🤖 pr-review-toolkit · code-reviewer + comment-analyzer · medium</sub>

cli/jest.config.js
```diff
@@ -18,6 +18,24 @@ export default {
       'ts-jest',
       {
         useESM: false,
+        // TS151002 ("hybrid module kind is only supported in isolatedModules: true")
```
r-marques, line 21: **💡 NIT — The quoted text is a paraphrase of TS151002, not the message**

Optional. The quoted text is a paraphrase, not the message — the string in `ts-jest/dist/utils/messages.js:12` is `Using hybrid module kind (Node16/18/Next) is only supported in "isolatedModules: true"`, so a grep for the quoted phrase finds nothing. Either drop the quotation marks or quote the first clause verbatim.

<sub>🤖 pr-review-toolkit · comment-analyzer · high</sub>

cli/jest.config.js
```diff
@@ -18,6 +18,24 @@ export default {
...
+        // UN-rewritten options, so module: Node16 plus "type": "module" in
+        // cli/package.json makes it emit ESM while this config runs CJS, and every suite
+        // dies on "Cannot use import statement outside a module". Both siblings (root
+        // SDK, openclaw) get away with isolatedModules because they run jest as ESM.
+        //
+        // Drop this line only when the CLI runs as ESM too AND sets isolatedModules —
+        // and note what that trades away: the transpile path reports syntactic and
+        // option diagnostics only, so this language-service path is currently the only
+        // type gate over cli/test/**, which has no typecheck:tests of its own.
+        diagnostics: { ignoreCodes: [151002] },
```
r-marques, line 38: **✅ NICE — Round-1 should-fix still fully addressed at the live head: 0 × TS151002, 33 + 28 green, narrowest silence, type gate demonstrably kept; mechanism paragraphs exact against ts-jest 29.4.9**

Round-1 should-fix remains addressed at the live head, re-measured on `d78bd73` (the rebase; `cli/jest.config.js`, `cli/tsconfig.json`, `cli/pnpm-lock.yaml` byte-identical to `0cd6913`, `cli/package.json` the same one `typescript` line):
- `pnpm test` with the local SDK symlinked the way `cli_sync_check` does: exit 0, **0** `TS151002` lines (14 on `02a080a1`), unit 4 suites / 33 tests, integration 2 suites / 28 tests. `tsc --noEmit -p cli/tsconfig.json` 0 errors, `pnpm sync-check` 44/44.
- Narrowest silence, verified in `config-set.js:287-300`: `ignoreCodes` is **pushed** onto the default list, `throws` stays `true`, `exclude` stays `[]`; `raiseDiagnostics` (`:541-560`) filters on exact code equality, and the 151002 raise passes no `filePath`, so `ignoreCodes` is the only knob that can reach it.
- Type gate still live: `const x: number = 'nope'` in `test/unit/simple.test.ts` → `pnpm test:unit` exit 1, `error TS2322`, `Test Suites: 1 failed, 3 passed`. Reverted.
- Mechanism paragraphs verified sentence-by-sentence against `ts-jest@29.4.9`: raised once per compiled file inside the language-service branch, gated on `isModernNodeModuleKind(this._initialCompilerOptions.module)` (`ts-compiler.js:157-168`); the CJS fixup to CommonJS/Node10 (`:110-121`, taken at `:141-142` because `useESM: false`); no language service under `isolatedModules` (`:74`); `tsTranspileModule` fed `_initialCompilerOptions` (`:245-251`) and reading `"type": "module"` via `impliedNodeFormat` (`transpile-module.js:131-138`); the transpile path reporting syntactic + options diagnostics only (`:151-154`). L21–30 and L35–36 are exact.

<sub>🤖 pr-review-toolkit · parent (measured) + code-reviewer + comment-analyzer · high</sub>

r-marques (commented):
## 🤖 Automated PR review — round 3 — 🟡 Mergeable with nits

Round 3 on **`d78bd73`**. The round-2 head `0cd6913` was rebased mid-review onto post-#451 `main` (`48ff17c`): the PR's three files are byte-identical (md5) and the `cli/package.json` hunk is the same one `typescript` line, so everything below was re-run on the live head. The delta since round 2 is one comment (`cli/jest.config.js:27-37`, +11/−5) above `diagnostics: { ignoreCodes: [151002] }`. Its mechanism paragraphs (L21–30, L35–36) are exact against `ts-jest@29.4.9` — every sentence maps to a line in `ts-compiler.js` / `transpile-module.js` / `config-set.js`, cited inline. One sentence is wrong, and it is the one round 2 handed you: **the root SDK does not run jest as ESM** (no `--experimental-vm-modules` anywhere but `openclaw/package.json:23`; `shouldLoadAsEsm.js:42-44` forces CJS, `isEsmMode` is false at `ts-compiler.js:141`; measured — its transform-cache output is `Object.defineProperty(exports, "__esModule"…)` with zero `import`s). It survives `isolatedModules` because `tests/jest.config.json` overrides `module: "ESNext"`, a non-Node16 kind that takes the CJS-rewritten `transpileModule` branch (`ts-compiler.js:237-243`). That also makes L34's "**only** ESM AND isolatedModules" too strong — a module-kind override in this transform removes the raise while keeping the language service (measured: 0 × TS151002, 33/33 green, TS2322 mutation still fails). One should-fix on the wording, with a drop-in replacement; the fix itself is unchanged and right.

**Round-2 disposition**
- Nit (`cli/jest.config.js:31`, comment precision) — **partially addressed**: the "hybrid transpiler" label is gone and the `tsTranspileModule` / un-rewritten-options mechanism is now stated exactly (verified, ✅ inline). The sibling clause I supplied in round 2 is false for the root SDK — corrected as the should-fix above; my error, not yours.

**Round-1 disposition**
- Should-fix (TS151002 × 14) — **remains addressed** on the live head, re-measured: `pnpm test` exit 0, **0** `TS151002`, unit 4 suites / 33 tests, integration 2 suites / 28 tests; `ignoreCodes` is the narrowest silence (`config-set.js:287-300` pushes onto the default list, `throws` and `exclude` untouched); the `const x: number = 'nope'` mutation still fails with TS2322, so ts-jest still type-checks the specs. Reverted, tree clean.
- Nit (`cli/package.json:63`, TS 7) — unchanged, non-blocking.

**What I reviewed**
- Delta `3cafc993..0cd6913` → 1 file, comment only; then `0cd6913` → `d78bd73` rebase, PR content unchanged. Full PR vs `origin/main`: 4 files, +73/−55.
- Panel: `pr-review-toolkit:code-reviewer`, `pr-review-toolkit:comment-analyzer`, delta-scoped, both reading `ts-jest@29.4.9` / `jest@29.7.0` source; both independently reached the root-SDK finding, and I re-derived it from the source and measured it before accepting it.
- Every command run by me in an isolated worktree at `d78bd73`, bootstrapped the way `testing.yml`'s `cli_sync_check` does (root `pnpm install --frozen-lockfile` + `pnpm build`, `cli/` install, local SDK symlinked into `cli/node_modules/@nevermined-io/payments`): `pnpm build` 0 · `tsc --noEmit -p cli/tsconfig.json` 0 errors · `pnpm sync-check` 44/44 · `pnpm test` 0 × TS151002, 33 + 28 green.
- Route-B mutation (`tsconfig: { module: 'CommonJS', moduleResolution: 'node10' }` in place of L38): 0 × TS151002, 33/33, TS2322 still red. Root-SDK CJS probe: one unit test with `--no-cache`, transform-cache entry inspected. All mutations reverted; `git status` clean before and after.
- CI on `d78bd73`: `Lint & Build`, `Unit & Integration`, `CLI Sync Check`, `OpenClaw Plugin Check`, `E2E` all green; auto-merge (squash) enabled by the author.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 1 should-fix (comment wording, drop-in text supplied), 2 nits (optional).

### 🧭 Inline comments (4)

- 🟡 **SHOULD FIX** — The root SDK does not run jest as ESM — this sentence (which round 2 supplied) is wrong for it, and L34's "only" forecloses a route that keeps the type gate — `cli/jest.config.js:34`
- 💡 **NIT** — "over cli/test/**" reads as directory-wide; the gate reaches only the files a run transforms (6 of 18 in CI) — `cli/jest.config.js:37`
- 💡 **NIT** — The quoted text is a paraphrase of TS151002, not the message — `cli/jest.config.js:21`
- ✅ **NICE** — Round-1 should-fix still fully addressed at the live head: 0 × TS151002, 33 + 28 green, narrowest silence, type gate demonstrably kept; mechanism paragraphs exact against ts-jest 29.4.9 — `cli/jest.config.js:38`

PR #449, opened 2026-09-14
fix(ci): merge main into Dependabot branches, keeping their rebase alive

eruizgar91:
## Root cause

`dependabot-update-branches.yml` sent `gh api --method POST .../pulls/{n}/update-branch`. That endpoint is **PUT**, so every call returned `gh: Not Found (HTTP 404)`, and the trailing `|| echo "PR #$pr skipped (already up to date or conflict)"` printed the 404 as a benign line while the job reported success.

No Dependabot branch has ever been updated by it: run `34743328585` (2026-09-13) 404s on all 21 open PRs, and #409's head was still its 2026-07-20 commit after ~50 pushes to `main`. With `strict: true` on `main`, those branches can never merge; the PRs pile up until `open-pull-requests-limit` is reached and Dependabot stops opening new ones.

## Why not `@dependabot rebase`

It is the better mechanism — Dependabot regenerates the `pnpm-lock.yaml` conflicts a merge cannot resolve — and round 2 of review asked for it. It does not work from this job: **Dependabot refuses commands from GitHub App identities** (*"Sorry, only users with push access can use that command"*, dependabot/dependabot-core#9147, still open with confirmations through 2026-07). This workflow runs as the release bot, so its comment would be ignored while `gh pr comment` exits 0 — the same silent failure moved somewhere new. Every `@dependabot rebase` measured on this repo was posted by a human account, mine included; that is what round 3 of review caught.

Using a user PAT with push access would make it work, and that is a maintainer's call, not something to ship around.

## What this does instead

Keep `update-branch`, accept its documented cost (*"Dependabot will stop rebasing a pull request once extra commits have been pushed to it"* — an automation a conflicted PR cannot use anyway), and remove the silence:

- **PUT**, and any non-2xx fails the job with the response body in an `::error`.
- **202 means queued**, so poll `compare/{base}...{head}` until `behind_by == 0` instead of printing "updated" one step ahead of the evidence.
- **DIRTY PRs get a `::warning`** naming the PR rather than being counted as updated — a merge cannot fix them and this job cannot ask Dependabot to.
- **UNKNOWN is re-asked 3×** rather than dropped: a push to `main` puts every open PR into that state, and this job runs seconds after exactly that event.
- **Oldest first.** `strict: true` puts every PR BEHIND again after each merge, so newest-first would re-update the same few forever and never reach the tail.
[...]

PR #454, opened 2026-09-15
ci: sweep weekly, serve approved first, and let dev-only majors through

eruizgar91:
Three gaps left by #449, each visible in the queue as it stands today.

## Nothing wakes the job unless something merges

The `push` trigger chains — a merge is itself a push to `main`, which starts the next wave — but only while the PRs being served actually merge. A wave spent entirely on majors awaiting human review ends the chain, and the backlog then sits until a person lands something. That is the state the queue reached this morning.

A `schedule:` of Monday 07:00 UTC is the fallback: an hour after Dependabot's own weekly run (Monday 08:00 Europe/Madrid, per `dependabot.yml`), so the sweep sees whatever it opened that morning.

## Slots were spent on PRs that cannot merge

Ordering was least-recently-updated, and the oldest open PRs are precisely the stuck majors — the first real run of #449 updated four of them and one approved PR. Now: approved first, least-recently-updated within each half. Every wave that holds an approved PR ends in a merge, which is what starts the next one.

## Every major waited on a human, including dev-only ones

6 of the 7 majors blocked today touch only devDependencies — `jest`, `@types/jest`, `ts-morph`, `eslint-config-prettier`, `langchain`, `cohere-ai`. The published surface cannot change, and the full CI already covers the risk: build, lint, `typecheck:tests`, unit, integration and E2E, plus the CLI and OpenClaw suites. Those are approved; anything touching a runtime dependency is held as before.

**How the decision is made, and why not `dependency-type`.** `fetch-metadata` does not document what that output reports when one PR bumps several packages — and a grouped PR is exactly where a production dependency could ride along with dev ones. So the policy reads each name from `steps.metadata.outputs.dependency-names` against the manifests, which is deterministic whatever the action decides to report.

The manifests are read from `pull_request.base.sha`, never from the PR's own head. A Dependabot branch is untrusted content; letting it supply the manifest would let the PR answer its own question. Nothing is executed from the checkout and no credentials are persisted.

## Verification

The policy step, extracted and run against this repo's real manifests:

| update-type | dependency-names | approve |
|---|---|---|
| patch | `axios` | ✅ patch or minor |
| minor | `express` | ✅ patch or minor |
| major | `jest, @types/jest` | ✅ every package is a devDependency |
| major | `ts-morph` | ✅ every package is a devDependency |
| major | `inquirer` | ❌ runtime in `cli/package.json` |
| major | `jest, inquirer` | ❌ runtime in `cli/package.json` |
| major | `some-package-nobody-declares` | ❌ undeclared |
| major | *(empty)* | ❌ fetch-metadata named no dependency |

The mixed group is the one that matters: a dev package and a runtime package in the same PR is held.

Ordering verified separately against a mixed list — approved PRs first, original order preserved within each half.


r-marques:
:eyes: Reviewing

r-marques (commented):

Three targeted changes to the Dependabot pipeline: a weekly `schedule:` fallback for the branch sweep, approved-first ordering, and — the substantive one — widening auto-approval from patch/minor to majors whose every package is a devDependency, decided by reading the base branch's manifests rather than `fetch-metadata`'s `dependency-type`. The mechanics are sound: the base-`sha` checkout is the right call, nothing from the PR head executes, the bash is clean under `set -euo pipefail`, and the PR body's eight-row table reproduces exactly against the real manifests.

The blocker is what the approval is *gated on*. The comment says "the full CI (… plus the CLI and OpenClaw suites) is the gate", but `main`'s required contexts are `E2E`, `Lint & Build`, `Unit & Integration` only — `CLI Sync Check` and `OpenClaw Plugin Check` are advisory, and no required job installs, builds or tests `cli/` or `openclaw/` (no workspace; separate lockfiles). GitHub auto-merge waits for required checks only, so a `jest`/`ts-morph` major in a sub-package merges on root-only green with its own suite red. Three of the six majors the PR says it would approve today (#325, #340, #323) are exactly that. A five-minute protection change (make the two contexts required) removes the blocker.

Then four should-fixes, each one or two lines: `peerDependencies`/`optionalDependencies` are not counted as runtime, so `@langchain/langgraph`, `@modelcontextprotocol/sdk` and `langsmith` majors — the SDK's published peer contract — are approved as dev-only; an empty or unrecognised `update-type` now falls through to approval where the old `if:` held it; the Monday cron is "an hour after Dependabot" only in summer (same minute Nov–Mar); and `reason` is interpolated into `run:` where the sibling steps use `env:`. Two are decisions: whether to hold `typescript` and `@types/node` (`build: tsc` — they shape the published artifact), and how a held major signals a human (an idempotent label).

**What I reviewed**
- Diff range `origin/main...HEAD` — 2 files, +100/−8, head `7d92c786`; read with `.github/dependabot.yml`, `testing.yml` and the three manifests.
- Panel: code-reviewer (it extracted the policy step verbatim via PyYAML and ran it under `bash -eo pipefail`; read fetch-metadata's `output.ts` / `update_metadata.ts` / `verified_commits.ts`; queried branch protection; read #386's diff), plus my own execution of the step against the real manifests for the table's 8 rows and 8 more.
- Executed, not inferred: every table row; `@langchain/langgraph`, `@modelcontextprotocol/sdk`, `langsmith`, `typescript` → `approve=true`; `@nevermined-io/payments` → held (via `cli/` `dependencies`); `actions/checkout` → held (undeclared); `UPDATE_TYPE=''` and `UPDATE_TYPE=something-unrecognised` with `NAMES=jest` → `approve=true`; the jq ordering on a mixed fixture; Madrid 08:00 → 06:00 UTC (Sep) / 07:00 UTC (Dec, Mar).
- Verified against GitHub: required contexts on `main`; #325/#340/#323 open and `REVIEW_REQUIRED`; #386 changes only the dev pin for `@langchain/langgraph` and no `package.json` peer line.
- Not run: the workflow itself (a `push`/`pull_request` trigger cannot be exercised pre-merge — the file says so).

**Verdict:** 🔴 Blocked — 1 blocker, 6 should-fix, 4 nits.

### 🧭 Inline comments (14)

- 🔴 **BLOCKER** — The CLI and OpenClaw suites this approval leans on are NOT required checks — a cli/openclaw dev major auto-merges on root-only green — `.github/workflows/dependabot-auto-merge.yml:71`
- 🟡 **SHOULD FIX** — `peerDependencies` / `optionalDependencies` are not counted as runtime — a major on a dev+peer package is approved as "every package is a devDependency" — `.github/workflows/dependabot-auto-merge.yml:112`
- 🟡 **SHOULD FIX** — No explicit `semver-major` arm — an empty or unrecognised `update-type` now falls through and is APPROVED when the names are dev-only — `.github/workflows/dependabot-auto-merge.yml:88`
- 🟡 **SHOULD FIX** — `0 7 * * 1` is "an hour after Dependabot" only under CEST — from late October to late March it is the same minute — `.github/workflows/dependabot-update-branches.yml:18`
- 🟡 **SHOULD FIX** — `${{ steps.policy.outputs.reason }}` is interpolated straight into `run:` — the same file routes `PR_URL` through `env:` two steps later — `.github/workflows/dependabot-auto-merge.yml:134`
- 🟡 **SHOULD FIX** — "The SDK's published surface cannot change" is not strictly true for the toolchain — `typescript` and `@types/node` majors are approved today (your call) — `.github/workflows/dependabot-auto-merge.yml:69`
- 🟡 **SHOULD FIX** — A held major now has a computed `reason`, and nothing surfaces it to a human — `.github/workflows/dependabot-auto-merge.yml:127`
- 💡 **NIT** — `actions/checkout@v4` while `testing.yml` is on `@v7` — this pin manufactures a held actions PR on day one — `.github/workflows/dependabot-auto-merge.yml:61`
- 💡 **NIT** — Trim strips leading whitespace only, and the unquoted `for name in $NAMES` is subject to pathname expansion — `.github/workflows/dependabot-auto-merge.yml:107`
- 💡 **NIT** — "does not document what that output reports" — the docs don't, but the source does; say "undocumented", not "unknown" — `.github/workflows/dependabot-auto-merge.yml:74`
- 💡 **NIT** — Say explicitly that the `github-actions` ecosystem lands in the "undeclared → held" branch by design — `.github/workflows/dependabot-auto-merge.yml:101`
- ✅ **NICE** — The base-branch checkout is right, and nothing from the PR head executes — `.github/workflows/dependabot-auto-merge.yml:63`
- ✅ **NICE** — The bash holds up under `set -euo pipefail`, and the PR's verification table reproduces exactly — `.github/workflows/dependabot-auto-merge.yml:121`
- ✅ **NICE** — jq ordering is correct and conflicted approved PRs do not eat slots — `.github/workflows/dependabot-update-branches.yml:112`

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          fetch-depth: 1
+          persist-credentials: false
+
+      # Patch and minor are approved as before. Majors are approved only when
+      # every package they bump is a devDependency in every manifest that
+      # declares it — the SDK's published surface cannot change, and the full
+      # CI (build, lint, typecheck:tests, unit, integration, E2E, plus the CLI
+      # and OpenClaw suites) is the gate.
```
r-marques: **🔴 BLOCKER — The CLI and OpenClaw suites this approval leans on are NOT required checks — a cli/openclaw dev major auto-merges on root-only green**

`gh api repos/nevermined-io/payments/branches/main/protection` → `required_status_checks.contexts = ["E2E", "Lint & Build", "Unit & Integration"]`. `CLI Sync Check` and `OpenClaw Plugin Check` are not in it, and GitHub auto-merge waits for *required* checks only. None of the three required jobs touches the sub-packages: there is no `pnpm-workspace.yaml` (`cli/` and `openclaw/` carry their own lockfiles), root `build` is `tsc` over `src/**`, root `lint` is `eslint ./src`, root jest's `rootDir` is `tests/`. The only jobs that install those lockfiles, build those packages, run `pnpm sync-check` (ts-morph) and run their tests are the two non-required ones (`testing.yml:48`, `:92`).

So for a `jest` / `@types/jest` / `ts-morph` major in `/cli` or `/openclaw` the policy says approve (:130), auto-merge is enabled, and those suites can be red without blocking the merge. This is not hypothetical: of the six majors the PR body says it would approve today, **three** are exactly these — #325 (`jest`+`@types/jest` in `/cli`), #340 (same in `/openclaw`), #323 (`ts-morph` 22→28 in `/cli`), all `REVIEW_REQUIRED` right now.

Fix, either: (a) make `CLI Sync Check` and `OpenClaw Plugin Check` required before this merges (`gh api -X POST repos/nevermined-io/payments/branches/main/protection/required_status_checks/contexts -f 'contexts[]=CLI Sync Check' -f 'contexts[]=OpenClaw Plugin Check'`) and have this comment name which contexts are *required*; or (b) hold any major whose names are only declared in `cli/` / `openclaw/` until those jobs are required. (a) is the one the comment already promises.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+          # safe reading of "unknown" is "not dev".
+          runtime=""
+          IFS=','
+          for name in $NAMES; do
+            name="${name#"${name%%[![:space:]]*}"}"
+            [ -z "$name" ] && continue
+            seen=0
+            for manifest in package.json cli/package.json openclaw/package.json; do
+              [ -f "$manifest" ] || continue
+              if jq -e --arg n "$name" '.dependencies // {} | has($n)' "$manifest" >/dev/null; then
```
r-marques: **🟡 SHOULD FIX — `peerDependencies` / `optionalDependencies` are not counted as runtime — a major on a dev+peer package is approved as "every package is a devDependency"**

Root `package.json` lists `@langchain/langgraph` (dev `1.2.0`, peer `>=1.0.0 <2`), `@modelcontextprotocol/sdk` (dev `^1.24.3`, peer `>=1.25.0`) and `langsmith` (dev `^0.7.4`, peer `>=0.7.0`) in **both** sections; `openclaw/package.json` has `@nevermined-io/payments` the same way. The peer range is the published contract, and the SDK imports them at runtime (`src/mcp/core/server-manager.ts`, `src/x402/langsmith/spans.ts`, `src/x402/langchain/agent.ts`). Executed against the real manifests: `NAMES='@langchain/langgraph'` → `approve=true`; same for `@modelcontextprotocol/sdk` and `langsmith`. Dependabot bumps only the dev pin and leaves the peer line (#386's `package.json` hunk changes `"@langchain/langgraph": "1.2.0"` → `"1.4.4"` and touches no `peerDependencies` line outside the lockfile), so a langgraph 2.x would auto-merge with the contract still `<2`: CI tests only v2 while consumers are told v1 is supported and v2 is not — a state only a human should resolve. Nothing in today's queue trips it, hence should-fix. One line: `jq -e --arg n "$name" '(.dependencies // {}) + (.peerDependencies // {}) + (.optionalDependencies // {}) | has($n)'`, and extend the comment at :100-101.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+      - name: Decide whether this PR may be approved
+        id: policy
+        env:
+          UPDATE_TYPE: ${{ steps.metadata.outputs.update-type }}
+          NAMES: ${{ steps.metadata.outputs.dependency-names }}
+        run: |
+          set -euo pipefail
+
+          case "$UPDATE_TYPE" in
+            version-update:semver-patch|version-update:semver-minor)
```
r-marques: **🟡 SHOULD FIX — No explicit `semver-major` arm — an empty or unrecognised `update-type` now falls through and is APPROVED when the names are dev-only**

Before this PR only an explicit patch/minor matched and everything else was held. Now the `case` has two arms and a fall-through, so `UPDATE_TYPE=''` (fetch-metadata writes `''` when it cannot classify a version pair) and `UPDATE_TYPE='anything-else'` both reach the manifest scan. Executed: `UPDATE_TYPE='' NAMES=jest` → `approve=true`; `UPDATE_TYPE=something-unrecognised NAMES=jest` → `approve=true`. Rare for npm, but the step's whole purpose is to fail closed on what it cannot classify. Add `version-update:semver-major) ;;` and a `*)` arm that writes `approve=false` with the reason before the scan.

<sub>🤖 pr-review-toolkit · code-reviewer · medium-high</sub>

.github/workflows/dependabot-update-branches.yml
```diff
@@ -6,6 +6,16 @@ on:
...
+  # A push is the normal trigger, and it chains: each merge is itself a push,
+  # so one wave feeds the next. That chain only holds while the PRs being served
+  # actually merge — a wave spent entirely on majors awaiting human review ends
+  # it, and nothing wakes the job again until someone lands something.
+  #
+  # Monday 07:00 UTC is the fallback for exactly that: an hour after Dependabot's
+  # own weekly run (Monday 08:00 Europe/Madrid, see dependabot.yml), so the sweep
+  # sees whatever it opened that morning.
+  schedule:
+    - cron: '0 7 * * 1'
```
r-marques: **🟡 SHOULD FIX — `0 7 * * 1` is "an hour after Dependabot" only under CEST — from late October to late March it is the same minute**

Madrid 08:00 Monday = 06:00 UTC while CEST holds (last Sunday of March → last Sunday of October) and **07:00 UTC** the rest of the year (`date -u -d 'TZ="Europe/Madrid" 2026-12-07 08:00'` → 07:00). So for five months the sweep fires as Dependabot starts, not after it. Two further caveats the comment at :14-16 should not lean on: `schedule` runs are commonly delayed 10–30 min under load, and the PRs Dependabot opens that morning are created on current `main` and need no sweep anyway — what the schedule really buys is a weekly wake-up for approved-but-behind PRs when no push happened. `0 9 * * 1` is after Dependabot year-round; reword the comment to say what it is for. Daily (`0 9 * * *`) is worth considering, since the chain can end on any weekday and the cost is one `gh pr list`.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+          if [ -n "$runtime" ]; then
+            echo "reason=major touching runtime or undeclared packages:$runtime" >> "$GITHUB_OUTPUT"
+            echo "approve=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "reason=major, every package is a devDependency" >> "$GITHUB_OUTPUT"
+            echo "approve=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Report the decision
+        run: echo "approve=${{ steps.policy.outputs.approve }} (${{ steps.policy.outputs.reason }})"
```
r-marques: **🟡 SHOULD FIX — `${{ steps.policy.outputs.reason }}` is interpolated straight into `run:` — the same file routes `PR_URL` through `env:` two steps later**

`reason` embeds dependency names from `fetch-metadata`. Not exploitable as things stand — the job only runs for `dependabot[bot]`-authored PRs (:17), fetch-metadata reads only Dependabot's signed first commit, and npm / `owner/repo` action names admit no shell metacharacters — but it is the exact pattern GitHub's hardening guide names, it is inconsistent with :139 / :146, and the first ecosystem with a looser name charset inherits it. `env: { APPROVE: ${{ steps.policy.outputs.approve }}, REASON: ${{ steps.policy.outputs.reason }} }` and `run: echo "approve=$APPROVE ($REASON)"`.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+      - name: Checkout the base branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          fetch-depth: 1
+          persist-credentials: false
+
+      # Patch and minor are approved as before. Majors are approved only when
+      # every package they bump is a devDependency in every manifest that
+      # declares it — the SDK's published surface cannot change, and the full
```
r-marques: **🟡 SHOULD FIX — "The SDK's published surface cannot change" is not strictly true for the toolchain — `typescript` and `@types/node` majors are approved today (your call)**

`build: tsc`, so a `typescript` major shapes `dist/` and every `.d.ts` directly; CI proves the build compiles, not that the emitted declarations are what consumers on older TS can parse. `@types/node` is the floor the SDK is typechecked against — bumping 20→25 while `engines.node` stays `>=20` lets code use Node 22+ APIs with no type error, which is precisely the human decision (#319, `@types/node` 20→25, was approved by hand today). Executed: both come back `approve=true`. `@babel/*` is neither in the build nor in the jest transform (ts-jest), so it needs no hold. Recommendation: a two-name hold — `case "$name" in typescript|@types/node) runtime="$runtime $name(toolchain)";; esac` before the manifest scan, with one line saying why. Product call, so labelled should-fix rather than blocker.

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+                seen=1
+              fi
+            done
+            [ "$seen" = "0" ] && runtime="$runtime $name(undeclared)"
+          done
+          unset IFS
+
+          if [ -n "$runtime" ]; then
+            echo "reason=major touching runtime or undeclared packages:$runtime" >> "$GITHUB_OUTPUT"
+            echo "approve=false" >> "$GITHUB_OUTPUT"
```
r-marques: **🟡 SHOULD FIX — A held major now has a computed `reason`, and nothing surfaces it to a human**

The reason lives only in the log of a job that is green either way; someone scanning the PR list cannot tell "held: runtime" from "held: undeclared" from "not processed yet" — and the PR's own argument is that majors "wait on a human", who needs to know which. Do **not** comment (the job re-runs on every `synchronize`, and every sweep merge is one — a comment per run would pile up on each held major). An idempotent label does it: `gh pr edit "$PR_URL" --add-label needs-human-review` when `approve != 'true'` (`pull-requests: write` is already granted), optionally plus `echo "::notice title=Held::$REASON"` so the reason shows on the check run.

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
       # already did — and `required_approving_review_count` stays 1, so this
       # step's approval is what satisfies it. If bumps ever stall at
       # REVIEW_REQUIRED again, read that setting first.
+      # The policy is read from the BASE branch, never from the PR's own head:
+      # a Dependabot branch is untrusted content, and reading a manifest it
+      # carries to decide whether to approve it would let the PR answer its own
+      # question. `base.sha` pins this to what is already on main, and no
+      # credentials are persisted since nothing here pushes.
+      - name: Checkout the base branch
+        uses: actions/checkout@v4
```
r-marques: **💡 NIT — `actions/checkout@v4` while `testing.yml` is on `@v7` — this pin manufactures a held actions PR on day one**

The `github-actions` Dependabot group (`patterns: "*"`, no `update-types` filter) will bundle a v4→v7 bump for this file into next Monday's grouped PR, and a major inside a group holds the whole group. Pin `@v7`.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+          fi
+
+          # A name counts as runtime if ANY manifest lists it under
+          # `dependencies`. Absent everywhere counts as runtime too: it means
+          # the package is not declared where this workflow can see it, and the
+          # safe reading of "unknown" is "not dev".
+          runtime=""
+          IFS=','
+          for name in $NAMES; do
+            name="${name#"${name%%[![:space:]]*}"}"
```
r-marques: **💡 NIT — Trim strips leading whitespace only, and the unquoted `for name in $NAMES` is subject to pathname expansion**

Safe today (fetch-metadata joins with `', '`, so only leading spaces occur, and npm names cannot contain `*?[`), and a stray trailing space fails closed (`jest ` reads as undeclared → held — measured). Cheap hardening: `set -f` at the top, and trim both ends (`name="${name%"${name##*[![:space:]]}"}"`).

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+          persist-credentials: false
+
+      # Patch and minor are approved as before. Majors are approved only when
+      # every package they bump is a devDependency in every manifest that
+      # declares it — the SDK's published surface cannot change, and the full
+      # CI (build, lint, typecheck:tests, unit, integration, E2E, plus the CLI
+      # and OpenClaw suites) is the gate.
+      #
+      # This reads the manifests rather than `dependency-type`, because
+      # fetch-metadata does not document what that output reports when a PR
```
r-marques: **💡 NIT — "does not document what that output reports" — the docs don't, but the source does; say "undocumented", not "unknown"**

fetch-metadata's `output.ts` resolves `maxDependencyTypes()` as `direct:production` > `direct:development` > `indirect` across all names and `maxSemver()` as the highest bump — which is also what makes trusting `update-type` on a grouped PR at :88 sound. Reading the manifests is still the better choice (it does not depend on Dependabot's own classification); just don't send the next reader off to re-derive it.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+          esac
+
+          if [ -z "$NAMES" ]; then
+            echo "reason=major, and fetch-metadata named no dependency" >> "$GITHUB_OUTPUT"
+            echo "approve=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          # A name counts as runtime if ANY manifest lists it under
+          # `dependencies`. Absent everywhere counts as runtime too: it means
```
r-marques: **💡 NIT — Say explicitly that the `github-actions` ecosystem lands in the "undeclared → held" branch by design**

`actions/checkout`, `dependabot/fetch-metadata` → `(undeclared)` → held. That is the right outcome (an actions major runs new code with the release-bot token, and an actions PR edits this very file under `pull_request`) and unchanged from before — but the comment speaks only of npm manifests, and a future reader could "fix" the loop by teaching it about `.github/workflows`.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
       # REVIEW_REQUIRED again, read that setting first.
+      # The policy is read from the BASE branch, never from the PR's own head:
+      # a Dependabot branch is untrusted content, and reading a manifest it
+      # carries to decide whether to approve it would let the PR answer its own
+      # question. `base.sha` pins this to what is already on main, and no
+      # credentials are persisted since nothing here pushes.
+      - name: Checkout the base branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
```
r-marques, line 63: **✅ NICE — The base-branch checkout is right, and nothing from the PR head executes**

`ref: base.sha` + `fetch-depth: 1` + `persist-credentials: false`, and the only consumer of the checkout is `jq` over three JSON files. A useful property that falls out of fetch-metadata reading `commits[0]` (Dependabot's signed commit): the sweep's later `Merge main into …` commits do not change the policy's inputs, so a held major stays held and an approved one stays approved across sweeps.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,96 @@ jobs:
...
+              if jq -e --arg n "$name" '.dependencies // {} | has($n)' "$manifest" >/dev/null; then
+                runtime="$runtime $name($manifest)"
+                seen=1
+                break
+              fi
+              if jq -e --arg n "$name" '.devDependencies // {} | has($n)' "$manifest" >/dev/null; then
+                seen=1
+              fi
+            done
+            [ "$seen" = "0" ] && runtime="$runtime $name(undeclared)"
```
r-marques: **✅ NICE — The bash holds up under `set -euo pipefail`, and the PR's verification table reproduces exactly**

`[ … ] && …` lists at :108 and :121 are exempt from errexit and a final iteration ending in `seen=1` does not abort the loop (verified); `jq -e` errors inside `if` read as "not runtime / not seen" → undeclared → held, so a broken or missing manifest fails closed; empty and trailing fields are skipped. All eight table rows came back identical against the real manifests, including `jest, inquirer` → `inquirer(cli/package.json)`.

.github/workflows/dependabot-update-branches.yml
```diff
@@ -89,13 +99,18 @@ jobs:
...
+          # approved ones first means every wave that has one ends in a merge.
           prs=$(gh pr list \
             --repo "$REPO" \
             --state open \
             --search "author:app/dependabot sort:updated-asc" \
             --limit 100 \
-            --json number,headRefName \
-            --jq '.[] | "\(.number) \(.headRefName)"')
+            --json number,headRefName,reviewDecision \
+            --jq '[.[] | select(.reviewDecision == "APPROVED")] + [.[] | select(.reviewDecision != "APPROVED")]
```
r-marques, line 115: **✅ NICE — jq ordering is correct and conflicted approved PRs do not eat slots**

`+` binds tighter than `|`, so this parses as `([approved] + [rest]) | .[] | …`; on a mixed fixture the approved half comes first with original order preserved within each half and `null` in the second half. An approved-but-DIRTY PR (#348 today) is served first every wave, hits 409, and `continue`s without consuming `MAX_MERGES`. `github.event.repository.default_branch` is present on the `schedule` payload too.

r-marques (changes requested):
🔴 **Blocked** — 1 blocker(s) to address before merge. Details are inline on the changed lines.

Inline review: https://github.com/nevermined-io/payments/pull/454#pullrequestreview-5211484304

eruizgar91:
All four in `27abda1`, and the blocker was the important one — it undercut the argument the policy rests on, not just the code.

**🔴 The required contexts are three, not seven.** I wrote "the full CI is the gate… plus the CLI and OpenClaw suites", but `CLI Sync Check` and `OpenClaw Plugin Check` are not required contexts — a dev major in `/cli` or `/openclaw` would have auto-merged on root-only green, with the suites that actually exercise it gating nothing. Majors now approve only when `directory` is the root; elsewhere a human decides. If you would rather those two became required contexts, that is a branch-protection change and a better fix than my restriction — happy to raise it, but not to make it silently.

**🟡 peer and optional count as runtime now.** Only `dependencies` was checked. This SDK declares `@langchain/core` and `@langchain/langgraph` as peers, so a major on either read as "every package is a devDependency" — measured, it now holds.

**🟡 Explicit `semver-major` arm.** An empty or unrecognised `update-type` fell through to the dev-only check. It is its own arm now, and anything else is held by name rather than by omission.

**🟡 The cron was right for half the year.** Dependabot's Monday 08:00 Europe/Madrid is 06:00 UTC under CEST and 07:00 under CET, so `0 7 * * 1` went from an hour's margin to none at the October switch. Moved to `0 9 * * 1`: two hours in summer, one in winter.

Policy step extracted and run against the real manifests:

| update-type | names | directory | → |
|---|---|---|---|
| patch | `axios` | `/` | ✅ patch or minor |
| major | `jest, @types/jest` | `/` | ✅ root, dev-only |
| major | `ts-morph` | `/cli` | ❌ suite is not a required check |
| major | `jest` | `/openclaw` | ❌ suite is not a required check |
| major | `inquirer` | `/` | ❌ runtime in cli |
| major | `@langchain/core` | `/` | ❌ peerDependency |
| *(empty)* | `jest` | `/` | ❌ unrecognised update-type |
| `semver-unknown` | `jest` | `/` | ❌ unrecognised update-type |

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+          done
+          unset IFS
+
+          [ -n "$runtime" ] && hold "major touching runtime or undeclared packages:$runtime"
+
+          echo "reason=major in the root, every package dev-only" >> "$GITHUB_OUTPUT"
+          echo "approve=true" >> "$GITHUB_OUTPUT"
+
+      - name: Report the decision
+        run: echo "approve=${{ steps.policy.outputs.approve }} (${{ steps.policy.outputs.reason }})"
```
r-marques: **🟡 SHOULD FIX — `${{ steps.policy.outputs.reason }}` is interpolated straight into `run:` — the same file routes `PR_URL` through `env:` two steps later — untouched, and `reason` now carries `${DIRECTORY}` too**

**Carried from round 1 — untouched at `27abda1`.** The `reason` string now also embeds `${DIRECTORY}` (branch-derived by fetch-metadata, so still a safe charset) — the surface grew while the sibling steps at :159/:166 keep using `env:`.

`reason` embeds dependency names from `fetch-metadata`. Not exploitable as things stand — the job only runs for `dependabot[bot]`-authored PRs (:17), fetch-metadata reads only Dependabot's signed first commit, and npm / `owner/repo` action names admit no shell metacharacters — but it is the exact pattern GitHub's hardening guide names, it is inconsistent with :139 / :146, and the first ecosystem with a looser name charset inherits it. `env: { APPROVE: ${{ steps.policy.outputs.approve }}, REASON: ${{ steps.policy.outputs.reason }} }` and `run: echo "approve=$APPROVE ($REASON)"`.

<sub>🤖 pr-review-toolkit · round-2 · high</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          fetch-depth: 1
+          persist-credentials: false
+
+      # Patch and minor are approved as before. A major is approved only when
+      # all three of these hold:
+      #
+      # 1. It updates the ROOT manifest. Branch protection requires exactly
```
r-marques, line 70: **🟡 SHOULD FIX — `typescript` and `@types/node` majors still approve — the decision from round 1 is open (your call)**

**Carried from round 1 — untouched at `27abda1`.** Executed on this head: `NAMES=typescript DIRECTORY=/` → `approve=true`, same for `@types/node`. The new rule 1 correctly scopes majors to the root, where these two live; rule 2's "declared *only* under `devDependencies`" is exactly the shape that lets the build toolchain through. Not a blocker — a product call — but decide it explicitly rather than by omission.

`build: tsc`, so a `typescript` major shapes `dist/` and every `.d.ts` directly; CI proves the build compiles, not that the emitted declarations are what consumers on older TS can parse. `@types/node` is the floor the SDK is typechecked against — bumping 20→25 while `engines.node` stays `>=20` lets code use Node 22+ APIs with no type error, which is precisely the human decision (#319, `@types/node` 20→25, was approved by hand today). Executed: both come back `approve=true`. `@babel/*` is neither in the build nor in the jest transform (ts-jest), so it needs no hold. Recommendation: a two-name hold — `case "$name" in typescript|@types/node) runtime="$runtime $name(toolchain)";; esac` before the manifest scan, with one line saying why. Product call, so labelled should-fix rather than blocker.

<sub>🤖 pr-review-toolkit · round-2 · medium</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+              done
+              if jq -e --arg n "$name" '.devDependencies // {} | has($n)' "$manifest" >/dev/null; then
+                seen=1
+              fi
+            done
+            [ "$seen" = "0" ] && runtime="$runtime $name(undeclared)"
+          done
+          unset IFS
+
+          [ -n "$runtime" ] && hold "major touching runtime or undeclared packages:$runtime"
```
r-marques: **🟡 SHOULD FIX — A held major now has a computed `reason`, and nothing surfaces it to a human — untouched, and the new `hold()` is the natural place for it**

**Carried from round 1 — untouched at `27abda1`.** `hold()` (:98) now centralises every refusal, so one `gh pr edit "$PR_URL" --add-label needs-human-review` inside it (plus `echo "::notice title=Held::$1"`) covers all five hold reasons at once — and rule 1 just made "held" the *common* outcome for the sub-package majors that used to be silently left alone.

The reason lives only in the log of a job that is green either way; someone scanning the PR list cannot tell "held: runtime" from "held: undeclared" from "not processed yet" — and the PR's own argument is that majors "wait on a human", who needs to know which. Do **not** comment (the job re-runs on every `synchronize`, and every sweep merge is one — a comment per run would pile up on each held major). An idempotent label does it: `gh pr edit "$PR_URL" --add-label needs-human-review` when `approve != 'true'` (`pull-requests: write` is already granted), optionally plus `echo "::notice title=Held::$REASON"` so the reason shows on the check run.

<sub>🤖 pr-review-toolkit · round-2 · medium</sub>

r-marques (commented):
## 🤖 Automated PR review — 🟡 Mergeable with nits (round 2)

Round 2 on `27abda1`, one commit past round 1 (`ci: narrow the dev-major approval to what the required checks actually cover`). The blocker is fixed the honest way: a major is approved only when fetch-metadata's `directory` is the root — where the three required contexts actually run — so the `/cli` and `/openclaw` dev majors that would have auto-merged on root-only green (#325, #340, #323 in today's queue) now get a human, and the comment names the exact required contexts and why. The three should-fixes it was paired with are fixed too: `peerDependencies`/`optionalDependencies` count as runtime (the four peer packages now hold with the block named), `update-type` has an explicit `semver-major` arm and everything else holds by name, and the cron is `0 9 * * 1` with the CET/CEST arithmetic stated correctly. All of it re-executed against the real manifests on this head, not taken from the PR comment's table.

Three round-1 should-fixes are untouched and carry forward: `reason` is still interpolated into `run:` (and now embeds `${DIRECTORY}` too); `typescript`/`@types/node` majors still approve (a decision, not a defect — but decide it); and a held major still leaves no signal for the human it now waits for — the new `hold()` is the one-line place for a label. One new nit: `${DIRECTORY:-/}` reads an empty directory as root, the "unknown means root" shape the same commit removed from `update-type` (unreachable today, since an empty directory implies empty names). On Kike's open question — making `CLI Sync Check` and `OpenClaw Plugin Check` required contexts is the better long-term fix and a repo-settings call for Rodolphe; this PR's hold is correct on its own until then.

**What I reviewed**
- Round-1 findings archived to `findings.round1.json`; every blocker and should-fix re-checked against `27abda1` (2 files vs `main`; the round delta is the policy step and the cron comment).
- Executed, not inferred: the policy step extracted verbatim (PyYAML) and run under `bash -eo pipefail` with `GITHUB_OUTPUT` captured, for 17 `(update-type, names, directory)` inputs — the PR comment's eight plus `@langchain/langgraph`, `@modelcontextprotocol/sdk`, `langsmith`, `@langchain/core`, `typescript`, `@types/node`, `actions/checkout`, empty/unrecognised `update-type`, empty names, empty directory, and a `/cli` minor.
- Read fetch-metadata `v3` source (`update_metadata.ts`, `output.ts`) for how `directory` is derived and when it can be empty.
- CI on this head: `Lint & Build`, `Unit & Integration`, `E2E`, `CLI Sync Check`, `OpenClaw Plugin Check` all green; `auto-merge` skipping (not a Dependabot PR).

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 3 should-fix, 5 nits.

### 🧭 Inline comments (10)

- 🟡 **SHOULD FIX** — `${{ steps.policy.outputs.reason }}` is interpolated straight into `run:` — the same file routes `PR_URL` through `env:` two steps later — untouched, and `reason` now carries `${DIRECTORY}` too — `.github/workflows/dependabot-auto-merge.yml:154`
- 🟡 **SHOULD FIX** — `typescript` and `@types/node` majors still approve — the decision from round 1 is open (your call) — `.github/workflows/dependabot-auto-merge.yml:70`
- 🟡 **SHOULD FIX** — A held major now has a computed `reason`, and nothing surfaces it to a human — untouched, and the new `hold()` is the natural place for it — `.github/workflows/dependabot-auto-merge.yml:148`
- 💡 **NIT** — `${DIRECTORY:-/}` treats an EMPTY `directory` as the root — hold on empty for the same reason rule 3 holds on an empty `update-type` — `.github/workflows/dependabot-auto-merge.yml:115`
- 💡 **NIT** — `actions/checkout@v4` while `testing.yml` is on `@v7` — this pin manufactures a held actions PR on day one — `.github/workflows/dependabot-auto-merge.yml:61`
- 💡 **NIT** — Trim strips leading whitespace only, and the unquoted `for name in $NAMES` is subject to pathname expansion — `.github/workflows/dependabot-auto-merge.yml:129`
- 💡 **NIT** — "does not document what that output reports" — the docs don't, but the source does; say "undocumented", not "unknown" — `.github/workflows/dependabot-auto-merge.yml:86`
- 💡 **NIT** — Say explicitly that the `github-actions` ecosystem lands in the "undeclared → held" branch by design — `.github/workflows/dependabot-auto-merge.yml:134`
- ✅ **NICE** — Round-1 blocker closed — sub-package majors now wait for a human, and the comment states why with the exact required contexts — `.github/workflows/dependabot-auto-merge.yml:75`
- ✅ **NICE** — The other three round-1 should-fixes are fixed the way they were asked — `.github/workflows/dependabot-auto-merge.yml:109`

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+              echo "reason=patch or minor" >> "$GITHUB_OUTPUT"
+              echo "approve=true" >> "$GITHUB_OUTPUT"
+              exit 0;;
+            version-update:semver-major)
+              ;;
+            *)
+              hold "update-type '${UPDATE_TYPE:-<empty>}' is not one this policy recognises";;
+          esac
+
+          case "${DIRECTORY:-/}" in
```
r-marques: **💡 NIT — `${DIRECTORY:-/}` treats an EMPTY `directory` as the root — hold on empty for the same reason rule 3 holds on an empty `update-type`**

Executed: `UPDATE_TYPE=semver-major NAMES=jest DIRECTORY=''` → `approve=true`. It is not reachable today — fetch-metadata's `directory` is derived from the Dependabot branch name (`branchNameToDirectoryName` in `update_metadata.ts`: `/` for root, `/cli`, `/openclaw`) and is only empty when `updatedDependencies` is empty, in which case `NAMES` is empty too and :121 holds — but the default silently encodes "unknown means root" a few lines after the comment says "unknown means a human". `case "${DIRECTORY}" in /) ;; "") hold "fetch-metadata reported no directory";; *) hold …;; esac` makes the three rules read the same way.

<sub>🤖 pr-review-toolkit · round-2 · high</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
       # already did — and `required_approving_review_count` stays 1, so this
       # step's approval is what satisfies it. If bumps ever stall at
       # REVIEW_REQUIRED again, read that setting first.
+      # The policy is read from the BASE branch, never from the PR's own head:
+      # a Dependabot branch is untrusted content, and reading a manifest it
+      # carries to decide whether to approve it would let the PR answer its own
+      # question. `base.sha` pins this to what is already on main, and no
+      # credentials are persisted since nothing here pushes.
+      - name: Checkout the base branch
+        uses: actions/checkout@v4
```
r-marques: **💡 NIT — `actions/checkout@v4` while `testing.yml` is on `@v7` — this pin manufactures a held actions PR on day one**

**Carried from round 1 — untouched.** The `github-actions` Dependabot group (`patterns: "*"`, no `update-types` filter) will bundle a v4→v7 bump for this file into next Monday's grouped PR, and a major inside a group holds the whole group. Pin `@v7`.

<sub>🤖 pr-review-toolkit · round-2 · high</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+          [ -z "$NAMES" ] && hold "major, and fetch-metadata named no dependency"
+
+          # Runtime means any of dependencies / peerDependencies /
+          # optionalDependencies, in any manifest. Absent everywhere counts as
+          # runtime too: not declared where this workflow can see it, and the
+          # safe reading of "unknown" is "not dev".
+          runtime=""
+          IFS=','
+          for name in $NAMES; do
+            name="${name#"${name%%[![:space:]]*}"}"
```
r-marques: **💡 NIT — Trim strips leading whitespace only, and the unquoted `for name in $NAMES` is subject to pathname expansion**

**Carried from round 1 — untouched.** Safe today (fetch-metadata joins with `', '`, so only leading spaces occur, and npm names cannot contain `*?[`), and a stray trailing space fails closed (`jest ` reads as undeclared → held — measured). Cheap hardening: `set -f` at the top, and trim both ends (`name="${name%"${name##*[![:space:]]}"}"`).

<sub>🤖 pr-review-toolkit · round-2 · high</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+      # 2. Every package it bumps is declared *only* under `devDependencies` —
+      #    `dependencies`, `peerDependencies` and `optionalDependencies` all
+      #    count as runtime, because a peer or optional major changes what a
+      #    consumer resolves just as surely as a direct one does.
+      # 3. `update-type` is explicitly `version-update:semver-major`. Anything
+      #    else — an empty string, a value this action has not shipped yet — is
+      #    not a category to approve on: unknown means a human.
+      #
+      # The names are read off the base branch's manifests rather than
+      # `dependency-type`, because fetch-metadata does not document what that
```
r-marques: **💡 NIT — "does not document what that output reports" — the docs don't, but the source does; say "undocumented", not "unknown"**

**Carried from round 1 — the rewritten comment still says it.** fetch-metadata's `output.ts` resolves `maxDependencyTypes()` as `direct:production` > `direct:development` > `indirect` across all names and `maxSemver()` as the highest bump — which is also what makes trusting `update-type` on a grouped PR at :88 sound. Reading the manifests is still the better choice (it does not depend on Dependabot's own classification); just don't send the next reader off to re-derive it.

<sub>🤖 pr-review-toolkit · round-2 · high</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+          # safe reading of "unknown" is "not dev".
+          runtime=""
+          IFS=','
+          for name in $NAMES; do
+            name="${name#"${name%%[![:space:]]*}"}"
+            [ -z "$name" ] && continue
+            seen=0
+            for manifest in package.json cli/package.json openclaw/package.json; do
+              [ -f "$manifest" ] || continue
+              for block in dependencies peerDependencies optionalDependencies; do
```
r-marques: **💡 NIT — Say explicitly that the `github-actions` ecosystem lands in the "undeclared → held" branch by design**

**Carried from round 1 — untouched.** Executed on this head: `actions/checkout` → `(undeclared)` → held, and `DIRECTORY=/` for the actions ecosystem so rule 1 does not catch it first. `actions/checkout`, `dependabot/fetch-metadata` → `(undeclared)` → held. That is the right outcome (an actions major runs new code with the release-bot token, and an actions PR edits this very file under `pull_request`) and unchanged from before — but the comment speaks only of npm manifests, and a future reader could "fix" the loop by teaching it about `.github/workflows`.

<sub>🤖 pr-review-toolkit · round-2 · high</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+
+      # Patch and minor are approved as before. A major is approved only when
+      # all three of these hold:
+      #
+      # 1. It updates the ROOT manifest. Branch protection requires exactly
+      #    three contexts — `E2E`, `Lint & Build`, `Unit & Integration` — and
+      #    all three run against the root package. `CLI Sync Check` and
+      #    `OpenClaw Plugin Check` are NOT required, so a major in /cli or
+      #    /openclaw could auto-merge on root-only green, with the suites that
+      #    would actually exercise it never gating anything. Until those two are
```
r-marques, line 75: **✅ NICE — Round-1 blocker closed — sub-package majors now wait for a human, and the comment states why with the exact required contexts**

Executed on `27abda1` against the real manifests: `jest, @types/jest` in `/cli` and `/openclaw` and `ts-morph` in `/cli` → *"major in /cli — its suite is not a required check, so a human decides"*; the same names in `/` → approve. That is the three-of-six from today's queue (#325, #340, #323) held, and the root dev majors still served. `directory` is the right key: fetch-metadata derives it from the branch name, so a grouped PR in `/cli` reports `/cli`. On Kike's question — making `CLI Sync Check` and `OpenClaw Plugin Check` required contexts is the better long-term fix and it is a repo-settings decision, not this PR's; the hold is correct on its own until that lands.

<sub>🤖 pr-review-toolkit · round-2 verification · high</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,116 @@ jobs:
...
+            echo "approve=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          }
+
+          case "$UPDATE_TYPE" in
+            version-update:semver-patch|version-update:semver-minor)
+              echo "reason=patch or minor" >> "$GITHUB_OUTPUT"
+              echo "approve=true" >> "$GITHUB_OUTPUT"
+              exit 0;;
+            version-update:semver-major)
```
r-marques: **✅ NICE — The other three round-1 should-fixes are fixed the way they were asked**

Executed: `@langchain/langgraph`, `@modelcontextprotocol/sdk`, `langsmith`, `@langchain/core` → held with the block named (`package.json:peerDependencies`); `UPDATE_TYPE=''` and `'something-unrecognised'` → *"is not one this policy recognises"*; `NAMES=''` still held; patch/minor in `/cli` still approved (rule 1 is major-only, as it should be). `dependabot-update-branches.yml:21` is `0 9 * * 1` with a comment that now states the CET/CEST arithmetic correctly. `hold()` collapsing five refusals into one path is a good shape — and the reason strings are now specific enough to be worth surfacing (see the label item).

<sub>🤖 pr-review-toolkit · round-2 verification · high</sub>

r-marques (approved):
🟡 **Approved with comments** — no blockers; 3 should-fix, 5 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/454#pullrequestreview-5212467328

eruizgar91:
All three in `0f78cf4`.

**`reason` no longer reaches `run:` through `${{ }}`.** It is built from `directory` and `dependency-names`, both PR-controlled, and an expression is substituted before the shell sees the script. Through `env:` now, like `PR_URL` two steps down — which was the inconsistency worth pointing at.

**A held PR now says so where someone will see it.** `hold()` emits `::notice title=Dependabot PR held for review::<reason>` and a `**Held for review:**` line in the run summary. The job is green either way, so a step output was the wrong place for the one fact a human needs.

**`typescript` and `@types/node` hold.** Taking the call rather than leaving it open: they are devDependencies that decide what ships without shipping. `typescript` is the compiler every workspace builds and type-checks with, `@types/node` types the runtime surface the published `.d.ts` are emitted against, and #326 is the worked example — TypeScript 6 needed a `moduleResolution` change before it could even build, and ts-jest's peer range still excludes 7. Neither is a thing to learn about from a merge commit.

Exercised on this head:

| update-type | names | directory | → |
|---|---|---|---|
| major | `jest, @types/jest` | `/` | ✅ root, dev-only |
| major | `typescript` | `/` | ❌ toolchain + notice + summary |
| major | `@types/node` | `/` | ❌ toolchain + notice + summary |
| major | `jest, typescript` | `/` | ❌ toolchain (grouped) |
| major | `ts-morph` | `/cli` | ❌ directory rule |
| patch | `typescript` | `/` | ✅ patch or minor |

The grouped row is the one that matters: a toolchain major riding along with ordinary dev packages still holds.

aaitor (commented):
Round 3: the three requested fixes landed. One test-rigor finding remains inline.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -52,19 +52,148 @@ jobs:
...
+          IFS=','
+          for name in $NAMES; do
+            name="${name#"${name%%[![:space:]]*}"}"
+            case "$name" in
+              typescript|@types/node) toolchain="$toolchain $name";;
+            esac
+          done
+          unset IFS
+
+          [ -n "$toolchain" ] && hold "major on the toolchain:$toolchain"
```
aaitor: **🟡 SHOULD FIX — The new toolchain hold is not pinned by a committed test**\n\nThe current PR’s `auto-merge` job is skipping because this is not a Dependabot PR, so none of the green checks executes this policy with `typescript`, `@types/node`, or a grouped dependency. Deleting this guard would therefore leave the suite green and silently restore automatic approval for those toolchain majors. The manual six-case table verifies this head, but it cannot catch a later regression. Please add a fixture-driven test for `typescript`, `@types/node`, and `jest, typescript` (or extract this policy into a testable script) so the safety branch is mutation-resistant.

aaitor (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 0 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/454#pullrequestreview-5219119742

eruizgar91:
Fixed in `00ca007`, and you were right that the manual table could not carry this: it verified one head and nothing would have caught a later deletion.

The policy moved to `.github/scripts/dependabot-policy.sh` — same logic and inputs, plus a `POLICY_ROOT` so a test can point it at fixtures — and `.github/scripts/dependabot-policy.test.sh` runs it from `lint_build` on **every push**, not just Dependabot ones.

14 cases, including the three you named:

```
patch approves · minor approves · root dev-only major approves
typescript major holds · @types/node major holds · jest, typescript holds
/cli major holds · /openclaw major holds
runtime holds · peer holds · undeclared holds
empty update-type holds · unrecognised update-type holds · major with no names holds
```

Every held case also asserts the run-summary line, so the visibility added in `0f78cf4` cannot regress quietly either.

Fixtures rather than this repo's manifests, deliberately: a test that reads `package.json` starts passing for the wrong reason the day a dependency moves between blocks.

Mutation-checked, which is the part that answers your point:

```
$ # delete `[ -n "$toolchain" ] && hold ...`
$ .github/scripts/dependabot-policy.test.sh
FAIL typescript major holds: approve=true, expected false
FAIL @types/node major holds: approve=true, expected false
FAIL toolchain in a group holds: approve=true, expected false
dependabot-policy: 11 passed, 3 failed   → exit 1

$ # restore
dependabot-policy: 14 passed, 0 failed
```

.github/scripts/dependabot-policy.sh
```diff
@@ -0,0 +1,105 @@
...
+# the safe reading of "unknown" is "not dev".
+runtime=""
+IFS=','
+for name in $NAMES; do
+  name="${name#"${name%%[![:space:]]*}"}"
+  [ -z "$name" ] && continue
+  seen=0
+  for manifest in "$ROOT/package.json" "$ROOT/cli/package.json" "$ROOT/openclaw/package.json"; do
+    [ -f "$manifest" ] || continue
+    for block in dependencies peerDependencies optionalDependencies; do
```
aaitor, line 90: 🟡 **SHOULD FIX** — `optionalDependencies` is a distinct fail-closed branch, but the fixtures declare none and the 14 cases cover only direct and peer runtime dependencies. Removing `optionalDependencies` from this loop leaves the suite green and lets an optional-runtime major auto-approve as dev-only/unknown depending on its other declarations. Add an optional dependency fixture and a held case so every runtime block named by the policy is mutation-pinned.

aaitor (commented):
Round 4: the requested executable policy coverage landed. One unpinned runtime-dependency branch remains inline.

aaitor (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 0 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments/pull/454#pullrequestreview-5219595313

eruizgar91:
All of it in `664d364`. The `optionalDependencies` one was sharper than it read: the suite *had* a case for it, and the case was worthless — it asserted the reason contained `"runtime"`, and the held message is *"runtime or undeclared packages"*, so an optional dependency falling through to the **undeclared** branch matched just as well. Removing `optionalDependencies` from the loop left 17/17 green.

Each runtime case now asserts its own block, and the manifests are labelled `root` / `cli` / `openclaw` instead of by basename (which was the temp-dir name under test). Measured:

```
$ # drop optionalDependencies from the loop
FAIL optional major holds: reason '… fsevents(undeclared)' does not mention 'optionalDependencies'   → exit 1
$ # drop peerDependencies
FAIL peer major holds: reason '… @langchain/core(undeclared)' does not mention 'peerDependencies'    → exit 1
```

The rest:

- **Empty `directory` approved.** `${DIRECTORY:-/}` read it as the root. Held now, for the same reason an empty `update-type` is — fetch-metadata always sets it, so empty means something changed upstream. Case added; deleting the guard fails it.
- **`set -f`**, and the trim takes both ends.
- **`actions/checkout@v4` → `@v7`.** Good catch on the second-order effect: the `github-actions` group has no `update-types` filter, so that bump would have been bundled into next Monday's grouped PR and held the whole group on this one pin.
- **"The SDK's published surface cannot change" was wrong** and you were right to push on it — `build` is `tsc`, so a compiler major shapes `dist/` and every emitted `.d.ts`. That is precisely why `typescript` and `@types/node` hold, and the comment now says that rather than the general claim.
- **"undocumented", not "unknown"** — `output.ts` resolves `direct:production > direct:development > indirect`. The comment says so and keeps the manifest read for not depending on an implementation detail staying put.
- **github-actions → undeclared → held** is now stated as deliberate, with the reason (an actions major runs new code holding the release-bot token, and an actions PR edits this very workflow).

17 cases green.

PR #455, opened 2026-09-15
fix(mcp): advertise a tier-qualified authorization_endpoint in the MCP server's OAuth metadata

### LINKED ISSUE #447: fix(mcp): the SDK MCP OAuth metadata advertises a tier-blind authorization_endpoint — add ?network=sandbox|live
## What happens

`payments.mcp.start()` publishes its own OAuth Authorization Server metadata for the MCP server it hosts (`src/mcp/http/oauth-metadata.ts` → `buildOAuthUrls`):

```ts
authorizationUri: `${frontend}/oauth/authorize`,   // https://nevermined.app/oauth/authorize
tokenUri:         `${backend}/oauth/token`,        // https://api.sandbox.nevermined.app/oauth/token
```

So an MCP client that discovers this server is sent to a **tier-blind** consent URL, while the token endpoint is tier-specific. The Nevermined web app boots on whatever tier the user's browser last chose — **Live by default** — so a user of a `sandbox` MCP server who opens the consent link in a fresh browser profile is consented against the **live** API: the wrong `client_id` registry, the wrong wallet, and, for an account-level connector, *"Connector not authorized"*.

This is the same defect nevermined-io/nvm-monorepo#3430 fixes for the API's own RFC 8414 document. There, each tier's `authorization_endpoint` now carries the tier — `https://nevermined.app/oauth/authorize?network=sandbox|live` — and the web app's `/oauth/*` routes honour the param (auto-switch + reload at the entry route, tier badge on the consent screens). The SDK's document should say the same thing, since it knows its environment.

## Fix

In `buildOAuthUrls` / `getOAuthUrlsForEnvironment`, append `?network=<tier>` to `authorizationUri`, with the tier derived from the `EnvironmentName`:

| `EnvironmentName` | `network` |
| --- | --- |
| `sandbox`, `staging_sandbox` | `sandbox` |
| `live`, `staging_live` | `live` |
| `custom` | derive from the backend host (`api.sandbox.` / `api.live.`), else omit the param (the web app then keeps its current behaviour) |

- Param **name** is `network`, **values** `sandbox` / `live` — the embed widget's existing public param (nvm-monorepo#1787). Do not introduce `env`.
- Build the URL with `URL` + `searchParams` (not string concatenation) so anything later appended to it cannot produce a second `?`.
- RFC 6749 §3.1 lets the authorization endpoint carry a query component; MCP clients MUST retain it when adding their own parameters (the MCP TypeScript SDK does — `new URL(endpoint).searchParams.set(...)`).
- Update the `oauth-metadata` unit tests to pin the query component per environment, and `markdown/` docs that show the metadata.

## Acceptance

- `GET /.well-known/oauth-authorization-server` on a `sandbox` MCP server → `authorization_endpoint` ends in `/oauth/authorize?network=sandbox`; `live` → `?network=live`.
- Same value on `/.well-known/openid-configuration` (`getOidcConfiguration` shares `oauthUrls`).

Found while implementing nevermined-io/nvm-monorepo#3430 (docs: nevermined-io/docs#394).

r-marques:
> ✅ **Gate cleared (2026-09-18).** This PR was held draft until nevermined-io/nvm-monorepo#3515 — the consent page tolerating the official Python MCP client's second `?` on the advertised authorize URL — reached both tiers. It shipped in nvm v1.38.1; argocd has the webapp at v1.41.0 on prod and staging, and the served bundle at `nevermined.app` / `nevermined.dev` carries the tolerant parser. Rebased on `main` (CI-only commits, no overlap); `pnpm test:unit` (jest) 49 suites / 656 passed / 1 skipped, `pnpm build` green on this head. The self-review panel was posted and folded while it was a draft.

## Description

Closes #447 — the TypeScript twin of nevermined-io/nvm-monorepo#3430 (API + webapp, merged as `2fc26dc5c`).

The MCP server that `payments.mcp.start()` hosts publishes its own RFC 8414 / OIDC discovery documents (`buildAuthorizationServerMetadata`, `buildOidcConfiguration`, and the root `buildServerInfoResponse`) with `authorization_endpoint: ${frontend}/oauth/authorize` — **tier-blind**, although the SDK knows its environment. One Nevermined web app serves the consent screens for both API tiers and boots on whatever tier the user's browser last chose (Live by default), so a `sandbox` server's users landed on the **live** consent screen — wrong `client_id` registry, wrong wallet, and for an account-level connector *"Connector not authorized"*.

### What changed (`src/mcp/http/oauth-metadata.ts`)

- New `resolveOAuthTier(environment, backendUrl)`: `sandbox | staging_sandbox → sandbox`, `live | staging_live → live`; `custom` is derived from its backend host (`api.sandbox.` / `api.live.`), and when that cannot be classified the tier is **omitted** — a `localhost` stack keeps the pre-existing bare URL rather than a guessed tier that would send the human to the wrong backend.
- `buildOAuthUrls` stamps `?network=<tier>` on `authorizationUri` through `URL.searchParams` (one query string a client can extend with `&`); token / jwks / userinfo / issuer are untouched. An explicit `oauthUrls.authorizationUri` override is passed through unchanged.
- Param name + values are the API's and the embed widget's (`network`, `sandbox|live`; nvm-monorepo#3430 / #1787). `OAUTH_TIER_PARAM`, `OAuthTier` and `resolveOAuthTier` are exported from `@nevermined-io/payments/mcp` for integrators who build their own documents.
- `markdown/mcp-integration.md`: one paragraph under the discovery-endpoints table (`./scripts/generate-docs.sh` clean).

No public signature changed (`buildOAuthUrls` is module-private; the exported builders keep their shapes) — the advertised *value* changes, which is what integrators need to know: a hand-configured client should copy the endpoint from the discovery document, query string included.

Related to Issue #447

## Types of changes

- [ ] New feature
- [ ] Breaking change

## Checklist:

- [x] Tests Cover Changes
- [x] Documentation

## Test plan

- [x] `pnpm build && pnpm lint && pnpm typecheck:tests && pnpm test:unit` → green (lint: 5 pre-existing tsdoc/non-null warnings in files this PR does not touch, 0 errors); unit **49 suites / 652 passed, 1 skipped**.
- [x] `tests/unit/mcp/oauth_metadata_builders.test.ts` (+9): per-environment `authorization_endpoint` for all four named environments across all three documents; `network` is the param and the URL has exactly one query string; the tier is not stamped on token/jwks/userinfo/issuer; an `authorizationUri` override passes through; `resolveOAuthTier` for named / custom-derived / unclassifiable inputs; `custom` against localhost stays bare.
- [x] Mutation: dropping the tier from `authorizationUri` → **5 failed / 39 passed**. Replacing the `URL` build with string concat is an *equivalent* mutant today (every `Environments` frontend is an origin, so the base never carries a query) — the `URL` path is there for a frontend that does.
- [ ] e2e (`tests/e2e/test_mcp_oauth_e2e.test.ts`) asserts `authorization_endpoint` is defined and equal across the two documents; not re-run locally (needs staging credentials) — unchanged expectations.

https://claude.ai/code/session_01Cpykv9mTQL88E3oXwA11QF


src/mcp/http/oauth-metadata.ts
```diff
@@ -38,14 +112,30 @@ function buildOAuthUrls(frontendUrl: string, backendUrl: string): OAuthUrls {
...
+  environment: EnvironmentName,
+  backendForTier?: string,
+): OAuthUrls {
+  const known = environment in Environments
+  const effective: EnvironmentName = known ? environment : 'sandbox'
+  const envConfig = Environments[effective]
+  return buildOAuthUrls(
+    envConfig.frontend,
+    envConfig.backend,
+    resolveOAuthTier(effective, backendForTier || envConfig.backend),
```
r-marques: **🟡 SHOULD FIX — Under `custom` the tier was derived from `Environments.custom.backend`, not the backend the document publishes**

Reproduced by the panel: `environment: 'custom'` + `oauthUrls.tokenUri` pointing at `api.sandbox.nevermined.app` + `NVM_FRONTEND_URL=https://nevermined.app` published a sandbox `token_endpoint` next to a BARE `authorization_endpoint` — the exact #447 bug, silently. `resolveAuthorizationServer` in this file already derives the AS origin from the overridden `tokenUri`. **Folded:** `getOAuthUrls` passes `overrides?.tokenUri` as the backend the tier is derived from; pinned through the public path (`custom` + tokenUri override → `?network=sandbox` / `live` / bare for localhost) and by module reload under `NVM_BACKEND_URL`. Mutants "tier from the wrong field" (3 tests) and "ignore the override" (1 test) fail.

<sub>🤖 pr-review-toolkit · silent-failure-hunter + pr-test-analyzer · high</sub>

src/mcp/http/oauth-metadata.ts
```diff
@@ -13,23 +13,97 @@ import type {
...
+    default: {
+      let labels: string[]
+      try {
+        // WHATWG `hostname` is lowercased and carries no port/credentials/path.
+        labels = new URL(backendUrl).hostname.split('.')
+      } catch {
+        return undefined
+      }
+      const tierAfterApi = labels[labels.indexOf('api') + 1]
+      if (labels.includes('api') && (OAUTH_TIERS as readonly string[]).includes(tierAfterApi)) {
```
r-marques: **🟡 SHOULD FIX — Anchor the host classifier on the `api.<tier>` label pair — and cover the hosts the API actually serves**

`startsWith('api.sandbox.')` missed branded per-org subdomains (`<slug>.api.sandbox.nevermined.app`) and `mcp.api.sandbox.…`, while nothing pinned that a bare `sandbox` label elsewhere is NOT a tier (`includes('sandbox')` survived). **Folded:** the labels are split and the tier is the label right after `api`; pinned positive (`acme.api.sandbox.…`, `mcp.api.live.…`, uppercase + port) and negative (`sandbox.nevermined.app`, `api.nevermined.app/sandbox`). The "match any label" mutant fails 1 test.

<sub>🤖 pr-review-toolkit · code-reviewer + pr-test-analyzer · high</sub>

markdown/mcp-integration.md
```diff
@@ -31,6 +31,8 @@ When you start an MCP server, the library automatically exposes OAuth 2.1 discov
 
 These endpoints are generated automatically—no manual configuration required.
 
+The `authorization_endpoint` they advertise names the API tier your server runs against — `https://nevermined.app/oauth/authorize?network=sandbox` for `sandbox`, `?network=live` for `live` — because one Nevermined web app serves the consent screens for both tiers and boots on whatever tier the user's browser last chose. Clients keep that query string when they add their own parameters (RFC 6749 §3.1); if you configure a client by hand, copy the endpoint from your server's discovery document, query string included. A `custom` environment derives the tier from its backend host (`api.sandbox.…` / `api.live.…`, branded subdomains included); behind a host the SDK cannot classify — `localhost`, a proxy — no tier is stamped, so set `oauthUrls.authorizationUri` to the webapp URL **with** `?network=<tier>` yourself.
```
r-marques, line 34: **🟡 SHOULD FIX — An unclassifiable `custom` host silently reverted to the tier-blind URL, with no doc line naming the remedy**

Omit-never-guess is the right posture for a pure builder (the monorepo MCP refuses with a message; here the metadata endpoint must not crash), but the operator had no way to learn that `oauthUrls.authorizationUri` is the override — and `custom` IS reachable in production (`resolveEnvironmentName` defaults to it for a key without a recognised prefix). **Folded:** the remedy is on `OAuthUrls.authorizationUri`'s type doc, `resolveOAuthTier`'s JSDoc and this paragraph (with the branded-subdomain coverage stated).

<sub>🤖 pr-review-toolkit · silent-failure-hunter + code-reviewer · high</sub>

tests/unit/mcp/oauth_metadata_builders.test.ts
```diff
@@ -369,4 +372,129 @@ describe('OAuth Metadata Builders', () => {
...
+      process.env.NVM_FRONTEND_URL = frontend
+      const mod = await import('../../../src/mcp/http/oauth-metadata.js')
+      return mod.getOAuthUrls('custom')
+    }
+
+    test.each([
+      ['https://api.sandbox.nevermined.app', 'https://nevermined.app/oauth/authorize?network=sandbox'],
+      ['https://api.live.nevermined.app/', 'https://nevermined.app/oauth/authorize?network=live'],
+      ['http://localhost:3001', 'https://nevermined.app/oauth/authorize'],
+    ])('NVM_BACKEND_URL=%s → %s', async (backend, expected) => {
```
r-marques, line 584: **💡 NIT — The `custom`-stays-bare test hid its assertion behind `if (!process.env.NVM_BACKEND_URL)`**

It degraded to a vacuous `toContain('/oauth/authorize')` on any box with that var set, and went RED with `NVM_FRONTEND_URL` set and `NVM_BACKEND_URL` unset. **Folded:** replaced by the repo's `jest.resetModules()` + env + dynamic `import()` pattern (`tests/unit/environment-from-key-prefix.test.ts`) with both vars set explicitly and restored in `afterEach`; three cases (`api.sandbox.` / `api.live.` / localhost).

<sub>🤖 pr-review-toolkit · pr-test-analyzer + silent-failure-hunter + code-reviewer · high</sub>

src/mcp/http/oauth-metadata.ts
```diff
@@ -13,23 +13,97 @@ import type {
...
+  }
+}
+
+/**
+ * Append `?network=<tier>` to the authorize URL. Goes through the URL API so the result is a
+ * canonical absolute URL with one query string; a frontend that is not an absolute URL (a relative
+ * `NVM_FRONTEND_URL` such as `/webapp`) cannot be parsed, and falls back to plain concatenation —
+ * the same shape the tier-less URL always had for that misconfiguration.
+ */
+function withTierParam(authorizeUrl: string, tier: OAuthTier | undefined): string {
```
r-marques, line 102: **💡 NIT — `withTierParam`'s JSDoc claimed a property the concat fallback also has**

"`&` rather than a second `?`" distinguished nothing — the fallback picks `&` too, and every `Environments` frontend is an origin, so URL-vs-concat are byte-identical over every sane input (the panel measured it). **Folded:** the JSDoc now says what the URL path actually buys (a canonical absolute URL) and when the catch is reachable (a relative `NVM_FRONTEND_URL`); no query-bearing-base test added, deliberately — it would pin garbage-in.

<sub>🤖 pr-review-toolkit · pr-test-analyzer · medium</sub>

r-marques (commented):
## 🤖 Automated PR review — 🟡 Mergeable with nits (all findings folded)

The SDK twin of nvm-monorepo#3430: the MCP server's own RFC 8414 / OIDC documents now name the API tier on `authorization_endpoint`. The core is right (`network`, `sandbox|live`, built through `URL`, omitted rather than guessed); the panel found one real gap — under `custom` the tier came from `Environments.custom.backend` rather than the backend the document actually publishes when `oauthUrls.tokenUri` is overridden, reproducing the original bug silently — plus a host classifier that missed the API's branded subdomains and was not pinned against a bare `sandbox` label. Both folded.

⚠️ **Do not ship this before nvm-monorepo#3515 is on both tiers.** The official Python MCP client concatenates a second `?` onto the authorize URL (`mcp` 1.30 → 2.2.0); the webapp on production is still the strict #3433 parser, so a tier-qualified SDK document would turn Python-MCP clients away until #3515 deploys. This PR stays draft until then.

**What I reviewed**
- Diff range `origin/main...HEAD` — 7 files, head `cf4764e5`.
- Panel: code-reviewer, silent-failure-hunter, pr-test-analyzer (each delivered via file; none could run the suite in the review worktree — the numbers are the author's, re-verified after folding: `pnpm build && pnpm lint && pnpm typecheck:tests && pnpm test:unit` → 49 suites / 656 tests; 4 mutants killed).

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 3 should-fix, 3 nits, 1 praise — all folded in `cf4764e`.

### 🧭 Inline comments (7)

- 🟡 **SHOULD FIX** — Under `custom` the tier was derived from `Environments.custom.backend`, not the backend the document publishes — `src/mcp/http/oauth-metadata.ts:137`
- 🟡 **SHOULD FIX** — Anchor the host classifier on the `api.<tier>` label pair — and cover the hosts the API actually serves — `src/mcp/http/oauth-metadata.ts:60`
- 🟡 **SHOULD FIX** — An unclassifiable `custom` host silently reverted to the tier-blind URL, with no doc line naming the remedy — `markdown/mcp-integration.md:34`
- 💡 **NIT** — The `custom`-stays-bare test hid its assertion behind `if (!process.env.NVM_BACKEND_URL)` — `tests/unit/mcp/oauth_metadata_builders.test.ts:496`
- 💡 **NIT** — `withTierParam`'s JSDoc claimed a property the concat fallback also has — `src/mcp/http/oauth-metadata.ts:74`
- 💡 **NIT** — Unknown environment key now yields `?network=sandbox` (an improvement) — pinned — `src/mcp/http/oauth-metadata.ts:133`
- ✅ **NICE** — Safe classification; the right failure mode — `src/mcp/http/oauth-metadata.ts:49`

src/mcp/http/oauth-metadata.ts
```diff
@@ -38,14 +112,30 @@ function buildOAuthUrls(frontendUrl: string, backendUrl: string): OAuthUrls {
...
-function getOAuthUrlsForEnvironment(environment: EnvironmentName): OAuthUrls {
-  const envConfig = Environments[environment] || Environments.sandbox
-  return buildOAuthUrls(envConfig.frontend, envConfig.backend)
+function getOAuthUrlsForEnvironment(
+  environment: EnvironmentName,
+  backendForTier?: string,
+): OAuthUrls {
+  const known = environment in Environments
+  const effective: EnvironmentName = known ? environment : 'sandbox'
+  const envConfig = Environments[effective]
```
r-marques, line 166: **💡 NIT — Unknown environment key now yields `?network=sandbox` (an improvement) — pinned**

The fallback's `token_endpoint` was already sandbox, so the document is now internally consistent instead of half-bare. **Folded:** `getOAuthUrls('bogus')` pins both; the "unknown → live" mutant fails 1 test. A `localhost` custom stack still omits the tier (as the issue specified) — the nvm MCP twin maps localhost → sandbox instead; the two knowingly differ and the doc line tells a local operator the override.

<sub>🤖 pr-review-toolkit · silent-failure-hunter + code-reviewer · high</sub>

src/mcp/http/oauth-metadata.ts
```diff
@@ -13,23 +13,97 @@ import type {
...
+export function resolveOAuthTier(
+  environment: EnvironmentName,
+  backendUrl: string,
+): OAuthTier | undefined {
+  switch (environment) {
+    case 'sandbox':
+    case 'staging_sandbox':
+      return 'sandbox'
+    case 'live':
+    case 'staging_live':
```
r-marques, line 59: **✅ NICE — Safe classification; the right failure mode**

`backendUrl` is operator config (`NVM_BACKEND_URL`), never attacker input, and the tier only qualifies the same operator's frontend — so `api.sandbox.evil.com` is self-inflicted, not a boundary crossing; WHATWG `hostname` lowercases and strips port/credentials/path; omitting on an unclassifiable host is the webapp's documented pre-#3430 path. No consumer in `cli/` or `openclaw/` reads `authorizationUri`; `payments.mcp.start()` passes no `oauthUrls`, so the primary path always gets the stamped URL.

eruizgar91:
:eyes: Reviewing

src/mcp/http/oauth-metadata.ts
```diff
@@ -13,23 +13,97 @@ import type {
...
+}
+
+/**
+ * Append `?network=<tier>` to the authorize URL. Goes through the URL API so the result is a
+ * canonical absolute URL with one query string; a frontend that is not an absolute URL (a relative
+ * `NVM_FRONTEND_URL` such as `/webapp`) cannot be parsed, and falls back to plain concatenation —
+ * the same shape the tier-less URL always had for that misconfiguration.
+ */
+function withTierParam(authorizeUrl: string, tier: OAuthTier | undefined): string {
+  if (!tier) return authorizeUrl
```
eruizgar91, line 103: **MEDIUM** — When `resolveOAuthTier` cannot classify a `custom` backend it returns `undefined` and this line silently hands back the pre-#3430 bare URL. Omitting rather than guessing is right; doing it **with no signal at all** is the gap, and it is exactly where this twin parts company with the monorepo — in all three monorepo implementations the unclassifiable case is *loud*:

- `apps/api/src/oauth/oauth-frontend-url.ts` — `tierQualifiedFrontendUrl` **throws** (`ENVIRONMENT must be one of sandbox|live, got …`); it can never emit a bare or `?network=undefined` URL, and `well-known.service.spec.ts` pins that refusal.
- `apps/mcp/src/mcp/setup-delegation.ts` — refuses the tool call and names the remedy (`Set MCP_EMBED_BASE_URL and MCP_EMBED_NETWORK, or point NVM_API_BASE_URL at a recognised Nevermined API host`).
- `apps/api/src/organizations/integration/environment-info.ts` — a render path that must not throw, so it degrades to a **visible** `Unknown`/`Unknown`.

Here it degrades invisibly: the discovery document is served 200, `authorization_endpoint` is the bare URL, `Cache-Control: public, max-age=3600` (`src/mcp/http/oauth-router.ts:144`) puts it in client caches for an hour, and the operator's first evidence is a user on the wrong consent screen — the #447 symptom this PR exists to remove. The `custom` path is production-reachable: `resolveEnvironmentName` falls back to `custom` for any key without a recognised prefix ([`src/api/base-payments.ts:136`](https://github.com/nevermined-io/payments/blob/bef9839e0d9fdd3b358f75011ee9f71c66188311/src/api/base-payments.ts#L133-L139)).

A throw is the wrong trade for a pure builder feeding a metadata endpoint — agreed. But a **warn-once** costs nothing and this repo already has the exact pattern, module-level flag and all, a few files away: [`environmentOptionDeprecationWarned`](https://github.com/nevermined-io/payments/blob/bef9839e0d9fdd3b358f75011ee9f71c66188311/src/api/base-payments.ts#L140-L152). Something like:

```ts
if (!tier && environment === 'custom' && !tierlessWarned) {
  tierlessWarned = true
  console.warn(
    `[Nevermined] Could not derive the API tier from backend host '…' — ` +
      `authorization_endpoint will be advertised without ?network=. ` +
      `Set oauthUrls.authorizationUri to the webapp URL including ?network=sandbox|live.`,
  )
}
```

(It needs `environment` in scope, so it fits more naturally in `getOAuthUrlsForEnvironment` than in `withTierParam`.) Bounded, in-file, no design decision left open — this should land here rather than as a follow-up.

src/mcp/http/oauth-metadata.ts
```diff
@@ -13,23 +13,97 @@ import type {
...
+      return 'live'
+    default: {
+      let labels: string[]
+      try {
+        // WHATWG `hostname` is lowercased and carries no port/credentials/path.
+        labels = new URL(backendUrl).hostname.split('.')
+      } catch {
+        return undefined
+      }
+      const tierAfterApi = labels[labels.indexOf('api') + 1]
```
eruizgar91: **MEDIUM** — This repo already had a tier classifier for the *same* `?network=` parameter, and the two now disagree on real inputs. [`cli/src/utils/widget-redirect-flow.ts:33-45`](https://github.com/nevermined-io/payments/blob/bef9839e0d9fdd3b358f75011ee9f71c66188311/cli/src/utils/widget-redirect-flow.ts#L33-L45):

```ts
case 'custom':
  return /(^|[.\/])live([.\/]|$)/.test((process.env.NVM_BACKEND_URL || '').toLowerCase())
    ? 'live'
    : 'sandbox'
```

Whole-URL regex + **default `sandbox`**, versus this function's `api.<tier>` label pair + **omit**. Running both over the same `NVM_BACKEND_URL` (node, both implementations verbatim):

| `NVM_BACKEND_URL` | `resolveOAuthTier` | `resolveEmbedNetwork` |
|---|---|---|
| `https://api.sandbox.nevermined.app` | `sandbox` | `sandbox` |
| `https://api.live.nevermined.app` | `live` | `live` |
| `http://localhost:3001` | *(omitted)* | `sandbox` |
| `https://live.example.com` | *(omitted)* | **`live`** |
| `https://api-live.example.com` | *(omitted)* | `sandbox` |
| `https://live.api.sandbox.nevermined.app` | **`sandbox`** | **`live`** |

So one `custom` deployment can hand the OAuth consent screen `?network=sandbox` and the embed card flow `?network=live`, from one env var. The CLI's `sandbox` default is also precisely the guess this PR's JSDoc argues against, and it is shipping today — a `live` custom deployment lands its card flows on the sandbox backend, which surfaces as `Invalid or expired widget session token` and names nothing about the tier.

Worth noting alongside it: the monorepo's classifier is **two-axis** — `apps/mcp/src/config.ts` requires *both* a `nevermined.{app,dev}` suffix (`deriveEmbedBaseUrl`) *and* a `sandbox`/`live` **segment anywhere in the host** (`deriveEmbedNetwork`), and refuses when either half fails (`api.nevermined.app` → `undefined`). This one requires neither the suffix nor a segment, only the `api.<tier>` pair — so `https://sandbox.nevermined.app` classifies there and not here, and `https://api.live.example.com` classifies here and not there. Tighter-and-omitting is a defensible choice; three different rules for one `network` value across two repos is not, and nothing in the code says they are meant to differ.

**This one is a legitimate follow-up rather than a fix here**: the reconciliation crosses a package boundary — `cli/` depends on `@nevermined-io/payments@^1.1.0` from npm, so it cannot call the newly-exported `resolveOAuthTier` until this ships — and picking the single rule (segment-anywhere vs `api.<tier>`) is a decision this PR has not made. Please file it and link it, and in the meantime a one-line cross-reference in both JSDoc blocks naming the other classifier would stop the next reader re-deriving a third rule.

src/mcp/http/oauth-metadata.ts
eruizgar91: **LOW** — `resolveOAuthTier` returns on the environment name before it ever looks at `backendForTier`, so the fix folded for `custom` does not cover the four named environments. Verified by running the shipped logic:

```
getOAuthUrls('sandbox', { tokenUri: 'https://api.live.nevermined.app/oauth/token' })
  authorization_endpoint = https://nevermined.app/oauth/authorize?network=sandbox
  token_endpoint         = https://api.live.nevermined.app/oauth/token
  authorization_servers  = ["https://api.live.nevermined.app"]   // via resolveAuthorizationServer
```

That is the same internally-inconsistent document your own folded finding described ("a sandbox `token_endpoint` next to a tier-blind authorize URL"), just with the mismatch on the other axis: the user consents on the **sandbox** screen, the code goes to the **live** token endpoint, and the rejection reads as a bad or expired code rather than a tier mismatch. `staging_sandbox` → `api.live.nevermined.dev` behaves the same way.

I do **not** think the fix is "always derive from the override" — that would regress a legitimate same-tier proxy (`environment: 'sandbox'` + `tokenUri: 'https://gw.corp.com/oauth/token'` classifies to nothing and would lose a tier it currently gets right). Named-environment precedence looks like the correct behaviour; what is missing is that it is neither pinned nor warned. Two bounded options, either of which lands here:

- a test asserting the named-environment tier is unaffected by a cross-tier `tokenUri` override, so the precedence is deliberate rather than incidental — the `@param backendForTier` doc says "under `custom`", but nothing holds it there; or
- reuse the warn-once from F1 when a named environment's tier and the override host's tier disagree, since that combination is always a misconfiguration.

eruizgar91 (commented):
## Review — tier-qualified `authorization_endpoint` (TS twin of nvm-monorepo#3430)

The core is correct and the RFC question checks out: `network`/`sandbox|live` match the API's own constants, the value is built through `URL.searchParams` so the endpoint carries exactly one query string, and only `authorizationUri` is stamped. Three comments inline, none blocking the mechanism — two MEDIUM (a silent degradation that every monorepo sibling makes loud, and a contradictory sibling classifier already shipping in `cli/`) and one LOW (the named-environment mirror of the `custom` fix folded during self-review).

### Verified, came back clean

- **RFC 8414 / RFC 9728 / OIDC Discovery permit the query component.** RFC 6749 §3.1 states the authorization endpoint URI MAY carry an `application/x-www-form-urlencoded` query component "which MUST be retained when adding additional query parameters"; RFC 8414 and OIDC Discovery inherit that definition. Only `issuer` is constrained against query/fragment, and `issuer` is untouched. The doc paragraph's §3.1 citation is accurate.
- **Nothing downstream re-joins the URL.** Swept the whole repo for `authorizationUri` / `authorization_endpoint` / `oauth/authorize`: the only readers are the three builders in this file, `http.types.ts`, the docs and the tests. `cli/` and `openclaw/` never read it (matches your own ✅ NICE comment).
- **The official MCP TypeScript client preserves it.** `@modelcontextprotocol/sdk` (this repo's dep, `^1.24.3`; checked 1.30.0) does `new URL(metadata.authorization_endpoint)` then `searchParams.set(...)` for `response_type`/`client_id`/`code_challenge`/`redirect_uri` — no concatenation, so the tier survives. The Python client's `f"{endpoint}?{urlencode(params)}"` is the one that needs #3515, as your gate note says.
- **The #3515 tolerance covers exactly what this emits, with nothing to spare.** `apps/webapp/src/lib/oauth-tier.ts` splits the value at its **first `?`**, requires the head to be exactly `sandbox`/`live`, and recovers exactly **one** `key=value` pair (only one can survive `URLSearchParams`' `&` split anyway). A second advertised parameter would not be recoverable — so the "exactly one query string" test in this PR is load-bearing, not cosmetic. Worth keeping that link in mind if anything is ever added to this URL.
- **Host-classifier edge cases behave as the new tests claim.** Ran the shipped `resolveOAuthTier` over branded/`mcp.`-prefixed hosts, uppercase + port, `sandbox.nevermined.app`, `api.nevermined.app/sandbox`, malformed input. The `labels[labels.indexOf('api') + 1]` line reads risky when `indexOf` returns `-1` (it evaluates `labels[0]`), but the `labels.includes('api')` conjunct in front of it makes that unreachable — correct as written.
- **The e2e assertion cannot see a `custom` run.** `TEST_ENVIRONMENT.includes('live')` would classify `custom` as `sandbox` while the implementation deliberately emits no `network` param — but `tests/e2e/global-setup.ts:145-155` throws for any `TEST_ENVIRONMENT` outside the four named environments, so that input is unreachable. Not a finding.

### Considered and deliberately discarded

- **`issuer` is the frontend here and the backend in the monorepo** (`well-known.service.ts` publishes `issuer: apiHost`). Two documents describing the same authorization server disagree on its identifier, and RFC 8414 §3.3 wants the `issuer` to match the URL the metadata was fetched from — which `https://nevermined.app` does not, for a document served at `http://<mcp-server>/.well-known/oauth-authorization-server`. Entirely pre-existing, untouched by this PR, and fixing it changes a published identifier; it needs its own issue, not this one.
- **For an RFC 9728-compliant MCP client this document may never be read.** `resolveAuthorizationServer` puts the **backend** origin in `authorization_servers`, so a spec-following client fetches the API's RFC 8414 document — already tier-qualified by #3430 — rather than this server's. The change is still right (the root server-info and any client reading this server's AS metadata directly do consume it), just narrower in blast radius than the PR description implies.
- **`api.live.evil.com` → `live`.** Agreed with your ✅ NICE reasoning: `backendUrl` is operator config and the tier only qualifies the same operator's own frontend. Raised only as supporting detail under the classifier-divergence comment, not as a security finding.
- **`Cache-Control: public, max-age=3600` on the discovery routes** means clients can serve the old bare URL for up to an hour after an operator upgrades. Pre-existing and correct for discovery documents; a rollout note, not a defect.

De-duplicated against your own COMMENTED self-review (7 comments, all folded) — none of the three below overlaps with it; F3 is explicitly the other half of your first SHOULD-FIX.

<sub>Scope: 7 files, `+256/-10`, head `bef9839e`. No build, lint or test run — those gate separately.</sub>

eruizgar91 (approved):

Inline review: https://github.com/nevermined-io/payments/pull/455#pullrequestreview-5247824669

src/mcp/http/oauth-metadata.ts
r-marques: Folded in `04022f5`, in `getOAuthUrlsForEnvironment` as you suggested (it has `environment` in scope): a `custom` backend the rule cannot classify still gets the bare URL, but now with ONE `console.warn` naming the host (`'localhost:3001'`) and the remedy (`oauthUrls.authorizationUri` with `?network=sandbox|live`). Module-level flag, same shape as `environmentOptionDeprecationWarned`. Tests: the warning fires once and a second build in the same module (including `buildAuthorizationServerMetadata`) stays silent; a malformed backend warns rather than throws; a classifiable backend does not warn. Mutants: never latching the flag → 1 red; dropping the block → 2 red.

src/mcp/http/oauth-metadata.ts
r-marques: Folded in `04022f5` — both of your options, since the second falls out of the first: named-environment precedence is now pinned by a `test.each` over `sandbox`/`staging_sandbox`/`live` with a cross-tier `tokenUri` (the environment's tier is published, the override's `token_endpoint` is kept), a same-tier / unclassifiable proxy override (`gw.corp.com`) keeps its tier with no warning, and when the override host classifies to the OTHER tier it warns once, naming both. The host match moved into `tierFromHost` so the cross-check shares the exact rule the `custom` branch uses. Mutants: letting the override win → 3 red; dropping the warning → 3 red.

src/mcp/http/oauth-metadata.ts
r-marques: Agreed on the diagnosis and the table — I re-ran both classifiers and get the same six rows. The JSDoc half is folded in `04022f5`: `resolveOAuthTier` names `resolveEmbedNetwork` (segment-anywhere, defaults to `sandbox`) and the monorepo's `deriveEmbedNetwork` (suffix + segment, refuses), and `resolveEmbedNetwork` points back, both saying "do not derive another rule here".

On filing: which single rule the three converge on (segment-anywhere vs `api.<tier>`; default-`sandbox` vs omit vs refuse) is Rodolphe's call rather than mine, and the CLI's `sandbox` default is shipping behaviour a `live` custom deployment relies on today — so I'm putting the question to him now and will file the issue with his decision recorded in it, then link it here. Leaving this thread open until then so it gates the merge rather than being resolved on a promise.

r-marques: Decision made (Rodolphe, 2026-09-18): **all three converge on the `api.<tier>` label-pair rule, never guessing.** Filed with that decision recorded, not as an open question:

- nevermined-io/payments#456 — the CLI drops the `sandbox` default, refuses when unmatched and gains an `NVM_EMBED_NETWORK=sandbox|live` override (the twin of the monorepo's `MCP_EMBED_NETWORK`); it can call the exported `resolveOAuthTier` once a release with this PR is published, or port `tierFromHost` verbatim.
- nevermined-io/nvm-monorepo#3638 — `apps/mcp`'s `deriveEmbedNetwork` aligns its match to the same pair (refuse-with-remedy unchanged; `deriveEmbedBaseUrl` keeps the suffix, it derives a host not a tier), ideally with the pair match living once in `@nevermined-io/commons` next to `OAUTH_TIER_PARAM`.

Both JSDoc cross-references now name those issues (`a53db32`). Resolving.

